PR #121, opened 2014-09-05
Applicants to form a team during application process

AkshataDM:
During the application process, students should be able to form a team. One student creates a new team and adds members - her team partner and coaches. 
After the applications are reviewed, the admin sets the selects each team by setting the is_selected column to true ( is_selected is an attribute of a team which is visible as the first field in the edit form of each team ). Once the admin sets this column to true, the team members can see the complete edit form which has fields like project name, blog field, etc. 

The validations for the project attributes in the team model has been removed. Instead, we created a separate project model and set up association between the project model and team. This was to make sure that the project field is not empty when a team is selected. The reason we went with this method was because the validations began to kick in for the project attribute whenever an admin set a team as selected and the form was not getting saved ( which we do not want ). 

We have commented out the validations for kind attribute in team model, because we are not sure if this should be changed by a team or organizer. Do let us know how this should be done!


svenfuchs:
:heart: 

What's the status of this? Would be great if this could be merged :)

/cc @AkshataDM @lucaspinto @pxlpnk @bastilian

berlintam:
ditto what @svenfuchs said -  it would be cool if this was up and running for the new RGSoC!

berlintam:
How about if we aim to get this thing merged by February 16th. What do you all think? @bastilian @lucaspinto @carpodaster @svenfuchs @AkshataDM @pxlpnk

lucaspinto:
I am in!

carpodaster:
Hey @AkshataDM, thank you for this PR! Due to the branch for this PR being the base for #123, the creation of a project when forming a team has already been merged :smile:

AkshataDM:
Yay! Thank you :)

lucaspinto:
\o/

berlintam:
MOST EXCITING THING EVER!!!!!!

PR #134, opened 2015-02-19
fix the checkbox label in the user edit form

alicetragedy:
we are the mediator in the war between bootstrap and simple_form :clap: 
(checkbox label fix)


carpodaster:
:+1:

PR #157, opened 2015-03-09
added extra fields as per #89

sareg0:
All of Anika's extra fields have been added, but I would super duper appreciate a second set of eyes on it :)
Many thanks!

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -25,6 +25,15 @@ class User < ActiveRecord::Base
     'coachingcompany' => 'Providing a coaching team from our company',
   }
 
+  MONTHS_LEARNING = [
+    "1-3 ",
+    "4-6 ",
+    "7-9 ",
+    "10-12 ",
+    "12-24 ",
+    "24 + ",
+  ]
+
   include ActiveModel::ForbiddenAttributesProtection
   include Authentication::ActiveRecordHelpers
   include ProfilesHelper
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -26,12 +26,12 @@ class User < ActiveRecord::Base
   }
 
   MONTHS_LEARNING = [
-    "1-3 ",
-    "4-6 ",
-    "7-9 ",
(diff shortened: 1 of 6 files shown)


app/views/users/edit.html.slim
```diff
@@ -43,14 +43,10 @@
   = f.input :company_name, label: 'Company name'
   = f.input :company_info, label: 'Company info', hint: 'Where is the office, how much space can you provide, etc.'
 
-  = f.input :bio, label: 'About you', hint: 'You can use basic <a href="http://en.wikipedia.org/wiki/Markdown" target="_blank">Markdown</a> here.'.html_safe
```
carpodaster: We still need that one. It's for the public user profile. Here's an [example](https://teams.railsgirlssummerofcode.org/users/3) (it's the 'About me' part).

sareg0:
boom! Put it right back :)

app/models/user.rb
```diff
@@ -25,6 +25,15 @@ class User < ActiveRecord::Base
     'coachingcompany' => 'Providing a coaching team from our company',
   }
 
+  MONTHS_LEARNING = [
+    "1-3 ",
```
carpodaster: I wouldn't include the trailing whitespace. They'll end up in the database, too

carpodaster:
Hey @sareg0 – thanks for the PR! :candy: 

Can you add the the new fields to `app/views/application_drafts/_student_fields.html.slim` as well (plus updating the `student_params` in `ApplicationDraftsController`) ? Ideally the fields should be shared really, but currently they're separate.

carpodaster:
@sareg shall I add the fields to the `_student_fields` partial?

sareg0:
hey @carpodaster is it ok if I do it in the morning? Or have you taken care of it?

carpodaster:
@sareg0 no, I haven't done anything. Sure thing, do it in the morning – ping me if you have kwistions!

Issue #144, opened 2015-03-02
Application form: ping coaches and mentors


carpodaster:
* [x] Send an email when added to a team
* [x] Add `application_drafts/index` listing all applications that they are part of
* [x] Mentors: add `PUT application_drafts#signoff` (irreversible!)

sareg0:
Mentors will sign off AFTER close of applications. We will give teams a week after the closing of applications to get the mentor to sign off
This will save building extra functionality, and also make sure that there is a set period Mentors and Students know about to get the signoff

carpodaster:
@marcgreenstock are you working on the last item, the mentor signoff?

#188 has a skipped example that's a placeholder for the mentor ping (some email with a link, stating that they can now sign off that application).

Technically, the sign off isn't a thing on the draft anymore as it's already converted to an application. The controller action may very well be `PUT applications#signoff`. Or we can signoff on the drafts controller, proxying the signoff to the `has_one` - associated application record. The latter doesn't feel right in a RESTful sense, but would have the charm of re-using the index template and just add a button to the application drafts listing.

Issue #124, opened 2015-01-29
Use a mail interceptor that works w/o Mailtrap


carpodaster:
* Mailtrap must be set up in `.(zsh|bash)rc`, lets move it to something like dotenv
* Fallback on a blackhole mail interceptor if Mailtrap is not configured

carpodaster:
Also, only configure Mailtrap in [config/environments/development.rb ](https://github.com/rails-girls-summer-of-code/rgsoc-teams/blob/master/config/environments/development.rb) if the required env-variables are setup.

carpodaster:
Almost done due to the awesome work of @marcgreenstock in #159. Just need a blackhole interceptor to prevent mails from being sent in development should sendmail be configured to fire incoming mails into space (it is on my machine)

marcgreenstock:
@carpodaster how about [Letter Opener](https://github.com/ryanb/letter_opener)?

carpodaster:
@marcgreenstock nah, I was just thinking about someting that would avoid emails being send in in the background. Something like this:

```ruby
# config/initializers/mail_interceptor.rb
Mail.register_interceptor(GitBasedMailInterceptor) if Rails.env.development?

# lib/git_based_mail_interceptor.rb
class GitBasedMailInterceptor
  def self.delivering_mail(message)
    message.to = `git config user.email`
    message.bcc = message.cc = nil
  end
end
```
(Disclaimer: not tested)

PR #159, opened 2015-03-10
Send email when added to team

marcgreenstock:
This addresses the first item on https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/144 (Send an email when added to a team).

A few notes on what I have done:

1. I have monkey patched Mail::Message to respond to `:deliver_later` in `config/initializers/monkey_patch_mail.rb`. I've done this to make upgrading to Rails 4.2 easier when we have ActiveJob available. It essentially puts the mailer in a background job. This should be removed when upgrading to 4.2 .
2. I've added the [foreman gem](https://github.com/ddollar/foreman). It is nicer and easier to have your environment variables in a `.env` file. I've also added `.env` to the .gitignore file. I can update the readme to reflect this in the Mailtrap section.
3. I had a bit of trouble with other tests failing because they were trying to deliver the email. I've set `action_mailer.raise_delivery_errors` to `false`. Let me know if this is unacceptable, and I'll find another way.

Wiggle

Bed time!

### DIFF
--- a/spec/mailers/role_mailer_spec.rb
+++ b/spec/mailers/role_mailer_spec.rb
@@ -1,0 +1,29 @@
+require "spec_helper"
+
+describe RoleMailer, type: :mailer do
+  describe 'user_added_to_team' do
+    let(:user) { FactoryGirl.create(:user) }
+    let(:team) { FactoryGirl.create(:team) }
+    let(:role) { FactoryGirl.create(:mentor_role, user: user, team: team) }
+
+    before do
+      subject
+    end
+
+    subject do
+      described_class.user_added_to_team(role)
+    end
+
+    it 'has ENV["EMAIL_FROM"] as from' do
+      expect(subject.from).to include(ENV['EMAIL_FROM'])
+    end
+
+    it 'has a the users email as to' do
+      expect(subject.to).to include(user.email)
+    end
+
+    it 'sets the subject' do
+      expect(subject.subject).to eq("[rgsoc-teams] You have been added to the #{team.name} team.")
+    end
+  end
+end
--- a/.env-example
+++ b/.env-example
@@ -1,0 +1,2 @@
+MAILTRAP_USER=InboxUsername
+MAILTRAP_PASSWORD=InboxPassword
--- a/.gitignore
+++ b/.gitignore
@@ -7,7 +7,7 @@
 /config/database.yml
 .ruby-version
 .ruby-gemset
-
+.env
 .DS_Store
 
 # Ignore bundler config
--- a/Gemfile
+++ b/Gemfile
@@ -55,6 +55,7 @@ group :development do
   gem 'pry'
   gem 'spring'
   gem 'spring-commands-rspec'
+  gem 'foreman'
 end
 
 group :test do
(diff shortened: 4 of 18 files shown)


marcgreenstock:
Oh, one more thing:

I took advantage of [Action Mailer Previews](http://guides.rubyonrails.org/4_1_release_notes.html#action-mailer-previews). In development you can go to http://localhost:3000/rails/mailers and see a preview of the emails.

spec/mailers/role_mailer_spec.rb
```diff
@@ -0,0 +1,29 @@
+require "spec_helper"
+
+describe RoleMailer, type: :mailer do
+  describe 'user_added_to_team' do
+    let(:user) { FactoryGirl.create(:user) }
+    let(:team) { FactoryGirl.create(:team) }
+    let(:role) { FactoryGirl.create(:mentor_role, user: user, team: team) }
+
+    before do
```
carpodaster: I don't think we need to explicitly call subject before

carpodaster: Nevermind, I just saw that you override subject to run the action :+1:

carpodaster:
Thanks @marcgreenstock ! So many goodies in this PR, I can't even :smile: 

Two minor things: 

1. Can you add your `.env` as `.env-example` to the repo? I will take that one as base for #124 then

2. About `action_mailer.raise_delivery_errors = false`: it's odd that we get errors. I'll check it out locally myself to see what kind of errors we get. The tests should never send real mail, but maybe there are some expectations failing due to mails being shoved into background queue.

marcgreenstock:
Hi @carpodaster just added to this RP. i like to use [letter_opener](https://github.com/ryanb/letter_opener) to open emails on development. Let me know if this is something you would like to use as a fallback to mailtrap.

anikalindtner:
thanks so much @marcgreenstock!!!! :sparkles:

marcgreenstock:
@carpodaster  regarding `raise_delivery_errors`, I'm thinking it is because of `require 'sucker_punch/testing/inline'` on [spec_helper.rb#L32](https://github.com/rails-girls-summer-of-code/rgsoc-teams/blob/master/spec/spec_helper.rb#L32). :)

carpodaster:
@marcgreenstock  I could make the tests pass with this when `raise_delivery_errors` was set back to true:

```diff
--- a/app/mailers/role_mailer.rb
+++ b/app/mailers/role_mailer.rb
@@ -6,2 +6,3 @@ class RoleMailer < ActionMailer::Base
   def user_added_to_team(role)
+    return unless role.user.email.present?
     @role = role
```
It's a bit too defensive for my taste, but technically we can't guarantee that every user has an email address. `User#complete_from_github` should grab it, but only if they entered a legal GH handle.

marcgreenstock:
@carpodaster no dramas, I've set it back to default in the test environment and added a robust check on the mailer. I think this is good practice anyway.

carpodaster:
@marcgreenstock Superawesome – ¡muchas gracias!  Let's :shipit:

alicetragedy:
:heart_eyes: 
also, DAT GIF

Issue #139, opened 2015-02-26
Upgrade to Rails v4.2


carpodaster:
* Upgrade to latest 4.1.x first
* Upgrade from 4.1.x to 4.2.0

Tools: railsdiff.org

* [ ] Remove monkey patch from `config/initializers/monkey_patch_mail.rb`

marcgreenstock:
wow, we merged? everything working as expected?

alicetragedy:
:tada: :tada: :tada: :tada: :tada:

carpodaster:
>  everything working as expected?

I hope so :trollface:   Nah seriously, I'm not seeing any strange behaviour. Thank you for your PR, @marcgreenstock !

PR #160, opened 2015-03-10
Upgrade to rails 4.2

marcgreenstock:
Seeing as though I left a mess with the mail monkey patch I decided to try and upgrade to Rails 4.2.

This resolves https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/139.

A few notes on what is done:

1. Rails upgraded to 4.2
2. cancan has been replaced by [cancancan](https://github.com/CanCanCommunity/cancancan)
3. All gems have been updated with the exception of [aws-sdk](https://github.com/aws/aws-sdk-ruby), changes to the aws-sdk are significant, see the [upgrading nodes](https://github.com/aws/aws-sdk-ruby/blob/master/UPGRADING.md). This is perhaps justification for a separate issue.
4. There were a number of deprecation warnings that have been solved, ~~there are still a few more left though~~.
5. I have converted all mailers to use `deliver_later`, ActiveJob style. In doing so I have deleted all the workers.

I had a problem with this https://github.com/rails-girls-summer-of-code/rgsoc-teams/commit/56207cd8d6ef37a7368194c7f8990f50476a343d see the commit description, I am not sure this is tested correctly.

Also in [comments_controller_spec.rb](https://github.com/rails-girls-summer-of-code/rgsoc-teams/blob/71cb736d9f84dfd967ce099821d48bac62916590/spec/controllers/comments_controller_spec.rb). I changed the spec to not test for the delivery, but for the enqueueing of the job. I'll run the coverage shortly to ensure the mailer is tested correctly.

I also noticed two pending tests: `ApplicationsController as an authenticated user POST create creates a new application` and `ApplicationsController as an authenticated user POST create fails to create invalid record` make a lot of noise, but don't fail the tests.

make it stop

Enjoy!


carpodaster:
And again, thank you @marcgreenstock for being so busy :bee: 

I'm afraid it may take some time to fully review this one, but I'll try to be quick about it!

marcgreenstock:
No dramas @carpodaster this isn't something you just merge, although that button does look pretty tantalising doesn't it?

spec/models/application_spec.rb
```diff
@@ -59,10 +59,9 @@
 
     it 'removes the flag if the value is 0' do
       flag = flags.sample.to_sym
-      subject.flags = [flag]
```
carpodaster: I'm not too sure on this fella, either. This meta-programming loop in `Application` is test covered according to https://codeclimate.com/github/rails-girls-summer-of-code/rgsoc-teams/Application, but it also has a ternary op in it so I don't know of both branches are really tested.

The elements of the `flags` array now being strings makes sense: I don't think AR serializes symbols. We just need to investigate why it would still return symbols occasionally then.

As for the assignment in L62: I don't know if it's part of the test (ie. checking whether assigning the full array instead of individual values through the meta-programmed quasi-boolean accessors) or just setting up the world for the test. I will check it our locally and play with it myself a little.

carpodaster:
Oh, can you run a `rake rails:update` and diff through the changes? I think it may have been skipped for 4.1 and I want to make sure we get all the new defaults. We can probably skip the use of `secrets.yml` for now since `config/initiliazer/secret_token.rb` is already setup securely – I can make the switch after this PR.

carpodaster:
> although that button does look pretty tantalising doesn't it?

Heh, it sure does! :smile:

spec/models/application_spec.rb
marcgreenstock: Thanks, a few things that concern me with this:

* I wasn't quite sure if `flags=` was ever accessed directly.
* Not sure how these were stored in production with AR 4.1, if they were strings or symbols.
* From what I could gather is using the `flags=` method AR stores keeps the instance as a symbol, then saves it as a string.

Anyway the doesn't give instil confidence in it and the code might be doing unpredictable things. I can take another crack at it.

marcgreenstock:
I didn't know about rake `rails:update`. I'll go though it and get all the goodies in there.

`secrets.yml` is a good abstraction of `ENV` variables as it keeps them all in one place. It's a good idea to have imo.

marcgreenstock:
There should be a PSA on comparing diffs while tired. SMTP settings are back in production.rb

marcgreenstock:
I've just added this to the PR https://github.com/marcgreenstock/rgsoc-teams/commit/48ede643c1b971f6a9e7fedf68c7b0793bfcb390 to send devise emails through activejob too see [Devise ActiveJob Integration](https://github.com/plataformatec/devise#activejob-integration) for more details.

carpodaster:
@marcgreenstock you know what? Let's merge this fella, wdyt? :shipit:

carpodaster:
@marcgreenstock we've got some merge conflicts now. Could you merge the current master into your branch and push again?

config/environments/production.rb
```diff
@@ -1,66 +1,82 @@
...
 
-  # Specifies the header that your server uses for sending files
-  # config.action_dispatch.x_sendfile_header = "X-Sendfile" # for apache
-  # config.action_dispatch.x_sendfile_header = 'X-Accel-Redirect' # for nginx
+  # Specifies the header that your server uses for sending files.
+  # config.action_dispatch.x_sendfile_header = 'X-Sendfile' # for Apache
+  # config.action_dispatch.x_sendfile_header = 'X-Accel-Redirect' # for NGINX
 
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
   config.force_ssl = ENV['USE_SSL']
```
pxlpnk: We should add a check here that this is actually set.

pxlpnk:
I just took a look at it and looks good so far. I would welcome a check for production that `USE_SSL` is not set. :ship:

config/environments/production.rb
carpodaster: I think it does no harm if it's not set. It will degrade to "normal" HTTP. Since we already had TLS-STS enabled, browser who had already accessed the app will still insist on using STS, but that's nothing we can remedy by checking for the ENV variable.

carpodaster:
Thanks for throwing another pair of eyes and brain matter on it, @pxlpnk ! :four_leaf_clover:

config/environments/production.rb
pxlpnk: I would have prevented the app from starting. and iirc we now have a SSL cert. I might also be over cautious about this things.

marcgreenstock:
I'm without a computer at the moment, spilling coffee on it is bad for its health. I'll be back up tomorrow and merge master into this branch. I was thinking of adding a few more tests, mostly around things that I am not 100% confident about. There is no rush for this one IMO, it's rather a conversation that will eventually get merged.

config/environments/production.rb
marcgreenstock: Ideally in production.rb we should set force_ssl to true. But that isn't the scope of this PR. This PR should leave everything as is, except for the rails version of course.

Perhaps another issue is in order for this topic?

carpodaster: I've changed it in master: 7664ddc10013cfb7c79a710939930f20bbd38a69. It should work with nil since it's falsey, but I haven't checked if it blocks booting the server.

pxlpnk:
If I can make some time on the weekend I would take on this PR and use the changes to make a fresh mergable version of it. Any objections?

carpodaster:
:+1:

PR #161, opened 2015-03-10
Application drafts index

marcgreenstock:
This resolves the second item in https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/144

It is not linked anywhere at the moment, feel free to shoot me directions on where this link should go and I'll pop it in.

The table is a bit sparse, not sure what needs to go in there, but the functionality is there, again, let me know what other columns should exist.

While looking at the routes I noticed this:

```Ruby
resources :application_drafts, except: [:show, :destroy] do
  put :apply, on: :member
end
```

The `put :apply` route doesn't seem to have an action, is this right?

### DIFF
--- a/.env-example
+++ b/.env-example
@@ -1,0 +1,2 @@
+MAILTRAP_USER=InboxUsername
+MAILTRAP_PASSWORD=InboxPassword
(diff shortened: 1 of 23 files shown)


marcgreenstock:
Hmmm... why did github include these other commits, is diff looks wrong

marcgreenstock:
Never mind closing this one off and making a new one

PR #162, opened 2015-03-10
Application drafts index

marcgreenstock:
This resolves the second item in https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/144

It is not linked anywhere at the moment, feel free to shoot me directions on where this link should go and I'll pop it in.

The table is a bit sparse, not sure what needs to go in there, but the functionality is there, again, let me know what other columns should exist.

While looking at the routes I noticed this:

```Ruby
resources :application_drafts, except: [:show, :destroy] do
  put :apply, on: :member
end
```

The `put :apply` route doesn't seem to have an action, is this right?

### DIFF
--- a/app/controllers/application_drafts_controller.rb
+++ b/app/controllers/application_drafts_controller.rb
@@ -6,6 +6,10 @@ class ApplicationDraftsController < ApplicationController
 
   helper_method :application_draft
 
+  def index
+    @application_drafts = current_user.application_drafts.order('created_at DESC')
+  end
+
   def new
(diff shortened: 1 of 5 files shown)


marcgreenstock:
Forgot to add a funny gif! Here you go...

mask

carpodaster:
> The `put :apply` route doesn't seem to have an action, is this right?

That's right, that was probably me being over-ambitious. It will be put to use when we create an application record from the draft (i.e. in #163)

carpodaster:
But hell, let's merge it – thanks a lot, @marcgreenstock ! :bouquet:

PR #167, opened 2015-03-17
Feature/validations for application draft

ramonh:
Together with @carpodaster, I've added validations for the `Student` fields, along with that of those of the `Coach`, and the `ApplicationDraft` itself.

### DIFF
--- a/spec/models/application_draft_spec.rb
+++ b/spec/models/application_draft_spec.rb
@@ -7,10 +7,8 @@ RSpec.describe ApplicationDraft do
     it { is_expected.to validate_presence_of :team }
 
     context 'apply validations' do
-      let(:student0) { Student.new }
-
       before do
-        allow(subject).to receive(:students).and_return([student0])
+        allow(subject).to receive(:students).and_return([])
       end
 
       shared_examples_for 'proxies :apply validation' do |attribute|
--- a/spec/models/application_draft_spec.rb
+++ b/spec/models/application_draft_spec.rb
@@ -37,17 +37,48 @@ RSpec.describe ApplicationDraft do
 
       end
 
-      context 'with two students' do
+      Student::REQUIRED_DRAFT_FIELDS.each do |attribute|
+        let(:ace_student)     { double.as_null_object }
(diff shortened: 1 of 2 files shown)


carpodaster:
Hm, it doesn't have all the patches. Can you PR against `rails-girls-summer-of-code:master` instead of `rails-girls-summer-of-code:feature/validations-for-application-draft`? That probably doesn't work after the PR has been issued and you may have to close this PR again.

screen shot 2015-03-17 at 20 46 13

carpodaster:
Then again, nevermind. We know the missing bits were working on sunday and this patchset is less noisy than comparing the whole set.

ramonh:
Oops, not sure I understood. I forked off the branch we were working on on Sunday

spec/models/application_draft_spec.rb
```diff
@@ -37,17 +57,48 @@
...
+          end
         end
 
         context 'both being valid' do
-          skip
+          let(:ace_student_2)   { double.as_null_object }
+          let(:value_2)         { SecureRandom.hex(12)  }
+
+          before do
+            allow(subject).to receive(:students).and_return([ace_student, ace_student_2])
```
carpodaster: You should be able to re-use the `ace_student` from up above as we don't have a mechanism in place that would check for the students not being the same reference. If it works w/o, lines 88-90 can be removed as well.

spec/models/application_draft_spec.rb
```diff
@@ -6,6 +6,26 @@
   context 'with validations' do
     it { is_expected.to validate_presence_of :team }
 
+    context 'apply validations' do
+      let(:student0) { Student.new }
+
+      before do
+        allow(subject).to receive(:students).and_return([student0])
```
carpodaster: Will it break if we leave out stubbing the `students` tuple?

ramonh:
All done! Thanks for the feedback.

ramonh:
Yay!

sareg0:
Fantastic your guys!!

PR #171, opened 2015-03-24
[RFC] Use a state machine to manage ApplicationDraft states

cypher:
Previously, the state of an ApplicationDraft was controlled by setting
`*_by` attributes. While this works as long as there is a small number
of states, this gets complicated quickly if the possible states
increases, as well as managing transitions between states.

When an ApplicationDraft transitions from the 'draft' state to the
'applied' state by calling `submit_application`, it sets the
`applied_at` column to either a supplied DateTime, or the current
DateTime.

Missing/ToDo: The `ready?` method seems to be intended as a
guard to make sure an ApplicationDraft is ready to be submitted,
but it currently always returns false, and so it can't be used as a
transition guard.

### DIFF
--- a/app/models/application_draft.rb
+++ b/app/models/application_draft.rb
@@ -84,9 +84,7 @@ class ApplicationDraft < ActiveRecord::Base
         self.applied_at = applied_at_time || Time.now
       end
 
-      # If the transition should only happen if the object is `ready?`,
-      # add a :guard => :ready? here
-      transitions :from => :draft, :to => :applied
+      transitions :from => :draft, :to => :applied, :guard => :ready?
     end
   end
 
--- a/app/models/application_draft.rb
+++ b/app/models/application_draft.rb
@@ -71,11 +71,7 @@ class ApplicationDraft < ActiveRecord::Base
     false # valid?(:apply)
   end
 
-  def state
-    aasm_state
-  end
-
-  aasm :no_direct_assignment => true do
+  aasm :column => :state, :no_direct_assignment => true do
     state :draft, :initial => true
     state :applied
 
--- a/app/models/application_draft.rb
+++ b/app/models/application_draft.rb
@@ -2,6 +2,8 @@ class ApplicationDraft < ActiveRecord::Base
 
   include HasSeason
 
+  include AASM
+
   # FIXME
   STUDENT0_REQUIRED_FIELDS = Student::REQUIRED_DRAFT_FIELDS.map { |m| "student0_#{m}" }
   STUDENT1_REQUIRED_FIELDS = Student::REQUIRED_DRAFT_FIELDS.map { |m| "student1_#{m}" }
@@ -69,8 +71,19 @@ class ApplicationDraft < ActiveRecord::Base
     false # valid?(:apply)
   end
 
-  def state
-    (applied_at? ? 'applied' : 'draft').inquiry
+  aasm :column => :state, :no_direct_assignment => true do
+    state :draft, :initial => true
+    state :applied
+
+    event :submit_application do
+      after do |applied_at_time = nil|
+        self.applied_at = applied_at_time || Time.now
+      end
+
+      # If the transition should only happen if the object is `ready?`,
+      # add a :guard => :ready? here
+      transitions :from => :draft, :to => :applied
+    end
   end
 
   private
--- a/app/models/application_draft.rb
+++ b/app/models/application_draft.rb
@@ -71,7 +71,11 @@ class ApplicationDraft < ActiveRecord::Base
     false # valid?(:apply)
   end
 
-  aasm :column => :state, :no_direct_assignment => true do
+  def state
+    aasm_state
+  end
+
+  aasm :no_direct_assignment => true do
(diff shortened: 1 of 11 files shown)


db/schema.rb
```diff
@@ -49,6 +49,7 @@
     t.datetime "updated_at"
     t.datetime "applied_at"
     t.integer  "updater_id"
+    t.text     "state",                       default: "draft", null: false
```
carpodaster: I think the new default column is now `aasm_state` (which also should also remove the need for the  `column: :state` option in the model). Can you change it for the less generic name? It could also be a string column instead of a text one, but I guess that makes little difference

app/models/application_draft.rb
```diff
@@ -69,8 +71,19 @@ def ready?
...
+    state :applied
+
+    event :submit_application do
+      after do |applied_at_time = nil|
+        self.applied_at = applied_at_time || Time.now
+      end
+
+      # If the transition should only happen if the object is `ready?`,
+      # add a :guard => :ready? here
+      transitions :from => :draft, :to => :applied
```
carpodaster: The guard is indeed that the model should be valid in the `:apply` validation context (for which `#ready?` is just a proxy). It's short-circuited to `false` right now because we don't have the proper controller action in place that will actually accept the submitted application.

Can you add the guard so that we don't forget it later?

db/schema.rb
cypher: I decided to go with `state` because a few of the controllers/helpers depend on a `state` method/attribute being available, and `aasm` doesn't care either way.

But sure, I can rename it, and add a `state` that delegates to `aasm_state`. Or would you prefer to rename the calls as well?

cypher: Also, in my experience, `varchar` offers no gains over `text` in PostgreSQL, but `text` has the upside that you'll never have to recast the column in case you need something more than e.g. 255 chars.

app/models/application_draft.rb
cypher: Sure! I just left it out because it broke the tests :)

carpodaster:
Hey @cypher, thanks! I've made a few comments but on a more general level:

> While this works as long as there is a small number of states, this gets complicated quickly if the possible states increases, as well as managing transitions between states.

The thing is: it's very unlikely that there will be more states than `draft` and `applied`. `applied`  is the final state as the `ApplicationDraft` will have been transformed into an `Application` record by then.

With that in mind, I'm not really in favor of adding the external dependency of aasm while also increasing complexity (the `submit_application(*)` adds some magic for beginners / people not familiar with aasm)

Re: your presentation »it's never too late to add a finite state machine«: do you see any additional states for an `ApplicationDraft` heading our way?

db/schema.rb
carpodaster: > I decided to go with state because a few of the controllers/helpers depend on a state method/attribute being available

Ah, you're right. I thought the calls to `state` were all changed in favor of the aasm predicates. But in that case, it would even count as another argument against adding aasm as adding it doesn't fully hide away the underlying state implementation or state-holding data attributes.

carpodaster: > Also, in my experience, varchar offers no gains over text in PostgreSQL

Ok cool. Yes, I have had my fair share of truncated strings on varchar cols, too, and while I doubt we will ever have states > 255 chars, let's stick with text!

cypher:
TL;DR: A state machine adds little overhead, and provides stability and future extensibility.

I've started working on #144, and the way I'm currently implementing it is actually an additional state in ApplicationDraft. That is, I've added a `mentor_signoff` event, that transitions an ApplicationDraft from `applied` to `signed_off`, which in turn could then trigger the conversion of an ApplicationDraft into an Application. This also ensures that a signed off ApplicationDraft cannot return to a previous state due to a bug.

Additionally, something like a feedback loop might also be desirable at some point, that is, mentors could "reject" a draft, and return it to the applicants for additional editing. This could easily be implemented via the state machine, especially when it comes to notifying all the people involved.

In theory, you could even stop using an extra model for the draft, and keep everything in the `Application` model, and then simply use a state machine to manage the different validations at each stage (but that would most likely be quite an undertaking, and probably not worth it).

db/schema.rb
cypher: Sorry, saw your comment too late. I'll revert 8f6eefef.

Actually, I was able to get rid of most of the `state` calls, instead calling e.g. `applied?` directly on a draft. The only place where `state` is still used is when it's displayed in a template (`application_drafts_helper.rb`, line 12)

carpodaster:
> That is, I've added a `mentor_signoff` event, that transitions an ApplicationDraft from applied to `signed_off`

I buy it! Good stuff :smile:

app/models/application_draft.rb
carpodaster: Can you add a test that checks that one can / cannot transition to `applied` with stubbed return values of `ready?` ?

cypher: Will do.

alicetragedy:
yaaaaayyyyy :balloon:

Issue #163, opened 2015-03-10
Application form: create an application from a valid draft


carpodaster:
See also ~~#145~~

ramonh:
Hey @carpodaster :)

I've enabled the apply state validation for an `ApplicationDraft` in #173.

I'll begin investigating how we can create an `Application` from said draft.

cypher:
@ramonh I've converted `ApplicationDraft` to use a state machine in https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/171, and have started on working #144 to implement the mentor signoff, which in turn would then trigger the conversion of an `ApplicationDraft` to an `Application`. Maybe you'd like to collaborate on that?

carpodaster:
@ramonh @cypher Thinking about the mentor-signoff triggering the conversion: we have the not-entirely theoretical option that a mentor is not available to sign off an application draft in time. The selection committee might still decide on one such application on a case-by-case basis or contact the mentor themselves / ping them but in any case, there would be an application record to work on. Maybe we should make the mentor-sign off a mandatory state then?

carpodaster:
^-  Edit: should **not** make the mentor-signoff a mandatory state 
![Doh](https://rafinhaea7arte.files.wordpress.com/2010/04/homer-doh.gif)

ramonh:
Hey @carpodaster and @cypher. Sounds good. I'll take a look at this PR and see how I can help. :+1: 

In the meantime, I've also implementation the `ApplicationDrafts#apply` controller action here: ramonh@91d4f4bfd391b3aa5a2972cadaff1cd722ab8fe5

PR #173, opened 2015-03-25
Validate readiness of ApplicationDraft

ramonh:
This pull request enables the validation of `ApplicationDraft`'s `ready` state.

This brings us a step closer to completing #163.

### DIFF
--- a/app/models/application_draft.rb
+++ b/app/models/application_draft.rb
@@ -66,7 +66,7 @@ class ApplicationDraft < ActiveRecord::Base
   end
 
   def ready?
-    false # valid?(:apply)
+    valid?(:apply)
   end
 
   def state
--- a/spec/controllers/application_drafts_controller_spec.rb
+++ b/spec/controllers/application_drafts_controller_spec.rb
@@ -168,7 +168,7 @@ RSpec.describe ApplicationDraftsController do
 
       context 'for a valid draft' do
         before do
-          allow_any_instance_of(ApplicationDraft).to receive(:valid?).with(:apply).and_return(true)
+          allow_any_instance_of(ApplicationDraft).to receive(:ready?).and_return(true)
         end
 
         it 'is go' do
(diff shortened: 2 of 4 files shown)


spec/controllers/application_drafts_controller_spec.rb
```diff
@@ -168,7 +168,7 @@
 
       context 'for a valid draft' do
         before do
-          allow_any_instance_of(ApplicationDraft).to receive(:valid?).with(:apply).and_return(true)
+          allow_any_instance_of(ApplicationDraft).to receive(:ready?).and_return(true)
```
carpodaster: The tests should pass with the old stub in place since `#ready?` only forwards to that.

carpodaster: In fact, it _could_ fix the currently failing spec :wink:

app/models/application_draft.rb
```diff
@@ -66,7 +66,7 @@ def role_for(user)
   end
 
   def ready?
-    false # valid?(:apply)
+    valid?(:apply)
```
carpodaster: We can't merge this PR yet as that would enable the submit button in app/views/application_drafts/new.html.slim:63 but we do not have the `ApplicationDrafts#apply` controller action yet.

spec/controllers/application_drafts_controller_spec.rb
ramonh: This is true! However, the spec indeed fails as we're missing the ability to create an `Application` from an `ApplicationDraft`. I'll investigate.

app/models/application_draft.rb
ramonh: Ah yes, makes sense. I'll look into adding this.

ramonh:
I think it makes sense to close this PR for now, and focus on getting the `ApplicationDrafts#apply` controller action working.

PR #174, opened 2015-03-26
Use ApplicationDrafts#apply controller action instead of that for Application

ramonh:
While working on #163, I took a look at the Rails routes, and saw that we don't have a controller action for `Application#apply`, but we do have one for `ApplicationDraft#apply`. Using this one instead fixes a few issues I've been having and makes sense to me. Any chance that this is the action we wanted to use?

Thanks in advance!

### DIFF
--- a/app/views/application_drafts/new.html.slim
+++ b/app/views/application_drafts/new.html.slim
@@ -61,7 +61,7 @@ p.attention
 
 - if application_draft.draft?
   - if application_draft.ready?
-    = simple_form_for application_draft, url: apply_application_path(application_draft.application), method: :put do |f|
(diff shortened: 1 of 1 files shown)


carpodaster:
Hi @ramonh – thanks! I'm not sure if we have a test example that checks the branch nested under the `if application_draft.ready?`. Could you check if there is and if not add a simple view test for it? (located in `spec/views`). Don't hesitate to ping me for further questions – thanks in advance :smile:

carpodaster:
Hey @ramonh, I »stole« your commit and added it to my PR #188 – I'm closing this one.

PR #175, opened 2015-03-30
Fixed warnings during test ("rake spec") run

jasnow:
1. Changed cancan to cancancan (better maintianed) to get rid of these 2 warnings: 
"failure_message_for_should_not is deprecated" and
"failure_message_for_should is deprecated".
 2. Changing "template" to "view" in spec file to get rid of this warning:
"template is deprecated. Use view instead."
 3. Removed the "implicit join" config line to get rid of this warning: 
"Implicit join references were removed with Rails 4.1.Make sure to remove this configuration because it does nothing."
 4. SIDE EFFECT: Recreated schema.rb file (2 lines changed) while following README.


config/application.rb
```diff
@@ -53,7 +53,8 @@ class Application < Rails::Application
     # Version of your assets, change this if you want to expire all your assets
     config.assets.version = '1.0'
 
-    config.active_record.disable_implicit_join_references = true
+    # REMOVED IN RAILS 4.1.
```
carpodaster: Can you remove these two lines alltogether?

db/schema.rb
```diff
@@ -204,8 +204,8 @@
 
   create_table "teams", force: true do |t|
     t.string   "name"
-    t.datetime "created_at",      null: false
-    t.datetime "updated_at",      null: false
+    t.datetime "created_at",                      null: false
```
carpodaster: The generated in `db/schema.rb` will differ on different machines. We can keep it that way, but sooner or later someone will get a schema.rb that removes the whitespace again. I tend to ignore these changes as they only create noise in the commit log. But as said: let's keep them here :smile:

carpodaster:
Hi @jasnow – thanks for the PR! :smile: I've added a comment about comments (sooo meta ;) ). I'm happy to merge after the suggested change.

db/schema.rb
jasnow: Nice to know why - SUPER thanks.

carpodaster:
Thanks again @jasnow – merged! :)

Issue #168, opened 2015-03-18
Multiple Applications: Prioritize which one is the preferred one


carpodaster:
Student can hand in two applications, but it would be nice if they could state which of the two is their preferred one.

Any takers?

I mentally label this a nice-to-have one. :smile:

sareg0:
Hey @marcgreenstock do you think you have the time to take a gander at this one?

marcgreenstock:
sure, sounds simple.

@carpodaster just a `is_preferred` boolean column on the `application_drafts` table?
When updating, it sets the other associated rows to false or a referential column on the teams table?

carpodaster:
The simple boolean flag sounds good, @marcgreenstock – I was thinking about an integer field and a sorting plugin but since we only only two application, your approach is much better!

marcgreenstock:
Oh like a `priority` integer. That approach might be more scalable, in case you decide that more than two application drafts can be accepted. Besides we could use `acts_as_list` in this case and wash our hands of having to code up the update ourselves. I like your approach better :) I'll think about it over lunch and take care of it when i get back.

PR #185, opened 2015-04-07
Prioritize application drafts

marcgreenstock:
Resolves https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/168


app/views/application_drafts/index.html.slim
```diff
@@ -18,6 +19,12 @@ table.table.table-striped.table-bordered.table-condensed
   tbody
     - @application_drafts.each do |draft|
       tr
+        td
+          - if draft.position == 1
```
carpodaster: Does `draft.first?` work here?

app/views/application_drafts/index.html.slim
```diff
@@ -18,6 +19,12 @@ table.table.table-striped.table-bordered.table-condensed
...
+            i.icon-star title='Preferred application'
+          - else
+            a href=prioritize_application_draft_path(draft) title='Make preferred' data-method='put'
```
carpodaster: Will the `put` not be caught by the Rails UJS driver, firing an XHR? That way the user won't see the status change (the star icon). But I may be completely wrong on this :smiley_cat:

carpodaster:
Most excellent! :fireworks:

app/views/application_drafts/index.html.slim
marcgreenstock: Nah, it the jquery-ujs does a post rather than a xhr request, the controller then redirects as normal. So the change will be immediately be visible.

PR #188, opened 2015-04-07
Create application from draft: open submissions

### LINKED ISSUE #163: Application form: create an application from a valid draft

carpodaster:
So this fella finally allows the application drafts to be submitted. 

Closes #163 and also closes #174 (I cherry picked 9a56170696093a13d69d88bac1af05adacffdbf2).

I need another pair of eyes (or two pairs) on this – cc: @cypher @pxlpnk @marcgreenstock @svenfuchs @ramonh 

It adds/changes quite a few things (sorry for the huge patch):

* Added privacy for banking info and monetary requirements
* Allows mentors to browse application drafts of a team they're part of
* Removes a couple of hard attribute requirements from the `ApplicationDraft`
* Removes a few legacy attributes from `Application`
* Creates an application on `ApplicationDraft#submit_application`
* Snapshots the team members at the time of the application's submission

Missing:

* We need to hook in the mentor ping (Re: #144) but that I figured it's not a blocker for merging this PR.

Also: 

:boom: It bluntly copies all the data into the `Application#application_data` hstore column, completely ignoring the »real« columns that may exist. Due to the change in how we lodge applications now, **all the scoring logic** in `Application` has to be reviewed and/or adapted. I definitely need help with that!


app/controllers/application_drafts_controller.rb
```diff
@@ -55,6 +56,16 @@ def prioritize
     redirect_to application_drafts_url
   end
 
+  def apply
+    if application_draft.ready? && application_draft.submit_application
+      flash[:notice] = 'Your application has been submitted!'
+      ApplicationFormMailer.new_application(application_draft.application).deliver
```
marcgreenstock: Maybe use `.deliver_later` instead of `.deliver`. It uses the a background worker and is rails 4.2 compatible. Defined here: https://github.com/rails-girls-summer-of-code/rgsoc-teams/blob/master/config/initializers/monkey_patch_mail.rb#L6

app/controllers/application_drafts_controller.rb
```diff
@@ -111,14 +122,20 @@ def ensure_max_applications
...
+  end
+
   def current_team
-    current_student.current_team || coaches_team
+    current_student.current_team || extended_team
   end
 
-  def coaches_team
-    @coaches_team ||= begin
+  def extended_team
```
marcgreenstock: I am so lost with this method, I have no idea what's going on, could we consider a refactor?

app/models/creates_application_from_draft.rb
```diff
@@ -0,0 +1,70 @@
+class CreatesApplicationFromDraft
```
marcgreenstock: Me being picky, but could this be called `DraftToApplicationDelegator` or something to this effect? I don't think models should be named with verbs.

app/controllers/application_drafts_controller.rb
carpodaster: :+1:

app/controllers/application_drafts_controller.rb
carpodaster: We don't set the team explicitely and instead retrieve it via the `ApplicationDraft` instance (which we find with `params[:id]`). This method (together with `current_team` in line 131) wants to make sure that the team is discarded when the currently logged in user is not part of that team and that we have no start of the association chain when `application_draft` is looked up in line 71.

>  could we consider a refactor?

By all means YES. However, I would very much like to do this after this functional PR if you think this method does what it is supposed to do (see above).

app/models/creates_application_from_draft.rb
carpodaster: I've started naming classes like this a year or so ago and I like it for three reasons:

1. Classes ending on \*or or \*er (*Creator, *Finalizer, *Converter), to me, convey an overly ambitious attempt to be generic. This class has a very special purpose and doesn't aim to be generic at all.
2. I use the verb form because I want to model the _process_, as opposed to something _static_ (like a DB record)
3. I look at the name and I know exactly what it does, helping future versions of me sticking to its single responsibility :smile: 

That being said, I'm not completely against renaming it of course. I'm against the *Delegator suffix, though, and I don't have an idea for a noun-based class name that makes it equally clear what's going on in that object.

app/controllers/application_drafts_controller.rb
marcgreenstock: I get it, yeah, I didn't mean refactor now, just one day :)

app/models/creates_application_from_draft.rb
marcgreenstock: No problem, it's hard naming stuff.

carpodaster: That's a very diplomatic answer :wink:

carpodaster:
@marcgreenstock and @ramonh – you have looked into this PR (thanks a lot for that!). Do you think we are Go for launch?

ramonh:
Hm, I've been testing it manually and I'd say it works. One question, is it supposed to be possible to apply as a single student?

ramonh:
Oh it's merged, :)

carpodaster:
> One question, is it supposed to be possible to apply as a single student?

Yes that works. A pair of students are preferred, though, but that is part of the selection process.

ramonh:
Okay, makes sense, thanks!

ramonh:
Is it possible to test this on the staging app?

carpodaster:
We have a staging app but I haven't used it before. I'm afraid I don't have the time to set it up now :disappointed:

ramonh:
No problem! I'll look into it

PR #195, opened 2015-04-13
Sign off drafts

marcgreenstock:
It's a little rough and kind of rushed, but here it finally is, the last part of https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/144

### DIFF
--- a/app/models/application_draft.rb
+++ b/app/models/application_draft.rb
@@ -1,6 +1,4 @@
 class ApplicationDraft < ActiveRecord::Base
-  class CurrentUserNotProvided < StandardError; end
-
   include HasSeason
 
   include AASM
@@ -85,6 +83,7 @@ class ApplicationDraft < ActiveRecord::Base
   aasm :column => :state, :no_direct_assignment => true do
     state :draft, :initial => true
     state :applied
+    state :signed_off
 
     event :submit_application do
       after do |applied_at_time = nil|
@@ -94,30 +93,24 @@ class ApplicationDraft < ActiveRecord::Base
 
       transitions :from => :draft, :to => :applied, :guard => :ready?
     end
-  end
 
-  def signed_off?
-    signed_off_by.present? && signed_off_at.present?
-  end
+    event :sign_off, :guard => :can_sign_off? do
+      after do
+        update(
+          signed_off_by: current_user.id,
+          signed_off_at: Time.now.utc
+        )
+      end
 
-  def sign_off!
-    raise CurrentUserNotProvided unless current_user.present?
-    if !as_mentor?
-      errors.add :base, 'You need to be a mentor of this application to sign it off.'
-      false
-    elsif signed_off?
-      errors.add :base, 'Application has already been signed off.'
-      false
-    else
-      update(
-        signed_off_by: current_user.id,
-        signed_off_at: Time.now.utc
-      )
-      true
+      transitions :from => :applied, :to => :signed_off
     end
   end
 
   private
+
+  def can_sign_off?
+    current_user.present? and as_mentor?
+  end
 
   def mentor_required
     unless (team || Team.new).mentors.any?
--- a/app/models/application_draft.rb
+++ b/app/models/application_draft.rb
@@ -1,4 +1,5 @@
 class ApplicationDraft < ActiveRecord::Base
+  class CurrentUserNotProvided < StandardError; end
 
   include HasSeason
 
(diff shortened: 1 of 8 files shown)


marcgreenstock:
screen shot 2015-04-13 at 23 33 13

app/models/application_draft.rb
```diff
@@ -94,6 +96,27 @@ def ready?
...
+    elsif signed_off?
+      errors.add :base, 'Application has already been signed off.'
+      false
+    else
+      update(
+        signed_off_by: current_user.id,
+        signed_off_at: Time.now.utc
+      )
+      true
+    end
```
carpodaster: Maybe just return `errors.present?` instead of the explicit bools?

carpodaster:
:+1: @marcgreenstock 

@cypher made `ApplicationDraft` a finite state machine using AASM. Do you think you can make `#sign_off!` an event / transition just like `#submit_application´ ? If not, we can always add that after the merge (@cypher maybe? :) )

Do preserve the students' application data and protect the application from indirectly be modified after it has been submitted, `#submit_application` creates an `Application´ record.

It would be best to also update the associated application with the signoff data since the selection committee will only operate on applications, not application drafts. Adding them to the `Application#application_data` hstore hash is sufficient, I think.

cypher:
I can certainly help out with this once I'm home from work :)

marcgreenstock:
I'll be back at my computer in a few hours and can take care of making signed off a state and fix up any suggestions.

cypher:
@marcgreenstock I've written a quick patch that should demonstrate how using the state machine for this case would work, but I haven't yet updated the tests.

(To apply the patch, copy it into a file, then run `git apply patchfile`, or if you're on OS X, copy it, then run `pbpaste | git apply`)

```patch
diff --git i/app/models/application_draft.rb w/app/models/application_draft.rb
index c466ac8..cf46c1a 100644
--- i/app/models/application_draft.rb
+++ w/app/models/application_draft.rb
@@ -85,6 +85,7 @@ class ApplicationDraft < ActiveRecord::Base
   aasm :column => :state, :no_direct_assignment => true do
     state :draft, :initial => true
     state :applied
+    state :signed_off
 
     event :submit_application do
       after do |applied_at_time = nil|
@@ -94,29 +95,28 @@ class ApplicationDraft < ActiveRecord::Base
 
       transitions :from => :draft, :to => :applied, :guard => :ready?
     end
-  end
 
-  def signed_off?
-    signed_off_by.present? && signed_off_at.present?
-  end
+    event :sign_off, :guard => :can_sign_off? do
+      after do
+        update(
+          signed_off_by: current_user.id,
+          signed_off_at: Time.now.utc
+        )
+      end
 
-  def sign_off!
-    raise CurrentUserNotProvided unless current_user.present?
-    if !as_mentor?
-      errors.add :base, 'You need to be a mentor of this application to sign it off.'
-      false
-    elsif signed_off?
-      errors.add :base, 'Application has already been signed off.'
-      false
-    else
-      update(
-        signed_off_by: current_user.id,
-        signed_off_at: Time.now.utc
-      )
-      true
+      transitions :from => :applied, :to => :signed_off
     end
   end
 
+  def can_sign_off?
+    raise CurrentUserNotProvided unless current_user.present?
+
+    errors.add :base, 'You need to be a mentor of this application to sign it off.' if !as_mentor?
+    errors.add :base, 'Application has already been signed off.' if signed_off?
+
+    return errors.present?
+  end
+
   private
 
   def mentor_required


```

marcgreenstock:
Thanks @cypher. One question for you and/or @carpodaster, what is the expected flow of state?

Should the mentor sign off first before the application is submitted or after? Currently with @cypher's patch the can not transition from `draft` to `signed_off`. Is this the expected behaviour?

cypher:
As far as I know/understand, yes, that's the expected flow.

carpodaster:
The flow should be `draft -> applied -> signed_off`. The application drafts we're after are all in the `applied` state and an `Application` record has already been created for them (to freeze and protect them from modification).

marcgreenstock:
@cypher, @carpodaster, cool thanks, just fixing the tests then.

carpodaster:
Morning @marcgreenstock – any progress with this? Can I help?

marcgreenstock:
hi @carpodaster sorry, i know this was urgent, it's now complete i think, can you take a look?

cypher:
@marcgreenstock Thanks! Everything looks good to me, so I merged it.

Issue #206, opened 2015-04-17
Applications view 


anikalindtner:
It's been hard to read the applications, especially with the fields not showing the questions we asked, but just some field_name-thingy. I'd love it if we could make it this year a bit more readable and easier to rate in showing the exact question we asked the teams above their answers. For now it looks like this:

![screen shot 2015-04-17 at 18 25 40](https://cloud.githubusercontent.com/assets/1711357/7206225/8a39ccc4-e52f-11e4-8bce-d639e000436e.png)

I guess there is probably an easy way to switch out the "project_name" with "What project are you planning to work on".

Seems to me that we could jut exchange the "f.input :," with the "label:" and it should show the questions instead of the weird description?

carpodaster:
The way the data is presented is super lazy: The view just iterates over whatever field exists in the application and displays its name together with its value. That's actually why there was little to no adaption needed even though we virtually turned the application process upside down.

There is no way of automatically infering the human-readable question without hard-coding and duplicating the field -> question association. It's certainly not rocket science, but I wouldn't do it for said duplication / hard coupling reasons.

What I can offer is using [this translation table](https://github.com/rails-girls-summer-of-code/rgsoc-teams/blob/master/config/locales/en.yml#L13) which is used to prettify error messages on an `ApplicationDraft`. Would that help / solve it?

ramonh:
Thanks, @carpodaster! That was very helpful

carpodaster:
Done in #208

PR #207, opened 2015-04-18
Display proper labels for Application data

ramonh:
This adds a constant to the `Application` model that defines the labels that will be displayed when viewing an `Application`.

The `Application#show` view will now display these labels.

This aims to address #206

### DIFF
--- a/app/models/application.rb
+++ b/app/models/application.rb
@@ -52,6 +52,10 @@ class Application < ActiveRecord::Base
         location: 'City'
   }
 
+  def self.data_label(key)
+    APPLICATION_DATA_LABELS[key] ? APPLICATION_DATA_LABELS[key] : human_attribute_name(key)
+  end
+
   def name
     [team.try(:name), project_name].reject(&:blank?).join ' - '
   end
--- a/app/models/application.rb
+++ b/app/models/application.rb
@@ -18,6 +18,40 @@ class Application < ActiveRecord::Base
   scope :hidden, -> { where('applications.hidden IS NOT NULL and applications.hidden = ?', true) }
   scope :visible, -> { where('applications.hidden IS NULL or applications.hidden = ?', false) }
 
+  APPLICATION_DATA_LABELS = {
+        project_url: 'Project URL',
+        coaches_hours_per_week: "Coaches' Hours-per-Week",
+        coaches_why_team_successful: "Coaches' Success-Statement",
+        student0_application_about: 'Background (1st student)',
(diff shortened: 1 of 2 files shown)


app/models/application.rb
```diff
@@ -18,6 +18,44 @@ class Application < ActiveRecord::Base
...
+        student1_banking_info: 'Banking Information (2nd student)',
+        student1_application_minimum_money: 'Minimum Money (2nd student)',
+        paid: 'Paid?',
+        rgsoc_only: 'Exclusively for RGSoC Alumni?',
+        misc_info: 'Additional information',
+        location: 'City'
+  }
+
+  def self.data_label(key)
+    APPLICATION_DATA_LABELS[key] ? APPLICATION_DATA_LABELS[key] : human_attribute_name(key)
```
carpodaster: I would rather not copy the field translation copied to a constant (would require editing the semantics in two different places again). An idea would be to change the method body:

```ruby
def self.data_label(key)
  ApplicationDraft.human_attribute_name(key) ? ApplicationDraft.human_attribute_name(key) : human_attribute_name(key)
end
```
but unfortunately `human_attribute_name` always returns a value.

OTOH, we only need this in the iteration in ` app/views/applications/show.html.slim:38`, so technically we never access direct database column attibutes of `Application`. We could use that one:

```ruby
def self.data_label(key)
  ApplicationDraft.human_attribute_name(key)
end
```

What dou you think?

ramonh:
Thanks for the feedback!

I'm not sure I understood correctly, but calling `human_attribute_name` seems to return the 'English-ified' key. So for example:
```ruby
ApplicationDraft.human_attribute_name(:misc_info)
```
would return 'Misc info', which is not quite what we want.

Have I misunderstood?

ramonh:
Quick follow-up:

I've looked into it, and calling `human_attribute_name` should be returning the values from the locale file. I'll look into why this isn't happening.

ramonh:
Ok now I got everything I need to get going. Thanks for clarifying, @carpodaster!

carpodaster:
For "normal" columns / attributes, that's correct. It looks up in the locales files and if it doesn't find something specific it just calls `titleize` on the column name, I think.

However, all application draft attributes are saved in the `application_data` hash within the `Application` record. That hash what we're iterating over and what's displayed in the show view. Since it's hash-data we cannot use the locale file.

ramonh:
Got it - Thanks @carpodaster! Will be closing this to rewrite things.

PR #212, opened 2015-04-20
added mentors to list of mailable roles

sareg0:
This is to add the mentors role to the list of available roles to send an email to.

specs were run, no failures.


carpodaster:
Looks good!

PR #213, opened 2015-04-21
added 2 tests for Student Controller

sareg0:
This is to see if the stuff we are writing is what is needed, and also just to push things in small chunks.

We tried to write a test that tested something along the lines of "assigns the @rating to user.ratings", but we found the rating Model pretty tricky to understand. 

Anywho. Enjoy! Feedback welcome.

### DIFF
--- a/spec/controllers/applications/students_controller_spec.rb
+++ b/spec/controllers/applications/students_controller_spec.rb
@@ -1,6 +1,8 @@
 require 'spec_helper'
 
 describe Applications::StudentsController, type: :controller do
+  render_views
+
   describe "GET show" do
     let(:current_user) { FactoryGirl.create(:reviewer) }
     let(:user) { FactoryGirl.create(:user) }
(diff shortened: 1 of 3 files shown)


spec/controllers/applications/students_controller_spec.rb
```diff
@@ -0,0 +1,29 @@
+require 'spec_helper'
+
+describe Applications::StudentsController, type: :controller do
```
carpodaster: can you add a `render_views` to it? That way we have a regression test for errors in the templates.

carpodaster:
> we found the rating Model pretty tricky to understand. 

Yeah, same :( That's why I hoped adding tests/specs would shed some light on its background magic

sareg0:
okay, well I'll keep working on it and will continue adding to this pull request. I suspect I may need to have a meeting with the magician :)

sareg0:
i have added a few more tests, but there are two failures in the previous spec files. I just thought I would commit piecemeal, and perhaps someone has some feedback?

spec/factories/ratings.rb
```diff
@@ -1,6 +1,6 @@
 FactoryGirl.define do
   factory :rating do
     user
-    application
+    association :rateable
```
carpodaster: This is were the test break. I think it can be fixed with `association :rateable, factory: :application`, but then again, I don't know why it is a polymorphic association. We only rate applications, right?

carpodaster:
Hey @sareg0, are you still working on this? I was thinking: since we talked about rewriting the rating engine for next year, maybe it's ok to discontinue and close this?

carpodaster:
@sareg0 I'm closing this. Rumor has it we will rewrite the selection sub-app completely :wink:

PR #216, opened 2015-04-28
Group todo page

marcgreenstock:
@sareg0 told me the todo page was a bit tricky to use so I thought I would help here.

This should group the students and applications with their teams, and show items that still need rating.

I've not loaded this in my browser cause I find seeding the db with what I need tricky but the tests pass :)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -48,6 +48,7 @@ group :development, :test do
   gem 'factory_girl_rails'
   gem 'database_cleaner'
   gem 'byebug'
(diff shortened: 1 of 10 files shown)


sareg0:
@carpodaster is there anything to improve before this one could be merged?

carpodaster:
I only had a brief look myself and didn't test it in my browser either. But if it compiles, ship it! (yolo?) :wink: Thanks, @marcgreenstock !

PR #218, opened 2015-04-30
Dont show teams without apps in todos

marcgreenstock:
I understand the todos page shouldn't show teams that don't have applications. This PR should resolve that.

Note: I'm using a counter_cache of applications on teams. While it is possible to use a join to count, it's terribly inefficient.

### DIFF
--- a/app/controllers/applications/todos_controller.rb
+++ b/app/controllers/applications/todos_controller.rb
@@ -6,6 +6,8 @@ class Applications::TodosController < ApplicationController
   include TodoHelper
 
   def index
-    @teams = Team.includes(:students, :applications)
+    @teams = Team.joins(:applications)
+                 .where('"applications_count" > 0')
+                 .includes(:students, :applications)
   end
 end
--- a/app/models/application.rb
+++ b/app/models/application.rb
@@ -52,7 +52,7 @@ class Application < ActiveRecord::Base
   include HasSeason
 
   belongs_to :application_draft
-  belongs_to :team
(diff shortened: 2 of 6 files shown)


app/models/application.rb
```diff
@@ -52,7 +52,7 @@ def sort(data)
   include HasSeason
 
   belongs_to :application_draft
-  belongs_to :team
+  belongs_to :team, inverse_of: :applications, counter_cache: true
```
carpodaster: Is inverse_of still a thing? Isn't it deprecated?

app/controllers/applications/todos_controller.rb
```diff
@@ -6,6 +6,8 @@ class Applications::TodosController < ApplicationController
   include TodoHelper
 
   def index
-    @teams = Team.includes(:students, :applications)
+    @teams = Team.joins(:applications)
```
carpodaster: I _think_ it works without the explicit `joins(:applications)`

carpodaster:
We need a `counter_cache: true` in `Team` on the has_many applications to make the auto-updating on the counter cache work, no?

marcgreenstock:
@carpodaster `counter_cache` is an option on the `belongs_to` http://guides.rubyonrails.org/association_basics.html#counter-cache

carpodaster:
Ah, I totally overlooked it in the PR. I only had eyes for the `inverse_of` :wink:

sareg0:
can we merge this baby in now @carpodaster ?

sareg0:
super sweet! Thanks @carpodaster

PR #222, opened 2015-05-08
show monies needed on app table

marcgreenstock:
This will replace the coaching company column with the monies needed. Making it easier to review applications.

/cc @sareg0


carpodaster:
:moneybag:

Issue #232, opened 2015-06-18
make mailer checkbox available for selected teams only


sareg0:
This is something requested by @anikalindtner, to make it possible to email just selected teams

Currently there is a 'teams' checkbox in the /mailers area, but this would email all teams; selected AND rejected.

@marcgreenstock has volunteered to take a look at this, but if someone else wants to grab it, that is also fine.

It is time sensitive. We would like it to be implemented as soon as possible.

PR #235, opened 2015-06-24
created a group option for the mailings

marcgreenstock:
This adds radio buttons to mailings#new, that allows "Everyone", "Selected Teams" and "Unselected Teams".

I think this should satisfy https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/232

### DIFF
--- a/app/models/recipients.rb
+++ b/app/models/recipients.rb
@@ -1,5 +1,5 @@
 class Recipients
-  delegate :group, :to, :cc, :bcc, to: :mailing
+  delegate :seasons, :group, :to, :cc, :bcc, to: :mailing
 
   attr_reader :mailing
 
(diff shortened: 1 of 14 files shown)


app/models/recipients.rb
```diff
@@ -16,7 +16,16 @@ def user_emails
   end
 
   def users
-    User.joins(:roles).where( roles: {name: roles}).uniq
+    case group.to_sym
+    when :selected_teams
+      User.joins(:teams, :roles)
+        .where(teams: { kind: Team::KINDS }, roles: { name: roles }).uniq
```
carpodaster: Do we need a season classifier? It will pull the currently selected teams for this year allright, but next year it will also include the almuni teams.

```patch
- .where(teams: { kind: Team::KINDS }, roles: { name: roles }).uniq
+ .where(teams: { kind: Team::KINDS, season: Season.current }, roles: { name: roles }).uniq
```

Or add another enum value that explicitly only selects the current year's students?

carpodaster:
Cool, @marcgreenstock - looks good! :rocket: I've commented on a season selector.

app/models/recipients.rb
marcgreenstock: I had a suspicion this would be needed, but I wasn't sure. I'm on it!

marcgreenstock:
I've added a seasons option to the mailings too now. I realise I should add some tests around it but i'm short on time. Let me know if you think this is required and I'll retrospectively add some.

carpodaster:
If you can spare the time, tests would be good. If for no other reason than to avoid regressions in the future.

marcgreenstock:
@carpodaster just added some specs

carpodaster:
@marcgreenstock awesome, looks super. I didn't know about FactoryGirl's `transient` feature, thanks :smile: Merging! :shipit:

PR #243, opened 2015-07-02
Only send checkup comments to rgsoc email

### LINKED ISSUE #238: Checked up on them: add summer-of-code email to recipients
Right now, when a supervisor comments on their team's progress, an email update is sent to all supervisors – but not to orgas. We should change that so the email is also sent to the summer-of-code email address (either as direct recipient or cc).

carpodaster:
Now fo' realz: Closes #238.

@alicetragedy could you check if I haven't made another mistake please? :smile:


alicetragedy:
looks pretty good to me :sunny: 
I think you can yolomerge!

PR #272, opened 2015-07-14
Seeding dev database with users  

F3PiX:


F3PiX:
OK @carpodaster ?

carpodaster:
@F3PiX I'm thinking that it would supply more value to have the users arranged in teams (where applicable) since teams (both sponsored/voluntary and unaccepted) is what we're working with mostly. Or do you plan to add it in an upcoming PR?

F3PiX:
@carpodaster I do indeed!

Issue #244, opened 2015-07-02
Add pagination to users#index


carpodaster:
The users' list at "Community" is growing long.

F3PiX:
:+1:

PR #273, opened 2015-07-14
Adding pagination to Community / users list

F3PiX:
OK @carpodaster ?


F3PiX:
Oow I forgot to mention: issue #244

config/initializers/kaminari_config.rb
```diff
@@ -0,0 +1,10 @@
+Kaminari.configure do |config|
+   config.default_per_page = 15
```
carpodaster: This will affect the whole app. Kaminari's default is 25 per page and I would rather stick with that.

spec/views/users/index.html.slim_spec.rb
```diff
@@ -1,18 +0,0 @@
...
-    ])
-    allow(view).to receive(:current_season).and_return(build_stubbed(:season))
-    allow(controller).to receive(:current_user).and_return(double.as_null_object)
-  end
-
-  it 'renders a list of users' do
-    render
-    assert_select 'tr>td', text: 'Name', count: 2
-  end
-end
```
carpodaster: Removing a test file because it fails is normally not an option :wink: but we talked about it already: view template tests are brittle anyway so :+1: for throwing it away.

config/initializers/kaminari_config.rb
F3PiX: Done!

spec/views/users/index.html.slim_spec.rb
F3PiX: LOL

carpodaster:
Looks good! Thanks @F3PiX :tulip:

Issue #289, opened 2015-07-21
Layout error in header


roosdebildt:
Created huge horizontal scrolling area.
All huge styling changes create bugs, right *embarrassed*.... right?

![screen shot 2015-07-21 at 18 05 12](https://cloud.githubusercontent.com/assets/11146524/8805742/34a60edc-2fd3-11e5-9f8e-2e8595b8dff1.png)

+ this:

![screen shot 2015-07-21 at 18 08 18](https://cloud.githubusercontent.com/assets/11146524/8805822/99eea312-2fd3-11e5-9dd9-e8981b3dd209.png)

Easy fixes though. I'm on it.

alicetragedy:
@roosdebildt no worries, tbh I didn't even notice :speak_no_evil:

sareg0:
hehe, I noticed, but it was already there, it was just more acute after the layout changes. You didn't cause, you just made it more obvious it needed to be fixed :) Glass half full ;)

alicetragedy:
closed via https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/290

PR #290, opened 2015-07-21
header and font bugfixes #289

roosdebildt:
see #289.


sareg0:
this is cool, it has been a problem we haven't gotten around to fixing until now.

Issue #267, opened 2015-07-13
Supervising area (team profile page) layout error


idengager:
So I put in a long long comment together with the first checkup for my team ([Rubyherzlein](https://teams.railsgirlssummerofcode.org/teams/9)) and this happened:
![screen shot 2015-07-13 at 12 07 34](https://cloud.githubusercontent.com/assets/2712383/8647459/628c2eb6-2958-11e5-91a9-35039852a59b.png)

I wouldn't mind it as such, but the comment covers the `Checked up on them button` so that it can't be clicked. Because this is just a layout issue I was able to work around it by just hiding the `information` element, but, well, this isn't sustainable.

alicetragedy:
@idengager thanks for opening the issue! I actually noticed this on the weekend as well and was planning to fix it ;) I'll get that done sometime this week – hopefully sooner rather than later!

sareg0:
ps I'm looking into this

F3PiX:
@sareg0 We will be working on the supervisor dashboard from Wednesday on. Just saying... (but you know how fast we are, so if it needs a quick fix ;-)... )

sareg0:
believe you me, this is a Quick fix (not the capital Q ;) ), and the supervisor's dashboard needs an overhaul anyway. This fix is just a stopgap, so feel free to chuck it out when you get to the dashboard. It's just so it's not so annoying right now @F3PiX

sareg0:
Pull request now lodged

alicetragedy:
closed via https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/291

F3PiX:
We will do our best to make it a wonderfull dashboard for all SUPERvisors :-)

PR #291, opened 2015-07-21
fix text flowing over supervisors' check up box

sareg0:
This is a quick fix for issue #267.

Below is a the screenshot of what it will look like when merged.

supervisorcomments

Note that there is now a scroll, to stop the text flowing out of the 'y' bounds of the grey box. To make reading easier, the text is now aligned left instead of right. It creates two more distinct columns. By no means perfect, and this may be temporary before Team Cheesy gets to the supervisors' dashboard.

All tests passing still.

### DIFF
--- a/app/assets/stylesheets/global.css.sass
+++ b/app/assets/stylesheets/global.css.sass
@@ -1,5 +1,5 @@
 // Global styles
-// 
+//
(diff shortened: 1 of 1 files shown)


alicetragedy:
looks good to me – I'm happy to merge :)

PR #296, opened 2015-07-28
save error and sent at on submissions

marcgreenstock:
In the upgrade to 4.2 I think I removed this: https://github.com/rails-girls-summer-of-code/rgsoc-teams/commit/1f4bb2ca03c87e34e681301fd24efb091bc02a4f#diff-a96b7164d6c7405c9b73329852bb415dL15

That was what was persisting the error and setting the sent_at on the submissions.


carpodaster:
:tada: Thanks!

PR #298, opened 2015-08-09
fix sending of emails to teamless users

marcgreenstock:
This one was hard. When joining two roles and teams together, AR ensures that the records are connected and excludes records that are not connected via all three models. So in the case of organizers, not belonging to a team via their `organizer` associated role excluded them from the recipients.

Although I have added some more test, there are many permutations I have not considered, so I doubt this will be the last time we will need to do something with the mailer.

### DIFF
--- a/app/models/recipients.rb
+++ b/app/models/recipients.rb
@@ -17,23 +17,32 @@ class Recipients
 
   def users
     return @users if @users
-    @users = User.joins(:teams, :roles).where(roles: { name: roles })
-    if group.to_sym == :selected_teams
-      @users = @users.where(teams: { kind: Team::KINDS })
-    elsif group.to_sym == :unselected_teams
-      @users = @users.where(teams: { kind: nil })
-    end
-    if seasons.any?
-      @users = @users.where(teams: {
-        season: Season.where(name: seasons).pluck(:id)
-      })
-    end
-    @users.uniq
+    @users = User.joins(:roles).where(roles: { name: roles })
+    @users = @users.where('roles.team_id IN (?) OR roles.name IN (?)', teams, teamless_roles) if filter_by_teams?
+    @users = @users.uniq
+    @users
+  end
+
+  def filter_by_teams?
+    group.to_sym != :everyone || seasons.any?
+  end
+
+  def teams
+    return @teams if @teams
+    @teams = Team
+    @teams = @teams.where(kind: group.to_sym == :selected_teams ? Team::KINDS : nil) if group.to_sym != :everyone
+    @teams = @teams.where(season: Season.where(name: seasons)) if seasons.any?
+    @teams = @teams.select(:id).pluck(:id)
+    @teams
   end
 
   def roles
     Array(to).map do |to|
       to == 'teams' ? %w(student coach mentor) : [to.singularize]
     end.flatten
+  end
+
+  def teamless_roles
+    (roles & %w{organizer developer helpdesk})
   end
 end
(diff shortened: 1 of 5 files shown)


sareg0:
hey there @carpodaster could you please review this pull request if you can find the time? I know that @anikalindtner would really need this fix to complete a task of her's

carpodaster:
@sareg0 will have a look into this later today

sareg0:
:open_hands: thank you!

app/models/recipients.rb
```diff
@@ -17,18 +17,23 @@ def user_emails
...
+  end
+
+  def filter_by_teams?
+    group.to_sym != :everyone || seasons.any?
+  end
+
+  def teams
+    return @teams if @teams
+    @teams = Team
+    @teams = @teams.where(kind: group.to_sym == :selected_teams ? Team::KINDS : nil) if group.to_sym != :everyone
```
carpodaster: `kind: nil` selects every user in the system that isn't part of a team that hasn't been accepted as either sponsored or voluntary team ever, including all applicants whose applications had been rejected.  Is that the intended behaviour?

app/models/recipients.rb
```diff
@@ -17,18 +17,23 @@ def user_emails
...
+    @teams = @teams.where(season: Season.where(name: seasons)) if seasons.any?
+    @teams = @teams.select(:id).pluck(:id)
```
carpodaster: Is the `select(:id)` needed?

app/models/recipients.rb
carpodaster: So that's equivalent to %w{organizer developer helpdesk} + :unselected_teams

app/models/recipients.rb
marcgreenstock: No, but only the id is needed, just conserving a little memory here.

carpodaster: Yep, I just thought that's what `pluck :id` already does. If not, don't mind me

app/models/recipients.rb
marcgreenstock: So this is a little tricky to understand, it took me a while too.

There are a number of filters, `group`, `seasons` and `to`.

If `group` is `:selected_teams` it will select any team that has a `kind` that is included in `Team::KINDS`, otherwise select where `kind` is specifically `nil`. `group: :everyone` will select every team.

It gets more tricky with [line 21](https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/298/files#diff-875df54ccb6e747fd165f1ebcf5d3266R21)

```ruby
@users = @users.where('roles.team_id IN (?) OR roles.name IN (?)', teams, teamless_roles) if filter_by_teams?
```

where if we need to filter by team it selects users who's role belongs to a team or users who's roles is teamless (organizer developer helpdesk) and has been included in the `to` filter.

I have considered rewriting the spec for `recipients.rb`, it sort of needs it because there are a lot of permutations that are not considered in the spec.

app/models/recipients.rb
marcgreenstock: I might have got this wrong, but when i was looking at it, pluck was not deselecting every other column.

carpodaster:
@marcgreenstock It definitely looks hard, yes. Thanks for the work on it!

>  there are many permutations I have not considered, so I doubt this will be the last time we will need to do something with the mailer.

Let's just merge and see what happens #yolo.  It looks good, but as you said: we may have overlooked a permutation.

Feel free to merge!

marcgreenstock:
@carpodaster, you want to deploy or shall I?

carpodaster:
@marcgreenstock It should auto-deploy through Travis

marcgreenstock:
@carpodaster yay for CI :)

PR #300, opened 2015-08-17
Moving the new status update form into index page

F3PiX:
This replaces the new status update button. This puts the form above the updates listing. Now, students can see and read their previous update while typing the new update. 

Also a bit of formatting and refactoring and a bit of ux tweaking


app/controllers/students/status_updates_controller.rb
```diff
@@ -55,4 +53,12 @@ def status_update_params
     params.require(:activity).permit(:title, :content)
   end
 
-end
+
+  private
+
+  def status_updates
+    #order = DESC; set in model
+    @status_updates = current_team.status_updates.ordered
```
carpodaster: This could do with a `||=` assignment (memoization). That way, subsequent calls to `status_updates` won't trigger another database request. It's only called once in the view so it doesn't matter with the current implementation. But since it's being exposed as a helper method, it's better to not care how or how often a view template will call this.

app/views/students/status_updates/index.html.slim
```diff
@@ -15,13 +16,12 @@ table.table.table-striped.table-bordered.table-condensed
       th
 
   tbody
-    - @status_updates.each do |status_update|
+    - status_updates.each do |status_update|
       tr
-        td = l status_update.created_at
+        /date format is Weekday, MM dd, hh:mm
+        td = l status_update.created_at, :format => ('%A, %B %e, %H:%M (utc)')
```
carpodaster: Can you use or define a date/datetime format in `config/locales/en.yml`? That way it can be re-used and we don't scatter the app with too many different ways of printing date formats.

carpodaster: Also, we can use local time: every user can define a timezone in their user profile. The Rails app will set the individual time zone automatically or default to UTC.

carpodaster:
Hey @F3PiX – looks good! I added two comments – once these are addressed, we can meerrrrrrge.

F3PiX:
Up to you, @carpodaster !  Thanks, on behalf of all the students ;-)

carpodaster:
Thanks for the PR! :smile:

PR #302, opened 2015-08-22
Applications state machine

F3PiX:
First try to a smoother workflow with the application form.
I added an extra state 'final' to the state machine, in between 'draft' and 'applied'. 
The new 'final' state marks the end of filling in the form (by both students and coaches). 'Final' starts a validation check and makes it clear that only the students are allowed to submit the final draft.
In the old situation, all the validations and notices were packed within the 'draft' state. 
I hope the extra state will make it easier to guide the users through the process of the application form. 
  
I am not totally happy with the states and transitions yet, but I am thinking about it, and explore the state machines possibilities. 
Before we go into that, I have a few questions about things I encountered.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -45,7 +45,9 @@ class ApplicationController < ActionController::Base
   end
 
   def current_season
-    @season ||= Season.current
+    #TEMP solution to find the application form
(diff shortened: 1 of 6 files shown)


F3PiX:
SORRY wrong Master !

PR #304, opened 2015-09-03
Season switch

F3PiX:
This switch enables developers to easily change the season's phase, to work in date dependent settings.


app/controllers/orga/seasons_controller.rb
```diff
@@ -44,6 +46,22 @@ def destroy
...
+  def switch_phase
+    return unless Rails.env.development?
+    @season = current_season
+    case params[:option]
+    when 'Application'
+      @season.fake_application_phase
+    when 'CodingSummer'
+      @season.fake_coding_phase
+    when 'RealTime'
+      @season.back_to_reality
```
carpodaster: Would it work to just delete the season? A new season should be created automatically. New seasons start out with default values for the current year so that would remove the requirement of `Season#back_to_reality` entirely.

config/routes.rb
```diff
@@ -71,9 +71,13 @@
 
   namespace :orga do
     resources :teams
-    resources :seasons
+    resources :seasons do
+      member do
+      get :switch_phase
```
carpodaster: Indentation

app/views/orga/seasons/index.html.slim
```diff
@@ -27,3 +27,5 @@ table.table.table-striped.table-bordered.table-condensed
 br
 
 = link_to 'New Season', [:new, :orga, :season], class: 'btn btn-default'
+
+= render partial: 'orga/seasons/switch' if Rails.env.development?
```
carpodaster: Oh, better yet: `unless Rails.env.production?`

app/controllers/orga/seasons_controller.rb
```diff
@@ -44,6 +46,22 @@ def destroy
     redirect_to orga_seasons_path, notice: "Season #{@season.name} has been deleted."
   end
 
+  # In dev env only, this offers a setting to switch parts of season to open / close
+  # Application opens the link to the application form; CodingSummer opens the link to students/status_updates
+  def switch_phase
+    return unless Rails.env.development?
```
carpodaster: `return if Rails.env.production?`

app/views/orga/seasons/_switch.html.erb
```diff
@@ -0,0 +1,17 @@
...
+    <%= label_tag :option_CodingSummer, "Coding Summer" %>

+    <%= radio_button_tag :option, "RealTime" %>
+    <%= label_tag :option_RealTime, "Back to Real Time" %>
+

    <%= submit_tag("Switch phase") %>
+  <% end %>
+
+<p><%= " current_season set to #{params[:option]} " %><br>
```
carpodaster: Looks like remainders of some debugging output

app/views/orga/seasons/index.html.slim
F3PiX: Are you sure? Wouldn't that cause trouble in test env with the tests if somebody leaves the season setting in a time travelled phase?

F3PiX:
Yay! Now the form is submitted with :patch, with path_helper in place and helper_method gone ( @nathanvda ). And a redirect to avoid the 'missing template' ( @YorickPeterse ).

app/controllers/orga/seasons_controller.rb
F3PiX: I tried it. If I only change :back_to_reality to :destroy, the ux for the developers would require more work. I can look into that in the upcoming weeks, in between the sprints. Can you live with this as is? For now?

app/views/orga/seasons/_switch.html.erb
F3PiX: It wasn't meant for debugging, but as simple user action feedback. :-)  Already changed that into a nice flash notice. I left the season.inspect though.

marcgreenstock:
I really like this idea, however I wonder if there is a simpler implementation such as using [Timecop](https://github.com/travisjeffery/timecop) and an environment variable... something like this in an initializer maybe:

```Ruby
if ENV['SEASON'].present?
  Timecop.travel(DateTime.now.change(year: ENV['SEASON'].to_i))
end
```

You would then start up the dev environment with `SEASON=2018 rails s` and every object that extends off Time will think the year is 2018.

F3PiX:
My assignment by the coaches was: not to use Timecop :-)

marcgreenstock:
@F3PiX ahh cool, no drama's then. I'll review the code with the intended implementation then.

marcgreenstock:
I've looked at the code and it's good in my option, but I feel slightly uncomfortable about the functionality. It's introducing environment conditional functionality, this stuff is super hard to test and debug. e.g. something is broken in production but you simply can't reproduce it in development.

I understand the need, and I think it is very helpful. Perhaps you can chat with your coaches and relay my concerns.

On a final note, open source is a democracy, so if I'm alone in this opinion then just get a few more +1's on this and i'll merge. @carpodaster I know your super busy but wdyt?

app/views/orga/seasons/index.html.slim
carpodaster: The test env is exactly what I had in mind: checking for developement only would skip that portion during our tests, meaning you could literally place a `fail` statement in the template without noticing it running the specs.

Every test should run on a clean test DB. If we find the tests are randomly failing, we may have spotted an example leaking data and/or a problem with the database cleansing.

app/controllers/orga/seasons_controller.rb
carpodaster: Sure, we can roll with it for now. Can you add an `OPTIMIZE` comment to that line?

carpodaster:
@marcgreenstock Not super-busy, my internet died :(  I totally share your opinion about hard-coding environment-checks but I'd rather have something _not_ work in production than _only_ in production.

I like the idea of Timecop, but never used it other than for testing. Maybe this is something we can consider to refine it since it would allow us to really warp into another season whereas this PR changes the _current_ season to a different phase. The Timecop Season Warp™ would still require a phase switcher to get into the proper phase we would like to see ourselves in.

I'm +1 for a merge.

carpodaster:
As discussed with @F3PiX: merging this to allow our RGSoC to move along. @F3PiX Can you note the comments made and tackle them post-merge?

PR #307, opened 2015-09-07
Supervisor's dashboard

F3PiX:


app/controllers/supervisor/base_controller.rb
```diff
@@ -0,0 +1,18 @@
...
+  protected
+
+  def must_be_supervisor
+    unless signed_in? && current_user.roles.includes?('supervisor')
+      redirect_to root_path, alert: 'Sorry, the dashboard is for supervisors only'
+    end
+  end
+
+  def supervised_teams
+    @supervised_teams = current_user.teams.where("roles.name = 'supervisor'")
```
marcgreenstock: It's picky and a bit of personal preference but I suggest using ActiveRecord's sql building syntax:
   
```ruby
@supervised_teams = current_user.teams.where(roles: { name: 'supervisor' })
```

It produces the almost the same SQL but without the hassle of writing any of it.

app/controllers/supervisor/dashboard_controller.rb
```diff
@@ -0,0 +1,6 @@
+class Supervisor::DashboardController < Supervisor::BaseController
+
+  def index
```
marcgreenstock: This method is unnecessary, it should be inferred by the view.

app/views/supervisor/dashboard/index.html.slim
```diff
@@ -0,0 +1,16 @@
+h1 = "Hi #{current_user.name}!"
+h2 This is your brand-new Supervisor Dashboard - with love from Team Cheesy
+p alpha version
+
+hr
+
+div.clearfix
+  - supervised_teams.each do |t|
+      div.dashboard-box
```
marcgreenstock: Is there one too many indentations here?

app/controllers/supervisor/dashboard_controller.rb
```diff
@@ -0,0 +1,6 @@
+class Supervisor::DashboardController < Supervisor::BaseController
```
marcgreenstock: It's better practice to write the module explicitly:

```ruby
module Supervisor
  class DashboardController < Supervisor::BaseController
end
```

app/controllers/supervisor/base_controller.rb
```diff
@@ -0,0 +1,18 @@
+class Supervisor::BaseController < ApplicationController
```
marcgreenstock: It's better practice to write the module explicitly:

```ruby
module Supervisor
  class BaseController < ApplicationController
end
```

marcgreenstock: Ideally this controller should be tested, especially the `must_be_supervisor` method.

marcgreenstock:
Really great work, I struggled to find much if anything wrong with it at all, I have added some suggestions, but they are more best practice suggestions more than anything.

app/controllers/supervisor/base_controller.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  before_action :must_be_supervisor
```
marcgreenstock: I just realised/remembered this project uses the [cancancan](https://github.com/CanCanCommunity/cancancan) gem for exactly this kind of purpose. Could you possibly try and use that to authorize supervisors? If you get stuck, feel free to ping me.

app/views/activities/index.html.slim
```diff
@@ -3,6 +3,10 @@ h1.header
   span Activities
   = link_to '<span>Atom Feed</span>'.html_safe, activities_path(format: :atom), class: 'atom'
 
+- if current_user.try(:supervisor?)
+  .alert.alert-success
+    'Hey Supervisor! See the dashboard link in the menu ^ ? That's your new dashboard!
```
carpodaster: Haha, love this :)

carpodaster:
:+1: Awesome.

Can I add one tiny feature request to that PR? (if you can't be bothered, just ignore me). Can you route `/supervisor/` to `/supervisor/dashboard`, as well?

carpodaster:
As discussed with @F3PiX: merging this to allow our RGSoC to move along. @F3PiX Can you note the comments made and tackle them post-merge?

app/controllers/supervisor/base_controller.rb
F3PiX: I followed the existing BaseControllers - apart from the fact that I wasn't aware that I was adding a Module ;-)  
So the question is: what is more important: Consistency between the Base Controllers, or follow this best practice? Or, @carpodaster: I _could_ change all of them?

carpodaster: It's a taste thing. Personally, I find the explicit module too verbose for a controller (and for multiple nested modules). In cases like that, I would stick with what the project already started out with.

Since we've now mixed both versions, I leave it up to you, @F3PiX, to decide ;)

app/controllers/supervisor/base_controller.rb
F3PiX: I will have to look into that again. One of the coaches suggested another solution, so it is on the agenda. 
Not only there is cancancan, but there is also this: https://github.com/rails-girls-summer-of-code/rgsoc-teams/blob/master/app/models/user.rb#L52
And there are all kinds of supervisor-checks, like this one in the header: 
https://github.com/rails-girls-summer-of-code/rgsoc-teams/blob/master/app/views/layouts/_header.html.slim#L26
I don't understand yet what the differences are, and when we can or cancan't use cancancan. Thanks for the tip; to be continued.

F3PiX:
Noted! And thanks!!

F3PiX:
@carpodaster I am not sure about the after-merge-workflow. Should I open this pull request again before pushing? Or open a new PR?

carpodaster:
@F3PiX New PR would be best

app/controllers/supervisor/base_controller.rb
F3PiX: @marcgreenstock Thanks for being pingable ;-)
What I found (first round). The `:supervise` ability needs a team to pass in. So, if I do: ```can? :supervise, @team``` it would work. BUT we don't have a team_id yet. I think because we only grant access to the supervisor's `user.roles`, not to the team roles. 
https://github.com/rails-girls-summer-of-code/rgsoc-teams/blob/master/app/models/ability.rb#L31
So, my conclusion is we can't use cancancan here to grant access. But we can use it in the views. Does this make sense?

F3PiX:
@marcgreenstock Could you help me out with Cancancan? I can think of a few strategies, but they will affect other parts of the app.

PR #310, opened 2015-09-12
Separate first and second round conferences

ramonh:
This hides the first round conferences in the conferences view and allows users to toggle them.

NB: All existing conferences will be set to have taken place in the first round


carpodaster:
Looks good – thanks! :)

PR #314, opened 2015-09-22
Adds team performance indicator to supervisor dashboard

roosdebildt:


app/controllers/supervisor/comments_controller.rb
```diff
@@ -1,5 +1,8 @@
 class Supervisor::CommentsController < Supervisor::BaseController
 
+
```
carpodaster: 3 superfluous lines

app/models/comment.rb
```diff
@@ -4,6 +4,7 @@ class Comment < ActiveRecord::Base
   belongs_to :application
 
   scope :recent, -> { order('created_at DESC').limit(3) }
+  scope :latest, -> { order('created_at DESC').limit(1) }
```
carpodaster: Let's use `.last` and not introduce a new scope for that

app/models/team_performance.rb
```diff
@@ -0,0 +1,72 @@
...
+    elsif @score >= 2
+      @performance = :orange
+    elsif @score == 0
+      @performance = :green
+    else
+      @performance = :orange
+    end
+  end
+end
+
```
carpodaster: Superfluous lines

app/views/supervisor/dashboard/index.html.slim
```diff
@@ -1,11 +1,29 @@
...
+    /this is where all the aggregated messages and info should go
+    /= flash[:notice]
 
 div.clearfix
   - supervised_teams.each do |t|
     div.dashboard-box
       section.team-title
         h3 = link_to "Team " + t.name, t
+        - if t.performance.score == :green
+          div = render 'green'
```
carpodaster: You can use the value of `t.performance.score` to DRY the render call

app/models/team_performance.rb
```diff
@@ -0,0 +1,72 @@
...
+  def initialize(team)
+    @team = team
+  end
+
+  def score
+    @score = 0
+
+    if @team.comments.any?
+      comments_score
+    elsif Time.now-2.days > Season.current.starts_at && Time.now < Season.current.ends_at
```
carpodaster: This is dificult to read and could do with a simplification. It's intention is to show the team as green for the first two days of the SoC?

There is the exact same line in line 22. Is it really depending on whether there are activities or comments, respectively? If not, maybe pull it out into a `early_bird_score` method?

app/models/team_performance.rb
```diff
@@ -0,0 +1,72 @@
...
+      @score += 3
+    else
+      @score += 0
+    end
+    evaluate_performance
+  end
+
+  def comments_score
+    latest_comment = @team.comments.ordered.first
+    if latest_comment.created_at <= Time.now-5.days
```
carpodaster: I find `5.days.ago` a lot easier to read.

app/models/team_performance.rb
```diff
@@ -0,0 +1,72 @@
...
+  def score
+    @score = 0
+
+    if @team.comments.any?
+      comments_score
+    elsif Time.now-2.days > Season.current.starts_at && Time.now < Season.current.ends_at
+      # Maybe we should change the +3 score temporarily for the upcoming week,
+      # because before Wed 16th comments-without-text were not being saved.
+      @score += 3
+    else
```
carpodaster: This conditional branch does nothing.

app/models/team_performance.rb
```diff
@@ -0,0 +1,72 @@
...
+      @score += 0
+    end
+
+    if @team.activities.any?
+      activity_score
+    elsif Time.now-2.days > Season.current.starts_at && Time.now < Season.current.ends_at
+      @score += 3
+    else
```
carpodaster: This conditional branch does nothing.

app/models/team_performance.rb
carpodaster: Oh wait, higher score means red, sorry. So it means all teams start off as red?

app/models/team_performance.rb
F3PiX: I tried that initially, but as I remember, I got unexpected behaviour regarding the UTC. I searched the code base to see how that was handled elsewhere. I didn't find the `days.ago` format, so I choose to go the safe way.

app/models/team_performance.rb
F3PiX: I added it as an escape for the few seconds in a Season that were not covered by the Time-statements, and I liked it much better than changing the operators to include `=`.

app/views/supervisor/dashboard/index.html.slim
F3PiX: @roosdebildt This one for you?

app/models/team_performance.rb
carpodaster: What I meant is that it doesn't change `@score`, so we might as well leave it alltogether

carpodaster:
Hey you two, can you add a screenie to see what the new scoring looks like? I'm pretty sure it looks awesome, but I'm so curious and I don't have the data set up locally right now.

app/models/team_performance.rb
```diff
@@ -0,0 +1,64 @@
...
+      @score += 1
+    elsif @team.last_activity.created_at > Time.now-3.days
+      @score += 0
+    else
+      @score += 1
+    end
+  end
+
+  def evaluate_performance
+    @performance = ' '
```
carpodaster: This line is not needed

roosdebildt:
This is how it looks like with the flash notices:

screen shot 2015-09-23 at 15 07 03

These are the icons we use for green, orange and red:
screen shot 2015-09-17 at 15 18 45

And this is how it looks like when the responsiveness is merged as well :
screen shot 2015-09-15 at 15 03 59

app/models/team_performance.rb
```diff
@@ -0,0 +1,64 @@
...
+      # because before Wed 16th comments-without-text were not being saved.
+      @score += 3
+    end
+
+    if @team.activities.any?
+      activity_score
+    elsif Time.now-2.days > Season.current.starts_at && Time.now < Season.current.ends_at
+      @score += 3
+    end
+    evaluate_performance
```
carpodaster: To me, `score` suggests a numerical return value but instead it returns a symbol. Does it stress you to tear the score-calculation and the semantical evaluation apart? The method `evaluate_performance` could be renamed to `performance` and return the symbol.

app/models/team_performance.rb
```diff
@@ -0,0 +1,64 @@
...
+  end
+
+  def comments_score
```
carpodaster: Since it's only called internally, you should make this method private.

app/models/team_performance.rb
```diff
@@ -0,0 +1,64 @@
...
+    elsif latest_comment.created_at <= Time.now-2.days
+      @score += 1
+    elsif latest_comment.created_at > Time.now-2.days
+      @score += 0
+    else
+      @score += 1
+    end
+  end
+
+  def activity_score
```
carpodaster: Since it's only called internally, you should make this method private.

app/models/team_performance.rb
carpodaster: Ok, scratch that. `Performance#performance` is a very bad name decision. `to_sym` would be consistent, but barely meaningful. Can't think of a good method name right now :(

app/views/supervisor/dashboard/index.html.slim
```diff
@@ -1,11 +1,26 @@
...
+    /= flash[:notice]
 
 div.clearfix
   - supervised_teams.each do |t|
     div.dashboard-box
       section.team-title
         h3 = link_to "Team " + t.name, t
+        div = render "#{ t.performance.score }"
+        - if t.performance.score == :orange
+          - flash.notice = "Team #{t.name} needs your attention"
```
carpodaster: Hm, does that work with multiple teams having the same score? It looks as if it would overwrite stuff.

F3PiX: It would. For now. Handling of the messaging will be in the next iteration. We probably are going to use the flash notice for a general notice that there are teams who need attention, and place the team-related message in the overview box. We need feedback from the supervisors to see if the team-specific message adds value at all, now that we have icons.

roosdebildt: :+1: on what Maud said.

F3PiX:
Oops, sorry, I wanted to close this comment screen :-)

app/models/team_performance.rb
```diff
@@ -0,0 +1,60 @@
...
+    @team = team
+  end
+
+  def buffer_days?
+    # the first few days, plus the days before and after the coding season
+    Time.now < Season.current.starts_at+2.days || !Season.current.started? || Time.now > Season.current.ends_at+2.days
+  end
+
+  def evaluation
+    @score = 0
```
carpodaster: Oh, you can put this into the initializer

app/views/supervisor/dashboard/index.html.slim
carpodaster: Ok, could you please add an issue to this repo once the PR has been merged? I consider this would introduce a bug (plus it's not _really_ elegant to set the flash hash in the view, and it might even stick around for the next request)

app/views/supervisor/dashboard/index.html.slim
```diff
@@ -1,11 +1,26 @@
...
+    p = link_to "Guide", 'https://github.com/rails-girls-summer-of-code/supervisors/blob/master/guide.md'
+    /this is where all the aggregated messages and info should go
+    /= flash[:notice]
 
 div.clearfix
   - supervised_teams.each do |t|
     div.dashboard-box
       section.team-title
         h3 = link_to "Team " + t.name, t
+        div = render "#{ t.performance.evaluation }"
```
carpodaster: It wouldn't surprise me if `render` implicitly calls `to_s` on its argument. Can you try omitting the interpolation call?

carpodaster:
Ok, I'm done. Just 1.5 things left – I'll merge right after that

app/views/supervisor/dashboard/index.html.slim
F3PiX: We will!

F3PiX:
Thanks!!!

PR #315, opened 2015-09-23
Remove jobs offers board

carpodaster:
As discussed a few weeks ago, the RGSoC Teams App is not the right place to promote job offers: it lacks visibility, job offers should go to the community mailing list instead.

@sareg0 We currently have one published job offer. Merging this PR will delete it, are you ok with that?


sareg0:
*sniff sniff* bye bye jobs board.

PR #316, opened 2015-09-24
DRYs up team table views

roosdebildt:
Moves duplicate code into partial. This issue was raised by one of the former sprints of Team Cheesy working on the supervisor dashboard (unfortunately I cannot find back which PR it exactly was).

### DIFF
--- a/app/views/teams/_table.html.slim
+++ b/app/views/teams/_table.html.slim
@@ -9,7 +9,7 @@ table.table.table-striped.table-bordered.table-condensed
       th = role.capitalize
     th Conferences
     th Location
-    - if current_user.roles.includes?('organizer')
+    - if current_user.admin?
       th Last checked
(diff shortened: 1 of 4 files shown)


app/views/teams/_table.html.slim
```diff
@@ -0,0 +1,33 @@
...
+      - @display_roles.each do |role|
+        td
+          ul
+            - team.send(role).each do |user|
+              li.user
+                = avatar_url(user, size: 20)
+                = link_to(user.name_or_handle, user)
+      td = links_to_conferences(team.members.map(&:conferences).flatten.uniq).join(', ').html_safe
+      td = team.students_location
+      - if current_user.roles.includes?('organizer')
```
carpodaster: This does not work as intended: `User#roles` returns a list of `Role` instances; comparing them with the string `"organizer"` will never be true. What you want it is `current_user.admin?`

PR #318, opened 2015-09-28
Simple Notepad in Dashboard

F3PiX:
Simple notepad added, and adjustments in UX.

### DIFF
--- a/app/controllers/supervisor/notes_controller.rb
+++ b/app/controllers/supervisor/notes_controller.rb
@@ -1,0 +1,38 @@
+class Supervisor::NotesController < Supervisor::BaseController
+
+  def index
+  end
+
+  def create
+    # @notepad = Note.find_by(id: current_user.id)
+    # if @notepad.save
+    #   redirect_to supervisor_dashboard_path
+    # else
+    #   puts " NOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOO dream on"
+    # end
+  end
+
+  def show
+  end
+
+  def edit
+  end
+
+  def update
+    @notepad = Note.find_by(id: @current_user.id)
+      if @notepad.update(notepad_params)
+          redirect_to supervisor_dashboard_path
+      else
+        puts "................NO Update <<<<<<<<<<<<<<<<<<<"
+        #what to do here?
+      end
+  end
+
+
+  private
+
+    def notepad_params
+      params.require(:note).permit(:body).merge(id: @current_user.id)
+    end
+
+end
--- a/app/controllers/supervisor/notes_controller.rb
+++ b/app/controllers/supervisor/notes_controller.rb
@@ -1,8 +1,8 @@
 class Supervisor::NotesController < Supervisor::BaseController
 
   def update
-    @notepad = @current_user.notepad
-    @notepad.update(notepad_params)
+    notepad = @current_user.notepad
+    notepad.update!(notepad_params)L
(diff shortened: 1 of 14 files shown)


app/models/user.rb
```diff
@@ -141,6 +141,10 @@ def current_student?
       student.any?
   end
 
+  def notepad
```
carpodaster: I'm not sure if I like adding `User#notepad` when it only applies to supervisor-users. I would love User have no knowledge of `Note` whatsoever. Could you move the `find_or_create` to the `Note` class and update the Supervisor::NotesController#update accordingly?

app/controllers/supervisor/notes_controller.rb
```diff
@@ -0,0 +1,15 @@
+class Supervisor::NotesController < Supervisor::BaseController
+
+  def update
+    notepad = @current_user.notepad
+    notepad.update!(notepad_params)L
```
carpodaster: The "L" shouldn't be there

app/views/supervisor/dashboard/index.html.slim
```diff
@@ -1,30 +1,43 @@
...
+                  .performance-icon = render 'orange'
+                td
+                  '= Needs Tender Loving Care
+              tr
+                td
+                  .performance-icon = render 'red'
+                td
+                  '= Red alert
+          div.col-sm-6 class=("col-lg-4 padding" if supervised_teams.count > 2)
+            = simple_form_for @current_user.notepad, url: supervisor_notes_path do |f|
```
carpodaster: It's better to access the `@current_user` w/o the @, ie. via its corresponding helper method.

app/controllers/supervisor/notes_controller.rb
```diff
@@ -0,0 +1,15 @@
+class Supervisor::NotesController < Supervisor::BaseController
+
+  def update
+    notepad = @current_user.notepad
```
carpodaster: It's better to access the `@current_user` w/o the @, ie. via its corresponding helper method.

config/routes.rb
```diff
@@ -80,9 +80,11 @@
   end
 
   get 'supervisor', to: 'supervisor/dashboard#index'
+  patch 'supervisor/notes', to: 'supervisor/notes#update'
   namespace :supervisor do
     get 'dashboard', to: 'dashboard#index'
     resources :comments, only: [:index, :create]
+    resources :notes
```
carpodaster: is this needed with the `patch` statement in place 4 lines above?

db/migrate/20150928180532_add_user_id_to_notes.rb
```diff
@@ -0,0 +1,5 @@
+class AddUserIdToNotes < ActiveRecord::Migration
+  def change
+    add_column :notes, :user_id, :integer
```
carpodaster: The foreign key should also have an index. Admittedly, the table won't go Big Data anytime soon, but it's good to make a habit out of it.

carpodaster:
Hi! I've made a few comments. Could you also add one or two screenies to illustrate what it looks like?

roosdebildt:
Screenshot (sorry I forget that all the time!):

screen shot 2015-09-28 at 18 10 22

F3PiX:
@carpodaster

F3PiX:
Aaaaaand back to you @carpodaster.

carpodaster:
:tada:

PR #321, opened 2015-10-25
Add Contibutor Covenant v1.2 Code of Conduct

carpodaster:
I chose the [Contributor Covenant](
http://contributor-covenant.org/) over the [RGSoC CoC](http://railsgirlssummerofcode.org/about/code-of-conduct/) as I find the CC CoC less verbose for an OSS project. What do you think?

Also, shall we link to the RGSoC CoC?

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,22 @@
+# Contributor Code of Conduct
+
+As contributors and maintainers of this project, and in the interest of fostering an open and welcoming community, we pledge to respect all people who contribute through reporting issues, posting feature requests, updating documentation, submitting pull requests or patches, and other activities.
+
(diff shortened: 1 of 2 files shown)


alicetragedy:
I like this and completely agree with you over the topic of verbosity. Contributor Covenant was created especially with OSS projects in mind, which makes it a good fit for us. I also see no necessary need for linking to the RGSoC CoC :shipit: 
Question: do you think we should also add it to the `summer-of-code` repo? I know that we have the CoC on the website, but I see it as being for the program itself and not necessarily in the context of contributions to the site as an OSS project – maybe it's not clear for someone checking out the repo that we have a CoC at all..

sareg0:
yes, +1 for a code specific to contributions, for all the reasons outlined. I'm good to push this through.

carpodaster:
> Question: do you think we should also add it to the summer-of-code repo? I know that we have the CoC on the website, but I see it as being for the program itself and not necessarily in the context of contributions to the site as an OSS project – maybe it's not clear for someone checking out the repo that we have a CoC at all..

Yes, let's add it to the GH Page repo as well!

ramonh:
This is great, thank you @carpodaster!

PR #330, opened 2015-11-30
Add form to create a project proposal

ramonh:
This adds the `C` in the `CRUD`. :)

The form can submit and create a new project proposal. The form then redirects to the project list

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -3,6 +3,7 @@ class ProjectsController < ApplicationController
   before_action :login_required, only: [:new]
 
   def new
+    @project = Project.new
   end
 
   def index
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -10,4 +10,25 @@ class ProjectsController < ApplicationController
     @projects = Project.all
   end
 
+  def create
+    @project = Project.new(project_params)
+
+    @project.submitter_id = current_user.id
+    respond_to do |format|
+      if @project.save
+        format.html { redirect_to projects_path, notice: 'Project was successfully submitted.' }
+      else
+        format.html { render action: :new }
+      end
+    end
+  end
+
+  private
+
+  def project_params
+    params.require(:project).permit(
+      :name, :mentor_name, :mentor_github_handle, :mentor_email,
+      :url, :description, :issues_and_features, :beginner_friendly
+    )
+  end
 end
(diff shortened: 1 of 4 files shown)


ramonh:
@carpodaster would you mind taking a look at this please?

app/controllers/projects_controller.rb
```diff
@@ -3,10 +3,32 @@ class ProjectsController < ApplicationController
...
   end
 
   def index
     @projects = Project.all
   end
 
+  def create
+    @project = Project.new(project_params)
+
+    @project.submitter_id = current_user.id
```
carpodaster: A taste thing, but I prefer to assign high level objects instead of IDs for AR associations. In this case: `@project.submitter = current_user`

app/views/projects/new.html.slim
```diff
@@ -2,5 +2,25 @@ nav.page
...
+  = f.input :description, wrapper_html: { class: 'col-md-6' }
+  = f.input :issues_and_features, wrapper_html: { class: 'col-md-6' }
+  = f.input :beginner_friendly, wrapper_html: { class: 'col-md-6' }
+
+  .actions
+    ul.list-inline
+      li = f.submit 'Submit', class: 'btn btn-success'
+      li = link_to 'Cancel', @user, class: 'btn btn-default btn-sm'
 
 img src="https://media.giphy.com/media/F8oLZ6P5WCAbS/giphy.gif"
```
carpodaster: Oh, could you remove my placeholder pls? :wink:

carpodaster:
@ramonh sorry for the delay and thank you for the PR! I've made two tiny comments. Feel free to merge! :)

ramonh:
No worries, @carpodaster. Thank you so much for the feedback! I'll make those changes and merge

app/controllers/projects_controller.rb
ramonh: Quick question for future reference: Should I squash these adjustments?

carpodaster: Short answer: I don't think we need to. 

Longer answer: I would really love to see squashed commits when there was a lot of noise due to some exploration – for instance when students are working on the app. But squashing commits is scary and no student should be bothered with it, hence: nope, let's not squash commit. :)

ramonh: Okeydokey, I'll keep that in mind. Thank you for your insight! :+1:

Issue #335, opened 2015-12-06
Improve header nav (mobile)


alicetragedy:
The header nav on mobile definitely needs some :heart: (see screenshot, which is what happens when clicking on the burger menu)
This is a low-priority issue, but still worth noting. I might be able to take a look at this next week but if anyone else wants to jump on this issue, feel free to do so!

<img width="679" alt="screen shot 2015-12-06 at 14 10 19" src="https://cloud.githubusercontent.com/assets/3143348/11613502/9c4e7bac-9c23-11e5-8381-c31a5f81a88c.png">

F3PiX:
Hi @alicetragedy ! I have the exact same problem in my current pet project. I surely want to take this one, if you think it is doable for me? And maybe you point me in the right direction?

alicetragedy:
Hi @F3PiX! For sure, I've assigned you to it as I'm absolutely certain that it's doable for you :+1: I will not be online much in the next few days, but ping me if you're stuck and I will get back to you about it :)

beanieboi:
did #358 fix the issue? i try to lean some CSS so i was looking at the issues tagged with "styling" :)

alicetragedy:
oops! thanks for asking @beanieboi — #358 did indeed solve the problem so I'm going to close this now. :) feel free to ping me if you want to pair on something css-related with me! :sparkles:

beanieboi:
@alicetragedy totally! would love to do that! i will look at some CSS issues and try to work on them and try to do as much as i can to learn. when i'm stuck, i'll ping you on slack so we can schedule something? does work for you? :sparkler:

alicetragedy:
@beanieboi sounds perfect! 👯

PR #358, opened 2015-12-25
Improve header nav (mobile)

F3PiX:
- [x] Step 1: Added navbar-default class to header.
- [x] Step 2: Restore regular menu

Issue #335

### DIFF
--- a/app/views/layouts/_header.html.slim
+++ b/app/views/layouts/_header.html.slim
@@ -10,8 +10,8 @@ section#header
 
     div.row
       div.col-xs-offset-7.col-xs-5.col-md-offset-4.col-md-8
-        nav.top
-          ul class="navbar-default collapse navbar-collapse" id="menu"
+        nav.top class="navbar-default collapse navbar-collapse" id="menu"
+          ul
             - if Season.projects_proposable?
               li class="announcement"
                 = link_to 'Submit your Project', new_project_path
--- a/app/views/layouts/_header.html.slim
+++ b/app/views/layouts/_header.html.slim
@@ -10,26 +10,27 @@ section#header
 
     div.row
       div.col-xs-offset-7.col-xs-5.col-md-offset-4.col-md-8
-        nav.top class="navbar-default collapse navbar-collapse" id="menu"
-          ul
-            - if Season.projects_proposable?
-              li class="announcement"
-                = link_to 'Submit your Project', new_project_path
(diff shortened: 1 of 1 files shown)


F3PiX:
@alicetragedy 
I added `navbar-default` class to the header. Now, it pushes the text out of the way. Is this the way you want to go?

Also, the regular menu is messed up now.

alicetragedy:
@F3PiX sorry for getting back to you so late. I finally managed to take a look and yes, I think that's the way we could go. It looks good for the mobile version :)
The next step is as you outlined above to restore the regular menu. If I recall correctly the [bootstrap documentation](https://getbootstrap.com/components/#navbar) should be of great help.

alicetragedy:
I am still looking into it @F3PiX but: I realised that first and foremost, in order to change it, the `navbar-default` should be moved out of the `ul` class (for example, into `nav.top`). 
After that, you can add a `background-color` to it. 
I think the only way to override the color of a `navbar-default` is to also add 
```
background-image: none
background-repeat: no-repeat
```
However there's still a couple of strange things happening with the css (the collapse will pick up that red color, which I am not sure we want). An alternative would be to create a `navbar-custom` class in the css (it will be the color of the navbar we currently have) and add to that extra specifications for the `collapse` (drop down menu). 

Maybe try some stuff around with the info above, I'll also dig deeper into it from my end and see if there's some totally obvious way I have not considered.

F3PiX:
Just moving the nav_bar + menu id out of the ul did the trick! 
Regarding the other problems: I would very much not try to solve that myself. It is way over my head.  Okay? One question though: wouldn't most of the (new) problems be solved if we replace the 'header'  by a regular bootstrap navbar? At the least, I suppose Bootstrap will take good care of the dropdown menu.

alicetragedy:
@F3PiX that's okay, I understand :)
so, it only took one more change for it to work properly (with your restoring the change, the menu was stuck collapsed in responsive view): the `collapse` class needed to be moved outside of the unordered list into its own div. I've also added `role="navigation"` to improve accessibility. 
This should now work properly as intended, and I think we can merge the PR. Great work Maud!! :tada:

F3PiX:
:8ball: Sorry I missed that. Glad to have you watching my back! :bow:

alicetragedy:
No worries! I think we were each other's rubber ducks ;)

Issue #225, opened 2015-05-26
Create bogus test data for local development


carpodaster:

alicetragedy:
+1 on this – it would be amazing, especially when it comes to the application process, and it would make it so much easier for everyone to contribute. yayy! :+1:

sareg0:
totes, it's a problem we had in trying to mess about with stuff

carpodaster:
Switched assignee as @F3PiX is already hard at work on this.

Maud, if you want to go an extra mile, you can make FactoryGirl work in development (currently only loaded for the test env) and use the defined factories in `seed.rb`.

The good thing: the defined factories are tested, the contents in `seed.rb` are not. Should our schema change or validations be added, logic `seeds.rb` is likely to break first. Factories are adapted more frequently.

F3PiX:
O that sounds interesting. I will look into that. 
(The seeding I am doing now is basically to get a feel for the app.)

F3PiX:
What seeds do we need most?
[ ] Filled in application forms
[ ] Teams (we do have user seeds, but they are not yet in teams)
[ ] ?

alicetragedy:
@F3PiX maybe filled in projects as well (the application process now involves selecting an approved project from a dropdown menu, so one or two projects should be available).

carpodaster:
@alicetragedy @F3PiX all of the above :wink: 

* An unaccepted team
* An accepted team
* A accepted, but voluntary team
* An application draft, unsubmitted
* An application draft, completed
* A  `proposed` project
* An `accepted` project
* A `rejected` project

F3PiX:
:sparkles: Thnx

alicetragedy:
Would be great to have this in time for the start of the program this year, so that if there is a team working on the teams app, they can get set up more easily.

carpodaster:
Hello @F3PiX – are you still working on this or do you have a branch that someone else can continue working on? The SoC starts in approx. two weeks and it would be nice to make the setup of a local teams app installation as easy as possible

carpodaster:
Done thanks to the amazing work of @F3PiX !

PR #370, opened 2016-02-10
WIP Update FFaker gem, now with CheesyLingo

F3PiX:
This version of the FFaker gem includes CheesyLingo. CheesyLingo is added to FFaker by Team Cheesy during Summer of Code 2015 :-)

Before I go further with implementing some seeds with cheesy FFaker fun : is this a good idea?  
Related to #225

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -48,7 +48,7 @@ group :development, :test do
   gem 'factory_girl_rails'
   gem 'database_cleaner'
   gem 'byebug', require: !ENV['RM_INFO'] #require parameter is workaround for RubyMine with Rails ~> 4.1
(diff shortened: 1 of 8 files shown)


carpodaster:
Hehe, thanks @F3PiX  :smile: 

> Before I go further with implementing some seeds with cheesy FFaker fun : is this a good idea?

I'd say if there's new bogus data coming up (re: #225) then sure! Further rewriting existing factories is probably overdoing it.

F3PiX:
Haha, we could argue that CheesyLingo itself qualifies as overdoing it :-) Yes, I will work on the bogus data. Thanks, Carsten.

Issue #365, opened 2016-02-03
Application process: Remove requirement of coaches' comments


carpodaster:
Remove 'why successful' field.

Q: Allow students to fill in the »How much time are your coaches packin'?« question or shall we remove it alltogether?

F3PiX:
From my experience, it is a) hard for students to foresee the availability of their coaches , and b) it is hear-say, so it is hard to compare between applications. On the other hand, it may be handy to ask the coaches directly how they see their commitment. We had very committed coaches, but came nevertheless coaching hours short, especially at times when some of the coaches were out of town.

ramonh:
Tricky indeed. I wonder, would it perhaps be wise to advise coaches before confirming their involvement, perhaps in the form of a checkbox, that they need to be as available as possible? Perhaps this could resolve itself as we make coaches more involved in the overall communication this time around.

In any case, I'll happily handle the removal of this field.

carpodaster:
Having meditated on this for a while now I'm in favor of leaving coaches out of the application process alltogether, save the fact that they have to accept being part of the team in the first place.

As @F3PiX pointed out, it's very difficult to estimate how much time will be available anyway so whatever is being entered may not be valid anymore come summer. Or could be fake in first place ;)

Very welcome side-effect: reducing another access level (i.e. 'coach') should also simplify the programming behind the application process.

ramonh:
Sounds good! I'll go ahead and remove both fields, then.

alicetragedy:
@carpodaster does this mean we are not entering any values that refer to the coaches' availability in the form at all? (fine with me, am just checking for guide update purposes)

carpodaster:
@alicetragedy ah, good that you have The Guides™ in mind. Ooops.

The only bit of information we could us is their general level of support, e.g. having a "coaching company" or some office space. I don't think we explicitely ask for that, or do we?

So my proposal is as follows:

* Do not enable / require to edit the application draft anmore
* Remove "why will this team be successful?"
* Remove "how many hours per week?"
* Make sure that we gather information about their general support structure

alicetragedy:
:+1: for removing ALL THE THINGS
when you write "Make sure that we gather information about their general support structure" do you mean that the coaches will be entering this somewhere? Or would you see this as the students' job? 
I believe this is info which the students/team could provide, if we really want to make the application non-editable by the coaches and have a hands-off approach (ie. we have the support of a coaching company/we have an office space in mind).
[...]

PR #371, opened 2016-02-10
Remove the fields to be filled in by coaches from the application process

ramonh:
This addresses #365.

The coach fields 'How many hours are you available for?" and "Why do you think this team will be successful?" Are removed with this.


carpodaster:
Awesome, thank you!

Issue #372, opened 2016-02-11
Increase session timeout for Github logins


F3PiX:
When we were working on the Supervisor Dashboard during SoC2015, some of the supervisors told us about annoyances with Github Authentication:
1 having to authenticate w GitHub "all the time" (which boiled down to: every day or every few days)
2 after auth w GitHub, you are not returning to the page where you came from

#1 Not sure if this managed by GitHub. If not, it can be managed by Devise (AFAIK): is it okay to extend the time-out (the user session without activity) to 1 or 2 days?

#2 We should redirect back to previous page.

alicetragedy:
I think Nr. 2 is definitely a huge priority. Thanks for adding this! 👍

anikalindtner:
AFIAK @carpodaster said that we can't do anything about the redirect to the previous page since that's handled by the GH auth thing? I also find this super annoying and would love to have this different.

carpodaster:
Ohai. I have may have just fixed no. 2 :relaxed: 

If no. 1 was reported as being annoying during the summer we should definitely fix it. It would be super-nice if you could work on that, Maud. Especially because I have no idea about devise's internals :grin:

alicetragedy:
Carsten's :sparkles: magical dev powers :sparkles: 

![tumblr_m9dlcwp2yr1qiyseco3_250](https://cloud.githubusercontent.com/assets/3143348/12989968/085e07d6-d108-11e5-8b14-13e311861614.gif)

anikalindtner:
WHOAA

F3PiX:
:bow:

beanieboi:
@F3PiX let me know if #373 fixed the issue, hopefully we can close this then :)

F3PiX:
Hi @beanieboi, I don't know how to check if it is fixed. I never had this issue myself. It _sounds_ like it is solved though ;-)

alicetragedy:
@carpodaster seems like the "redirect after login" issue is still not fixed — at least it seems like it doesn't work with the https://teams.railsgirlssummerofcode.org/apply link, which is a bummer because I would love to link to the application page directly from ie. the application guide. Right now, the "blogs" as an index page for the teams app (which is the page we link to by default, and the page which is shown by default when logging in) seem to be pretty confusing for most students applying. I'll investigate further, to see if there's anything we've overlooked..

alicetragedy:
I think we can close this!!!! :tada: :confetti_ball: :raised_hands:

PR #373, opened 2016-02-13
extend session validity

beanieboi:
as mentioned in #372. you get logged out after you close your browser. this extends the validity of the session to 1 week (just picked something random) and will also extend the session every time you visit.
if you visit the site, say, every day. it will never expire. if you are on vacation for 2 weeks, your session will time out and you have to login again.

### DIFF
--- a/config/initializers/session_store.rb
+++ b/config/initializers/session_store.rb
@@ -1,3 +1,6 @@
 # Be sure to restart your server when you modify this file.
 
-Rails.application.config.session_store :cookie_store, key: '_rgsoc-teams_session'
+Rails.application.config.session_store(:cookie_store, {
+  key: '_rgsoc-teams_session',
+  expire_after: 1.week
+})


carpodaster:
Thanks @beanieboi !

Are you sure this solves the auto-logout issue? I was thinking that a session cookie does not expire at all as long as there is no `expires_after` key. I was thinking that we devise is responsible for the overly early timeout.

beanieboi:
hey @carpodaster 

a session cookie expires when you close the browser. it's only valid for as long as you keep the browser (your session) running.

adding the `expires_after` converts the `_rgsoc-teams_session` cookie into a real cookie with a expire date in 1 week from now. but rails adds some automatic here. it updates the expire date every time you visit the page. if you visit the page today it will set the expire date to 7 days from now. if you visit the page again in 2 days from now, it will update the expire date again to 7 days from the page visit. this means if you visit the page every day the cookie will never expire. if you don't come back for over 7 days. you need to login again.

does that make sense? :D

carpodaster:
Yeah, it does make sense :)  My confusion stems from the fact that we're using cookie-based session management already. I know that closing the browser resets the session but I am not sure if this is the reason behind the reported logouts: I rarely ever close my browser (I close the laptop lid and continue where I left next time) but I do get these logouts as well. I cannot tell how often, though.

F3PiX:
@beanieboi Great! 
Do you happen to know if GitHub is managing session duration by itself in its OAuth access API? As I understood, every Oauth 'destination app' decides on some of its own settings when allowing access by OAuth. But I found no further details.

beanieboi:
@F3PiX @carpodaster hiiii!

we have to distinguish between 3 session here. 

- one session is the session of the teams app
- one session is the OAuth session. AFAIK the GitHub OAuth token has no expiration. this means the OAuth token we get back from GitHub can be used until it's manually revoked by the user. i don't know the teams app that good but if we use GitHub OAuth only for login this doesn't need to bother us since we only ask the GitHub OAuth API.
- one session is the session on GitHub.com, we can't control this one

When you are logged in with the teams app right now, you get logged out when you are closing the browser and you have to sign in with GitHub again. What is possible here is that you are also logged out of GitHub. so you actually have to perform 2 logins. in order to login to the teams app you now have to login into GitHub.
this PR will only fix the issue that you get logged out of the teams app and you can use the teams app without login in that often.

carpodaster:
Alright, let's get to merrrrge! Thanks a lot @beanieboi !

beanieboi:
:heart_eyes_cat:

F3PiX:
:sparkles:

PR #374, opened 2016-02-13
update link to imprint

beanieboi:
i just noticed that the link to the imprint returns a 404. let me know if the link is still incorrect, i grabbed it from the official site.


carpodaster:
Thanks! :bow:

Issue #375, opened 2016-02-13
rake db:migrate broken


beanieboi:
hey

i just tried to setup the teams app and the migrations are not running through.

```
undefined method `project_proposals_open_at' for #<Season:0x007fe1784709f8>/Users/ben/Development/rgsoc-teams/app/models/season.rb:96:in `set_application_dates'
/Users/ben/Development/rgsoc-teams/app/models/season.rb:9:in `current'
/Users/ben/Development/rgsoc-teams/db/migrate/20150625085615_add_seasons_to_mailings.rb:4:in `change'
NoMethodError: undefined method `project_proposals_open_at' for #<Season:0x007fe1784709f8>
/Users/ben/Development/rgsoc-teams/app/models/season.rb:96:in `set_application_dates'
/Users/ben/Development/rgsoc-teams/app/models/season.rb:9:in `current'
/Users/ben/Development/rgsoc-teams/db/migrate/20150625085615_add_seasons_to_mailings.rb:4:in `change'
Tasks: TOP => db:migrate
(See full trace by running task with --trace)
```

the problem is that in `20150625085615_add_seasons_to_mailings.rb` we try to access `Season.current` which tried to create a Season with attributes `project_proposals_open_at` that are not present yet in the current point of migrations.

what do you think would be a good fix? i can work on it!
what i usually do is to squash the migrations. there is no point in having 120 migration files. i would usually squash all migrations that are older than 1-2 months in 1 big migrations. this allows you to still revert deployments and do down migrations. i doubt that there is value in really old migration files.

let me know what you think about this. happy to work on any solution.

best
ben

beanieboi:
i can also move the line `Mailing.update_all(seasons: [Season.current.name])` from 
`20150625085615_add_seasons_to_mailings.rb` to `20151205133354_add_project_proposal_dates_to_seasons.rb`

it alters the migrations, but i think there is no way to fix this without changing the migrations.

F3PiX:
Just pulled upstream and run db:migrate. I did not have this migration problem. Does that prove anything at all? :-)

beanieboi:
hey @F3PiX 

did you start with a empty database? when i changed the two files like mentioned above it works for me. it's important that the db is empty and you run all migrations.

F3PiX:
Aha, no, it probably had to do only the latest migration. Thanks for clarifying.

carpodaster:
[...]

PR #376, opened 2016-02-13
squash migrations

beanieboi:
as discussed in #375 this squashes all migration made before 12/1/2015. this means there are only a few migrations left, which is nice if you need to revert them.

let me know what you think.


carpodaster:
Hey @beanieboi – sorry for the long wait and again, thanks for your PR. I've just push 1ec0cacd6d26f4e0fd07a68be03d69251c41b7ec which solves the problem behind this PR by fixing the offending migrations. I'm closing this PR in favor of that commit since it's less "invasive". I hope you're ok with this.

beanieboi:
thank you so much!

Issue #363, opened 2016-02-03
Rework team-forming process: no mentors, other student and coaches get invite


carpodaster:
Non-orga people can no longer add mentors to their teams.

Coaches added to the team will get an invite notification they can accept. Only then will they appear on the roster.

Idea: Even adding the 2nd student should be done by invite. As long as the "team" doesn't have a 2nd student, it will not be a `confirmed` team (adding a state machine here?) and it will not appear in the list of teams on the website.

Once the 2nd student accepts the invitation, the teams gets confirmed and is officially created.

carpodaster:
@ramonh @lucaspinto Whould anyone of you have time to work on this this week?

ramonh:
Sure thing @carpodaster I'll give it a go :)

ramonh:
What I'll try to do is break this into smaller pieces:

- Remove Mentors
- Remove and add coach 'invite' system
- Remove and invite second student, migrate to state machine

carpodaster:
Cool, @ramonh !

I would implement the other-student-invite first since only after accepting the invitation do we have an actual `confirmed` team (it may be `pending` first, in terms of the state machine).

I would leave the logic of a team to generally contain mentors or any other role (like supervisors later on). We just have to ensure that a team can only ever be created on the public website with 1st) another student and 2nd) one or more coaches, all of which have to accept an invitation.

Removing the second student should probably mean the entire team gets destroyed?

ramonh:
Makes sense, thank you @carpodaster. 

If I understood correctly then, it would be best for the team not to be `confirmed` until both a second student and at least one coach accept their invitations, correct?

carpodaster:
I would drop the requirement of having at least one coach to have a `confirmed` team. Pure pragmatic reasons: a confirmed team should be able to create an application draft. They can start filling out their stuff without having to wait for a coach to hop on (they may still be in the process of finding any).

ramonh:
Got it. I'll start having a look today. Thanks for your help! :+1:

carpodaster:
No, thank _you_, sir :)

ramonh:
:smile:

lucaspinto:
Alas, I won't have time before Saturday.

ramonh:
This probably has been answered elsewhere, but I can't seem to find it.

I've just about wrapped up the team revamping (#386), and have only the mentors to remove.

[...]

PR #378, opened 2016-02-23
Transition to state machine for teams and add requirement for team to have two students

ramonh:
Apologies for the long title!

What this does is start to address #363, by doing two things:

- Turn the team model object into a state machine with two states:
   - `pending`: This is when there's only one student in the team.
   - `confirmed`: This is when there's two students in the team.
- Add the requirement for a team to be confirmed before students can create one.

### DIFF
--- a/app/controllers/application_drafts_controller.rb
+++ b/app/controllers/application_drafts_controller.rb
@@ -13,7 +13,11 @@ class ApplicationDraftsController < ApplicationController
   end
 
   def new
-    redirect_to new_team_path, alert: 'You need to be in a team as a student' unless current_user.student?
+    if current_user.student?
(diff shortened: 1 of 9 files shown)


carpodaster:
Hey @ramonh, looks great! Just to be clear: there's currently no way for the 1st student to get a 2nd one into the team, right?

ramonh:
Thanks @carpodaster! I'm glad you like what you see so far.

I'm not sure, to be perfectly honest. Will get back to you on that.

To be clear, this doesn't add the invitation system yet. I didn't want to make an overloaded PR :sweat:

carpodaster:
Heh, makes perfect sense and I assumed as much. I just wanted to make sure. Please feel free to merge!

ramonh:
Awesome, thanks @carpodaster for checking and for confirming! Will do.

PR #381, opened 2016-02-26
fixed errors on starting of project

sonianand11:
I cloned fresh code(master branch) and was setting up project. I was facing errors while configuring it by `rake db:create db:migrate`

This PR solves following issue :
1) `undefined method project_proposals_open_at `
- While doing `rake db:migrate` the migration `20150625085615_add_seasons_to_mailings.rb` was raising error of `project_proposals_open_at` is undefined because of validation callback called while creating Season object. I fixed by moving `Mailing.update_all(seasons: [Season.current.name])` to `db/seeds.rb`

2) `undefined method published_at of nil class `
- The factory of `status_update` do not contain published_at attributes. So when we render root page it was raising error. Which is fixed by adding published trait in factory.

### DIFF
--- a/db/migrate/20150625085615_add_seasons_to_mailings.rb
+++ b/db/migrate/20150625085615_add_seasons_to_mailings.rb
@@ -1,6 +1,5 @@
 class AddSeasonsToMailings < ActiveRecord::Migration
   def change
     add_column :mailings, :seasons, :text, array: true, default: []
-    Mailing.update_all(seasons: [Season.current.name])
   end
 end
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -9,3 +9,5 @@ FactoryGirl.create(:supervisor)
 FactoryGirl.create_list(:user, 6)
 
 FactoryGirl.create_list(:status_update, 5)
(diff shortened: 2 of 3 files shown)


carpodaster:
Hi @sonianand11 - thank you for your PR and sorry for not getting back to you earlier. Your PR targets the problem described in #375, right? We already have a proposed fix for that (#376). I'll weigh both PRs and merge. We're currently in the pre-launch operation for this year's application phase so please excuse us if it may take a few more days.

sonianand11:
Hi @carpodaster, yes both PRs are identical issue. Take your time to merge it. I tried to cover all issue I was having while configuring app.

carpodaster:
Hey @sonianand11 – sorry for the long wait and again, thanks for your PR. I've just pushed 1ec0cacd6d26f4e0fd07a68be03d69251c41b7ec which solves the problem behind this PR by fixing the offending migrations. I'm closing this PR in favor of that commit since it's less "invasive". I hope you're ok with this.

sonianand11:
yes, I am ok with it.

PR #386, opened 2016-03-06
Add coach role confirmation

ramonh:
This is the next in steps toward resolving #363

It adds the following:

- Displaying the confirmation status of coaches in the team's view
- Coaches can confirm their status in the team's view
- Adds the following line to the email: 'If you've been added as a coach, please note that you need to confirm your role as a coach.'

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20160217141602) do
+ActiveRecord::Schema.define(version: 20160306133317) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
@@ -183,8 +183,9 @@ ActiveRecord::Schema.define(version: 20160217141602) do
     t.integer  "team_id"
     t.integer  "user_id"
     t.string   "name",       limit: 255
-    t.datetime "created_at",             null: false
-    t.datetime "updated_at",             null: false
+    t.datetime "created_at",                                 null: false
+    t.datetime "updated_at",                                 null: false
+    t.text     "state",                  default: "pending", null: false
   end
 
   create_table "seasons", force: :cascade do |t|
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20160306133317) do
+ActiveRecord::Schema.define(version: 20160309121328) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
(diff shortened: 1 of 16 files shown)


spec/models/role_spec.rb
```diff
@@ -15,6 +15,51 @@
...
+
+      it 'has the pending state' do
+        expect(subject).to be_pending
+      end
+    end
+
+    context 'when the user is added as a student' do
+      let(:role_name) { 'student' }
+
+      it 'has the pending state' do
```
alicetragedy: minor correction: should this line (as well as 49 and 57) read 'has the confirmed state' instead?

ramonh: Y... Yes. Thank you for catching that!

You caught me in my nasty copy pasting habit D:

alicetragedy: no worries, the same thing has happened to me SO often. Maybe that's why I caught it so quickly ;)

ramonh: Glad you understand. I fixed and squashed the typos!

db/schema.rb
```diff
@@ -124,6 +124,31 @@
     t.integer  "round",                    default: 1
   end
 
+  create_table "events", force: :cascade do |t|
```
alicetragedy: I'm not sure what this and the `job_offers` table are doing here. I think it might have to do with the problems you were having running the migrations — when you set it up from the db dump — but I really don't know what the best way to remove these without messing with the schema is.. 🙈

ramonh: Uh oh, hm. I take it these were supposed to be removed?

You've got me here. Maybe I can remove the changes to the schema?

alicetragedy: yes, they are supposed to be gone :) 
there's two migrations `remove_job_offers` and `drop_events` which supposedly make sure those tables are dropped. Not sure how this happened ¯\_(ツ)_/¯ 
As for fixing it, yes, maybe simply removing the changes to the schema (apart from `state` on the roles table obviously) might do it. Can you give it a try?

ramonh: Will do!

carpodaster:
@ramonh @alicetragedy I've added some commits to the PR (apologies, Ramon, I wanted to create a PR for the PR to have it merged into your remote branch, but I got fooled by git :( ).

It introduces confirmation tokens and basically everyone who knows the link to the token can confirm the coach role. The link, however, is only displayed to the invited coach.

Ideally, we should also send an email. `GUIDE_ROLES` already get an email when they are added (note to self: we need to remove the mentor from that list). We can send coaches a different flavor of email that directs them to the teams page where they can confirm themselves.

Other than that, I think this PR is ready to go. Thanks so much @ramonh !

carpodaster:
Oh, one more thing: we probably need to flag all existing coach-roles of past seasons' teams as confirmed.

ramonh:
Thank you so much for your help @carpodaster! I approve of these improvements :+1: 

No need to apologize. Git sometimes gets the best of us, and this creates less overhead so I'm all for it!

I prefer the use of tokens over my param solution. The improved displaying of buttons and flashes is great too. Thank you so much! I'd say it's good to go too.

Not sure how we'd do the latter, though. Would it make sense to have a migration?

carpodaster:
I'm not really sure how to go about it; we currently have reports about migrations being broken exactly because I introduced these kinds of data migrations in the past. Now that the respective code for it has changed, the migrations break. Still, we won't remember that we have to manually intervene after we deployed this branch to production so let's add it to the migration file.

ramonh:
Makes sense to me. I'll be on the move until late tonight. Could we perhaps wait on merging the PR or would you mind please adding that line to the migration?

carpodaster:
@ramonh done! Hacked together rather quickly so I hope I'm not causing havoc with our existing data :bomb:  :laughing:

ramonh:
Oh wow that is more than one line! Looks awesome. Thank you so much @carpodaster you've been a tremendous help.

carpodaster:
Bam, merged. Thanks a lot for this, @ramonh ! :gift:

alicetragedy:
YAAAYYYYY!! :tada: :confetti_ball: great work!

ramonh:
Thank you both of youse!

Issue #205, opened 2015-04-16
Team: ensure that student-restrictions cannot be tricked by editing a role directly


carpodaster:
There are (or will be) validations for a team's composition on `Team`, but they won't be fired when editing a role:

* [ ] Student must not be part of another team (#138)
* [ ] Team must not have more than two students (#204)

klappradla:
I'd love to look into this (and related) if noone is on it right now.
Would also suggest to, while on it, hide the `+ New Team` button in the team's index view for students already being part of a team. This button tricked me at first.

alicetragedy:
as far as I know @klappradla no one is on it. ( @ramonh @carpodaster correct me if I'm wrong ) so please go ahead :) +1 on hiding the `New Team` button, I think it's sensible. In fact, I believe there are quite a few UX things we could be fixing on the teams app ;) thank you!!

ramonh:
All :+1: from me!

klappradla:
I added the requested validation process in this PR: https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/389
It includes some rather subjective thoughts on the UX I would be totally fine to withdraw again. As the issue addresses more model-level stuff, I didn't really want to make any view- and UX-related decisions and therefore did not hide the `New Team` button yet.

If someone could give me a quick feedback wether UX-wise I'm getting this right or wrong, I'd take this as a basis to then go ahead with button +/-.

klappradla:
I'll try to submit the UX changes (hide `New Team` and remove `Add Member` button) tonight if I'm not to sore from work (will be Saturday otherwise).

@alicetragedy since you're changing quite a lot around the forms, is it still reasonable for me to start from **`master`** for this?

alicetragedy:
@klappradla absolutely. I'm not dealing with anything in the `teams` views, just `application_drafts`, so we won't run into any problems if you start from **`master`**. go ahead and — thank you! :sparkles:

carpodaster:
I've stuffed a few "security" holes in a85416d but I noticed another: When you from a new team, your own user is preset as the first role but you can choose freely between being a coach or a student. It makes no sense for other roles than a student to form a team, hence this shouldn't be possible. Any takers?

klappradla:
[...]

PR #389, opened 2016-03-06
Add validation to limit the max. number of students per team

klappradla:
Validation for the teams model to limit the max. number of students to two. Editing and deleting roles, etc. are taken into account, the validation process takes place in memory, not relying on data yet to be persisted.

ace624f adds the validation for the team model.

dda1f8f adds the classic behavior of firing the team's validations when saving a role, which will add the default Rails error message to the role: **`Team is invalid`**. I personally don't like this behavior, as I think it doesn't provide meaningful insights to the user. It rather adds complexity for her to understand what went wrong (*"2 errors prohibited this role from being saved ... Team is invalid"*).

Therefore 50812f9 undos firing the team model's validations when saving a role and instead displays the team's own error message in the role's form, if adding/editing a role does not suit the constraints of the team the role is connected to. As the **`Add a member`** button in the team's show view is just a shortcut duplicating the behavior of the normal **`Edit`**, the team-context is obvious anyways, so I consider showing the team errors more useful here.

If you don't like this approach, you can either just discard the last commit (50812f9) or tell me to rework it. If you like it, I would further suggest to fully ditch the **`Add a member`** button as it is available in an approachable way within the team's form anyways.

9c349da fixes some minor problems I ran into when bootstrapping the project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ Then install all dependencies:
 
 ```bash
 bundle install
-bundle exec rake db:drop db:create db:migrate
+bundle exec rake db:drop db:setup
 ```
 
 ### Mailtrap (optional)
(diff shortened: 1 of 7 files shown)


carpodaster:
Hey @klappradla - looks awesome, thank you!

As for the `Add a member` button: sure, let's remove it, but let's merge this PR first.

klappradla:
👍 I'll assign #205 to me and continue from there.

ramonh:
:tada: Thank you @klappradla!

alicetragedy:
awesomeeee!! thanks! :confetti_ball:

PR #390, opened 2016-03-07
Restructure and revamp application form

alicetragedy:
This pull request has several improvements to the application form, including:

- adding age dropdown
- adding goals and motivation text areas
- adding a programming background text area
- adding a dropdown about learning period of **language of chosen project**
- removing banking info
- adding a field for team info on working together
- revamping the look of the form by adding proper headings and replacing some radio buttons with dropdowns
- financial info as a dropdown with ranges (with additional text box for details if needed)

What is (in theory) still missing:
- programming skills haven't been changed (input needed)
- requirements as checkboxes


anikalindtner:
Awesome, thank you so much for working on this, @alicetragedy! What was the reasoning behind the drop down menue for the financial info? I'm not sure if that actually would fit the case... because if we give only let's say 1200-1300, but someone might only need 1250, wouldn't that be a bit of a pity that applicants can't specify this? 

Wdyt?

alicetragedy:
@anikalindtner as far as I remember (however, this conversation is pretty far away, so maybe someone will be able to fill in the blanks) we decided against a "set amount" and for a range, and wanted to change the question to be more general (something along the lines of "how much is needed in your city") because we felt that a) asking for students to break down costs would be too intrusive and b) we had the idea of having the students themselves look up in a calculator the living costs for their own city. 
this could be paired with an extra field where they explain any additional costs they may have (maybe due to childcare, disability issues, or rent split with a partner), and we calculate accordingly if needed / if we feel that it's necessary.

alicetragedy:
a menu with set values also means we can specify a range for the scholarships, which was important to us. ie. right now an applicant could just write $2000 if they feel like it, in a dropdown we could have $1500+ for everything above, and they'd need to specify why in a separate field. Either way we decide to go, I'd like to have one field that is specifically just for an integer (number) so that it's easy to find when looking through the applications, and stick all the extra explanations/calculations into a text field.

anikalindtner:
@alicetragedy ah yes, i remember. okay, so what we could do is have them give a monthly number in one field and post the link from http://www.numbeo.com/cost-of-living/ for their city next to it so that they'll know how we'll calculate their stipend. They can then add additional costs if their life demands are higher in an extra field. 

I think the range is a bit tricky, but probably best for us to process? So I'd say let's have all numbers from 100, 150, 200, 250, 300, 350 up until 1500 and then 1500+. We should also make clear that we use USD as a currency.

alicetragedy:
@anikalindtner okay, then I will have two fields:
a) range of USD from 100 to 1500+ (in 50 increments) with hint linking to cost-of-living site
b) a text field where they can add any additional information that might be relevant.

does that sound reasonable?

anikalindtner:
@alicetragedy sounds good! i'd actually make them paste the link to numbeo themselves into an extra text field for the city they are living in, just so they'll know what it says about their living costs and if that's different to them, they can comment on that. or is that too confusing? what do you think?

alicetragedy:
@anikalindtner I don't like the idea of having three fields for the financial information. I'd see us posting it as a hint basically a way for them to calculate that stuff on their own using the calculator ;) maybe if we need cross-checking, we can have them post the url in the text field.

anikalindtner:
@alicetragedy okay sure! i understand that! fine by me :)

alicetragedy:
awesome! thanks for all your feedback @anikalindtner, it really helped. :+1:

carpodaster:
It's a good idea to make the **application_money** field an integer one. What do you think about dropping the free-form `application_minimum_money` column from the DB? It may contain some sensitive personal data so it'd be good to flush it anyway.

I see a (small) problem with **application_language_learning_period**: we allow them to apply with two different projects and there may not be a one-size-fits-all answer for both of them.

Other than that: looks awesome! :ok_hand: :confetti_ball:

alicetragedy:
@carpodaster yes, I think it's good to stick to actual integers for that field, too. 
I'd still like to have a separate text field for financial stuff, in case they need to specify certain "special circumstances". However we can also drop `application_minimum_money` and create a brand new column (so all the info from the previous column we had is flushed) or just.. get rid of all the sensitive data we've had in there so far. Or do you disagree with this and think we should scrap it completely?
regarding the `application_language_learning_period`: yes, I briefly thought about that before adding it. I decided to still go ahead, because I felt it made give us some extra insight in the application, but it's an arbitrary decision I made and I understand how it could actually cause more havoc than anything else — happy to change or remove it (we can also redirect that to be the learning time for their primary project or whatever).

carpodaster:
Sorry Laura, I misread on of the diffs so that I thought the field would be removed from the form. That's why I suggested removing it from the database also. Nevermind then – these are totally not the droids you are looking for

carpodaster:
We have merge conflicts now, shall i (try to) resolve them manually and merge?

alicetragedy:
that would be great @carpodaster :+1:

carpodaster:
:balloon:

PR #395, opened 2016-03-10
Minor improvements on layout and links on help page

Nattivolk:
Only changes some small things to the layout and fixed the broken links.
Regarding the content I found "Your team's sources" a bit confusing and as it is saying that it doesn't do anything for now, should we remove that part?


carpodaster:
Cool, thanks @Nattivolk !

PR #398, opened 2016-03-11
Ability-restriction to prohibit students from creating a second team

klappradla:
Add ability restriction to prohibit students from creating a second team if they already got one. This also hides the `New Team` button on the team's index page if a logged in student-user already has a team (as discussed in #205).

From what I read out of the specs and other models, a user can only have one **`student`** role, so I took this as a basis to decide whether she can create a new team or not. Hope I am on the right mental model here.

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -15,6 +15,7 @@ class Ability
     can :crud, Team do |team|
       user.admin? or signed_in?(user) && team.new_record? or on_team?(user, team)
     end
+    cannot :create, Team if user.roles.student.exists?
 
     can :join, Team do |team|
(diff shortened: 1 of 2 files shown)


carpodaster:
Thank you for this, @klappradla!

A user can actually be a student in more than one team, just not within the same `Season`. Can you accomodate that in this PR?

klappradla:
Oh, didn't think that far, thanx! I'll update it.

klappradla:
Been running only the model specs for the whole time without realizing it 🙈
My brain is behind anything today, I have to get off computer screens. I'll close this PR and open a new one once I made a proper approach.

Issue #397, opened 2016-03-11
List projects


carpodaster:
* [x] A link to `/projects` (projects#index) should be displayed during the application phase
* [x] Projects should list how many applications (1st and 2nd choice) already reference them

michaelem:
I'll take a stab at this tonight.

carpodaster:
Cool, thank you @michaelem. I forgot to mention: please base your work on the `application-process-2016` remote branch. Only that branch has the foreign keys `project1_id` and `project2_id` (or some such) on `ApplicationDraft`.

carpodaster:
@michaelem I've merged the current state of `application-process-2016` into `master` just now. I plan to keep the feature/integration branch, though: it's set to auto-deploy to staging so it's probably still better to base everything against `application-process-2016`

michaelem:
Ok, I'll use application-process-2016 then!

carpodaster:
This is done, thanks @michaelem

PR #400, opened 2016-03-12
show number of applications per project

michaelem:
As discussed in #397 this adds counts for ApplicationDrafts that reference a project in the Projects index and in the Project show. Currently the state of the ApplicationDraft is not taken into account (should it be considered?).


carpodaster:
That looks might good, thank you @michaelem !

PR #404, opened 2016-03-12
Add ability rule to prevent students from creating a second team

klappradla:
Ability rule to prevent students from creating two teams for the same season.

I did not yet include anything on rules to prevent students from editing roles / edit their own rule when creating a team. Defining CanCan abilities with blocks is kind of a hassle when combined with `simple_form` and `nested_form`, making either the view or the rules incredibly ugly 🙈

But I'm sure there's a solution for it - I'll dig deeper. Until then, here's the team-rule first.

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -16,6 +16,10 @@ class Ability
       user.admin? or signed_in?(user) && team.new_record? or on_team?(user, team)
     end
 
+    cannot :create, Team do |team|
+      on_team_for_season?(user, team.season)
+    end
+
     can :join, Team do |team|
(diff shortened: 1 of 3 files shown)


app/models/ability.rb
```diff
@@ -16,6 +16,10 @@ def initialize(user)
       user.admin? or signed_in?(user) && team.new_record? or on_team?(user, team)
     end
 
+    cannot :create, Team do |team|
```
carpodaster: I wasn't aware that there is a `cannot` in CanCan(Can) :+1:

carpodaster:
Hey @klappradla – thank you! Especially for the fine-grained testing, it really works towards our oldest open issue: #6

Issue #202, opened 2015-04-15
Application form:  'Apply' action should save draft / assign new attributes before converting


carpodaster:
When a user changes data in the form field but doesn't explicitely click 'Save draft' before submitting the application, the most recent updates are lost as the controller action only calls the state machine transition without assigning potentially updated attributes.

michaelem:
Same thing happens when I change something and click 'Check Validity'. I guess this is not the desired behavior as well?

carpodaster:
Ah, good catch! Sure, this appies (tehehe…) to `Apply` as well.

I'm not sure I understand the show view idea, but I fully agree that students shouldn't guess which button does save and which one doesn't. Maybe just have all buttons save stuff?

michaelem:
I think we can probably close this since the buttons now simply disable when the form is changed.

alicetragedy:
:+1:

carpodaster:
Excellent work, @michaelem – thank you!

PR #409, opened 2016-03-13
a friendlier application draft form

michaelem:
This adds a separate tab for the application and validity buttons in the application draft form. Further it disables the 'apply' & 'check validity' if the user changes anything in the form. This takes care of #202 (in a way).


michaelem:
@alicetragedy what do you think?

alicetragedy:
This is definitely a lot better, especially when it comes to locking the validity check — thanks for this! 
I did first a quick review of the code which looked fine :+1: but when running it locally on an invalid form, clicking "Check validity" doesn't show me any error message and shows the success notice "Your application draft was saved." 
In itself, this isn't a huuuuge problem since we had some duplicate messages anyway but a) I want to find out why the error messages are not being shown anymore and b) because of how the required fields for the student, which are attached to the student model, are set up (see in `student.rb`), field validations are not shown in the "About You" tab the same way they are in the other ones. This is sort of a bummer, and I am not sure how to fix it. 😕

michaelem:
as for a) I think I know why: the button is now inside the form tag and triggers a submit of the entire form. I think I can just switch to a link though.

michaelem:
at least a) should be fixed now :)

alicetragedy:
yay!!!!! thank you @michaelem :tada: :confetti_ball:

Issue #403, opened 2016-03-12
Clicking "Apply now" fails even when I have a partner


alicetragedy:
This is something I only just noticed after pulling in the new changes onto my local / dev setup and it might be something that we will not necessarily be able to reproduce in the production environment.. but better be safe than sorry, since teams have already started forming for this year.

I have a current team (consisting of me and Sara) which I use for testing the UI and such; first and foremost, it seems like my application (which I used up until now for testing) is gone but that might be due to changes we've been doing recently. 
When the season is set to "application phase" and I click "Apply now", I get a notice that suggests I don't have a partner. My partner, however, is clearly visible in the below screenshot.

<img width="1171" alt="screen shot 2016-03-12 at 08 08 55" src="https://cloud.githubusercontent.com/assets/3143348/13721527/ac54e16e-e82a-11e5-817d-cd7e353fc10c.png">
<img width="1188" alt="screen shot 2016-03-12 at 08 10 14" src="https://cloud.githubusercontent.com/assets/3143348/13721528/ac5624e8-e82a-11e5-810e-66254190f0aa.png">

As I wrote above, this might be something that is due to the recent changes with the state machine and so on, just thought I should file it as an issue anyway — if my already existing team is impacted by the extra "guards" we've added, chances are we need to re-confirm / fix all new teams currently found in production.

carpodaster:
This sounds interesting. I'm thinking this is related to seasons. Can you check in the rails console which season is referenced by your Unicorn team? It's probably the 2015 season so I guess we're missing a season scope … _somewhere_

ramonh:
Hm, that error message is one I added for confirming teams. It seems that although you have two members, the state of the team never transitioned to `confirmed`.

Is it possible that you built the team before pulling in the new changes? This tells me in any case that older teams will run into this issue.

Maybe having a state machine is overkill and we need to just check that two students are present.

alicetragedy:
@ramonh yes, I did build the team before pulling in the changes (as outlined above already, hehe). And that's why I thought it would be significant for us maybe — as we already have teams that were created before you added the team confirmation, meaning they will run into this same problem.
@carpodaster the season that was referenced was first another, older, season (which, by the way, I wasn't finding anymore) and I've now changed it locally to the current one. this has obviously not helped confirm the team though, so I have manually confirmed it to get around it.

Maybe we should keep this issue open — we could even test this particular issue during testing tomorrow.

ramonh:
[...]

PR #417, opened 2016-03-14
Replace AASM confirmation with dynamic #confirmed? method

ramonh:
This aims to address #403.

Instead of saving the state in the database, we'd instead dynamically check if a team is confirmed by checking if two students are present.

The API stays the same, however.

### DIFF
--- a/spec/models/team_spec.rb
+++ b/spec/models/team_spec.rb
@@ -155,14 +155,12 @@ describe Team do
     let(:role_name) { 'student' }
     it 'will not be confirmed if only one student present' do
       team.attributes = { roles_attributes: [{ name: role_name, user_id: first_student.id }] }
-      team.save!
-      expect(team).not_to be_confirmed
+      expect { team.save! }.not_to change { team.confirmed? }.from false
     end
 
     it 'will confirm team through role assignment when second student present' do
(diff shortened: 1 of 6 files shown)


spec/models/team_spec.rb
```diff
@@ -148,31 +148,24 @@
...
-  describe '#state' do
+  describe '#confirmed?' do
     let(:first_student) { create(:user) }
     let(:second_student) { create(:user) }
     let(:team) { FactoryGirl.create(:team) }
     let(:role_name) { 'student' }
-    it 'returns "pending" when only one student present' do
+    it 'will not be confirmed if only one student present' do
       team.attributes = { roles_attributes: [{ name: role_name, user_id: first_student.id }] }
-      expect { team.save! }.not_to change { team.confirmed? }
```
carpodaster: I'd prefer the the `expect { }.to … ` syntax since that puts a object state-change under test. Can you change this to:

```ruby
expect { teanm.save! }.not_to change { team.confirmed? }
```

? Thank you!

(also the example below)

ramonh: Hey @carpodaster! Makes sense, thank you for the feedback! I'll fix this and squash it in.

carpodaster:
Thanks @ramonh !

ramonh:
And thank you for your help!

PR #418, opened 2016-03-14
show how many errors each tab has in the application draft form

### LINKED ISSUE #408: UI: Rework Error Message
The error messages should not be a huge text field as it is now, but just show how many errors you have and where. All error messages should be shown directly in the fields where the error accurs (this is already the case for *some* errors). 

### My proposal: 

+ [ ] all errors need to be shown directly in the field where the error occurs 
+ [ ] error messages should tell you **where**, that means **in which tab the info is missing**. This can be either:
  + in the summary:
" 
### About yourself 
error message
### About your project 
error message
"

  + or shown in the tabs directly (e.g. by having a red border around the ones who have error messages, so you know where to look)
screen shot 2016-03-13 at 20 00 28

### MUST FIX::

+ [x] "4 errors to go before you can apply:" <- this is invisible (red on red).

### NICE TO HAVE

+ [x] change black on red writing to black on white writing and add only red border / or change font to red on white without a border.

michaelem:
The downside of this is that the helper methods now need to know which tab has wich fields. (So if someone changes the template, they need to update the helper as well.) Fixes #408.

### DIFF
--- a/app/helpers/application_drafts_helper.rb
+++ b/app/helpers/application_drafts_helper.rb
@@ -1,4 +1,20 @@
 module ApplicationDraftsHelper
+  STUDENT_FIELDS = [:name,
+                    :application_age,
+                    :application_gender_identification,
+                    :application_location,
+                    :application_about,
+                    :application_code_background,
+                    :application_skills,
+                    :application_community_engagement,
+                    :application_learning_period,
+                    :application_language_learning_period,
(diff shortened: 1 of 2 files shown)


alicetragedy:
whaaaaaa this is so great!! thank you so much @michaelem :sparkles: 

batman_small

Issue #192, opened 2015-04-12
Radio buttons in application form do not correspond to user profile data


mymai91:
In the part "Show application specific information" I checked fields 
Please rate your level of programming experience (1 being the 'lowest' and 5 the 'highest').
For how many months have you actively been learning after your first workshop?
But the result doesn't show on our application. I don't know why.
This is my profile:
![my-profile](https://cloud.githubusercontent.com/assets/6791942/7103857/f884d844-e0e4-11e4-9306-b5b84f6671db.png)

This is my team application:

![team1](https://cloud.githubusercontent.com/assets/6791942/7103853/e139b772-e0e4-11e4-9137-78b82876352e.png)

Currently, I'm using Chrome Version 41.0.2272.104 (64-bit)

Could you help me check it?

carpodaster:
Hello @mymai91 – we are aware that this situation occurs for some applicants. Do the checkboxes show up properly for your pair, or is it just your fields that are not pre-set with values from your user profile.

To assure you: if the values are checked in your user profile, you're all good. It's weird that they don't show up in the application draft, and I have yet to get to the bottom of this, but the application data will be collected from your user profile once you hit 'Submit'.

mymai91:
@carpodaster: I got it. Thank you.

carpodaster:
This has been fixed for this year's application process! :confetti_ball:

PR #422, opened 2016-03-15
Make student input fields' name and id attributes unique

carpodaster:
This PR fixes long-timer #192 in my local setup. It's a simple patchset but I would appreciate an extra pair of eyes to double-check if this introduces a privilege escalation by means of url- or form input forging.

/cc @klappradla @cypher @ramonh @michaelem 

_Edit_: Some background, copied from yesterday's Slack:

> So, what I noticed yesterday: due to the fact that we render the `_student_fields` partial for both students, we get duplicate input names and DOM ids.
> 
> I’m guessing that’s why the form doesn’t preselect the skill level (remember: it ​_is_​ saved in the DB, it just doesn’t check the saved value in the form). And because the form fields are not named like the the corresponding attributes, the form helper doesn’t know which input field to highlight as erroneous.

### DIFF
--- a/app/views/application_drafts/_student_fields.html.slim
+++ b/app/views/application_drafts/_student_fields.html.slim
@@ -1,4 +1,4 @@
-= simple_fields_for (student || Student.new) do |s|
+= simple_fields_for (student || Student.new), index: student.id.to_i do |s|
 
   h2 Basics
   = s.input :name, label: 'Student Name', hint: 'Please provide your first and last name', input_html: { disabled: !may_edit?(s.object) }
--- a/app/controllers/application_drafts_controller.rb
+++ b/app/controllers/application_drafts_controller.rb
@@ -86,10 +86,8 @@ class ApplicationDraftsController < ApplicationController
   end
 
   def student_params
-    # TODO make sure we are the student set to be updated
     if application_draft.as_student? and params[:student]
-      # TODO: Do we need an index? Maybe just compare id with current_student.id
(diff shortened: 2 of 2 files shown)


app/views/application_drafts/_student_fields.html.slim
```diff
@@ -1,4 +1,4 @@
-= simple_fields_for (student || Student.new) do |s|
+= simple_fields_for (student || Student.new), index: student.id.to_i do |s|
```
carpodaster: This makes the `params` hash look like this:

```
# given that student's id is 42:
{ 'student' => { '42' => { 'foo' => 'bar', … }, 'application_draft' => { ... } }
```

klappradla:
Introducing the ID-index, you end up with the same type of params hash, you'd have with normal rails `fields_for` helpers and nested attributes: an array that gets converted to a hash like (because that's how arrays are passed as params afaik):
```ruby
{ 'students' => { '0' => { 'foo' => 'bar', … }, '1' => { 'foo' => 'bar', … }...}
```
Since fetching the params is bound to the `current_user`, I don't see a problem there.

carpodaster:
Thank you, @klappradla, for your feedback and for looking into this!

alicetragedy:
omg. it's fixed? IT'S FIXED?!?!?!!

taylor-dancing-02

anikalindtner:
whoa!!!! 

confetti_dog

PR #423, opened 2016-03-15
Streamline error indication look'n'feel for students' fields

carpodaster:
~~ **Note: This PR is based on #422 and has to wait until that one is merged.**~~  (*Update: Merged* :white_check_mark: )

It's a really hacky thing and needs some cleaning up in a follow-up session. It also adds blank select box options (see the commit msg in b9aec581f39c38a6d49953a92f9f9424992ed96a for its rationale). This is what it looks like now:

bildschirmfoto 2016-03-16 um 00 23 25
--------

bildschirmfoto 2016-03-16 um 00 23 41


alicetragedy:
seriously I'm at the point where I don't really care how hacky it is ;)
I'm happy to merge this!!

On another note, do you think it might make sense to remove either the "There are still some fields missing" or the full list of errors, if we have the indications on the tabs and the input fields themselves? I know that you're a pro-error-list, so I don't want to insist, if you think that might be better from a UX point of view.

carpodaster:
I've moved from the `pro-error-list` into the `indifferent` area ;)  I will remove it!

alicetragedy:
smalltypingcat

Issue #419, opened 2016-03-14
Application submission fails in a weird way


mkalininait:
If my coaches are not confirmed and I submit an application, I don't get any error messages on the Application form. Instead I get redirected to the page of my team.

![screen shot 2016-03-14 at 23 14 04](https://cloud.githubusercontent.com/assets/1745735/13762101/04787450-ea3d-11e5-8eee-5e3d984fb181.png)
  
My application is still available for editing and submitting.

Another thing you can see on this screenshot is that I have two coaches which are marked as confirmed, and I still see the message, that the coaches are not confirmed. We should probably restrict adding coaches, who are already students in this (or another) team.

alicetragedy:
yep, you are right. I guess this has to do with the student being a coach.

ramonh:
Definitely a bug having students be coaches of their own team, thanks for catching this!

I thought it would be helpful for students to see which of their coaches hasn't confirmed yet. Do you think it's better not to redirect to the team's page?

alicetragedy:
@ramonh maybe an alternative would be an error message "Your coaches have not confirmed their membership" with a link to the specific team in there? so the student can go and check out what the deal is, but only if they want to ;)

ramonh:
Yeah, that makes more sense! Thank you @alicetragedy

ramonh:
The above PR should take care of the redirecting. About the second problem, I could restrict students from also being coaches. Of course, if they were students last season and want to be coaches this season, this should be fine. I'll look into whether this is fine.

alicetragedy:
@ramonh did you get a chance to look into this?

ramonh:
Hey @alicetragedy, I did start but haven't finished yet! Will have a pr ready for tomorrow, hopefully.

carpodaster:
@ramonh @mkalininait Did the PR  (merged and now live on staging) fix the issue?

ramonh:
It should be fixed, @carpodaster. I've tried adding myself as a student and a coach to my team, and had no luck :) 
![screen shot 2016-03-18 at 10 31 25](https://cloud.githubusercontent.com/assets/656318/13873878/a4bc9338-ecf4-11e5-9d19-c5e6a32da235.png)

alicetragedy:
looks good, I think we can close this.

PR #425, opened 2016-03-16
Block user from being multiple members of a team

ramonh:
This aims to address the second half of #419, where students could list themselves as coaches of their own team.

One advantage that comes from this is that it also doesn't allow a student to be both students in a team!

### DIFF
--- a/app/models/team.rb
+++ b/app/models/team.rb
@@ -8,6 +8,7 @@ class Team < ActiveRecord::Base
   validates :name, uniqueness: true, allow_blank: true
   # validate :must_have_members
   validate :disallow_multiple_student_roles
+  validate :disallow_duplicate_members
   validate :limit_number_of_students
 
   attr_accessor :checked
@@ -150,6 +151,14 @@ class Team < ActiveRecord::Base
     return if students.empty?
     msg = MSGS[:"duplicate_student_roles_#{students.size == 1 ? 'singular' : 'plural'}"]
     errors.add :base, msg % students.map(&:name).join(', ')
+  end
+
+  def disallow_duplicate_members
+    new_members = roles.map { |role| role.user }
+    duplicate_role_users = new_members.find_all { |member| new_members.count(member) > 1 }.uniq
+    return if duplicate_role_users.empty?
+    msg = "%s can't have more than one role in this team!"
+    errors.add :base, msg % duplicate_role_users.map(&:name).join(', ')
   end
 
   def limit_number_of_students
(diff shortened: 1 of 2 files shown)


alicetragedy:
I have a meeting soon, so will take a lookie later (if someone doesn't get there before me). 
Thanks in advance for this! I'm so excited!!!

alicetragedy:
@ramonh I don't know why, but this still lets me add myself as a coach on the same time as I am a student via the interface. It also lets me add myself to that same team as a mentor, etc.. via the separate "Add a member" interface. :confused: 
The first might have something to do with the confirmation (maybe `has_many :members, -> { uniq }, class_name: 'User'` doesn't take into account the non-confirmed role as a role to double-check for uniqueness of name) but the second, I have absolutely no idea.

ramonh:
Well that's no good, sorry about that! I'll look into why this is taking place. I'm afraid I won't have any further time tonight but will tomorrow, if that's okay

ramonh:
But thank you so much for looking into this, @alicetragedy

alicetragedy:
no worries, I am just super sorry I only got to looking at this now. should have done this first thing in the morning. :(

ramonh:
Not at all! I shoulda tested this better :sweat:

alicetragedy:
If it can help, maybe adding a private method in the team similar to `disallow_multiple_student_roles` and validating that upon creation of the team member / team update would help, that's all that comes to mind right now.

ramonh:
Thanks for the suggestion! I thought that unique members might've been elegant, but alas! (I'm trying to squeeze in a quick look now :sweat_smile: )

ramonh:
Ok tiny update: It seems to be working! I'll need to set up the error message, clean up some tests so that it looks like I TDD'ed it (koff koff!). And then we're good to go. I'll have to do that later on, though. Thanks again, @alicetragedy

alicetragedy:
perfect!!!!!! :+1:

alicetragedy:
This looks good/better now. Not sure if you want to double-check @carpodaster but I'm ready to MERGE ALL THE THINGS :boom:

carpodaster:
> clean up some tests so that it looks like I TDD'ed it (koff koff!)

:joy:

carpodaster:
:+1: Go ahead and merge it – i haven't paid much attention to this PR and its discussion and I feel super-bad about it.

It's awesome that we have this unique-check now, thanks @ramonh. At the same time – and like I said on Slack earlier today – I still believe that there are probably a hundred different way of getting a team into a borked state, we might as well just roll with it :trollface: #yolo

alicetragedy:
@carpodaster as a wise woman once said,

"we can fix it in production" :joy:

ramonh:
Agreed, there will always be weak points that will come up. Thank you for your help!

PR #442, opened 2016-03-21
Distinguish between drafts and actual applications for project counters

### LINKED ISSUE #440: Change appliaction count in projects view to only count submitted applications
For now - as I understand it - the applications for a project are counted when someone creates a draft. I'd say we should only count submitted applications.

Reason: last year showed that a lot of drafts are not even going to be sent off and thus a higher count on projects might scare some teams away, but in the end it could turn out that actually no one applied for this project, because it was just all drafts.

carpodaster:
Attempts to fix #440. It also adds some tooltips.

bildschirmfoto 2016-03-21 um 17 27 27


alicetragedy:
looks good!!! :tada:

Issue #399, opened 2016-03-11
Attendance must be "seasoned"


carpodaster:
The `Attendance` (conference feature) does not yet have a reference to the `Season` it is valid for.

F3PiX:
Is this still needed? Or do we go with the current way of 'showing the conferences during the active season'?

carpodaster:
I think we can close this – at least for now. Conferences are seasoned and by that, the season of an attendance can easily be infered.

PR #444, opened 2016-03-23
conference fixes

alicetragedy:
I know this is a stupid quick fix, but the conferences in the teams overview are quite confusing for students — especially if past attendees of conferences belong to a team, as the conferences are not "seasoned" yet — so I'm commenting them out until we find a better way to show them based on the current season (see https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/399)

This also adds consistent styling to the back navigation and all the buttons in a conference show view. 🎉

### DIFF
--- a/app/views/conferences/show.html.slim
+++ b/app/views/conferences/show.html.slim
@@ -1,14 +1,16 @@
 nav.page
-  ul
-    li = link_to 'All conferences', conferences_path, class: 'back'
-
-- if can?(:crud, Conference)
-  nav.actions
+  .back-nav
     ul
-      li = link_to 'Edit', edit_conference_path(conference), class: 'edit'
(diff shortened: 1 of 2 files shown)


alicetragedy:
@carpodaster is it better to actually delete the code rather than comment it out? I could drop a quick `rake note` in there so we can easily track why it was commented out in the first place, but I'm not sure if it's possible to add this to views..

carpodaster:
@alicetragedy I think leaving it in is perfectly fine in this case. But yeah, adding a `TODO` (or something), sounds like a good idea, too

F3PiX:
Can we close this?

alicetragedy:
@F3PiX how do you mean? A merged PR is automatically "closed" — or do you mean the related issue? I'm not sure if there is one, but I can look. I often forget to close those when I merge a PR 🙈

F3PiX:
Sorry, my mistake 🐒

Issue #445, opened 2016-03-23
The applications list shows all applications instead of those for the current season


ramonh:
Going to the list of applications (the path is `applications/todos`) shows a list not just of the current season, but all of them.

ramonh:
Works great as of #447. Thank you @althaire for your help! :tada:

PR #447, opened 2016-03-23
Filter out teams not from current season when reviewing

Althaire:
This addresses #445 by only listing teams from the current season when reviewing.
:octopus: :sparkles:


alicetragedy:
yayyy!!!!! thank you so much @Althaire :confetti_ball: :tada:

Issue #294, opened 2015-07-24
conferences area: lightning talks


alicetragedy:
In the conferences area, it would be useful to have a field that shows whether the conference allows/has lightning talk slots or not (a checkbox would be enough) – this would cut down on work when emailing the students selected for the conferences and having to look up manually if the conferences offer lightning talk slots.

PR #452, opened 2016-04-10
Worked on Lightning Talk Slots checkbox in conferences display page and form (Issue #294)

roxxx33:
@carpodaster 
I have added Lightning Talk Slots checkbox to new conference form and also a field for displaying the same in the list of conferences page.
new_conference_form
conferences_display
display_individual_conference
Made changes in the following files:
1. db/schema.rb
2. conferences_controller.rb
3. views/conferences/_form.html.slim
4. views/conferences/show.html.slim
5. views/conferences/index.html.slim

### DIFF
--- a/app/views/conferences/_form.html.slim
+++ b/app/views/conferences/_form.html.slim
@@ -16,7 +16,7 @@
   = f.input :flights
   = f.input :round
   = f.input :accomodation
-  = f.input :lightningtalkslots, input_html: {value: nil}, label: "Lightning Talk Slots"
+  = f.input :lightningtalkslots, label: "Lightning Talk Slots"
 
   h3 Attendees
   fieldset.attendees
--- a/app/views/conferences/_form.html.slim
+++ b/app/views/conferences/_form.html.slim
@@ -16,7 +16,7 @@
   = f.input :flights
   = f.input :round
   = f.input :accomodation
-  = f.input :lightningtalkslots, input_html: {value: false}, label: "Lightning Talk Slots"
+  = f.input :lightningtalkslots, input_html: {value: nil}, label: "Lightning Talk Slots"
 
   h3 Attendees
   fieldset.attendees
--- a/app/views/conferences/_form.html.slim
+++ b/app/views/conferences/_form.html.slim
@@ -16,6 +16,7 @@
   = f.input :flights
   = f.input :round
   = f.input :accomodation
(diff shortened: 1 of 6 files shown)


app/views/conferences/_form.html.slim
```diff
@@ -16,6 +16,7 @@
   = f.input :flights
   = f.input :round
   = f.input :accomodation
+  = f.input :lightningtalkslots, input_html: {value: false}, label: "Lightning Talk Slots"
```
carpodaster: You don't have to assign `false` here. The default value is `nil` (which is »falsey«). In fact, forcing the value to `false` will overwrite the previous value when a conference is edited at a later time.

carpodaster:
Hey @roxxx33 - thanks! I've added a comment on a commit. In addition to that, this PR doesn't include the migration file that adds the new database column; it only contains the updated `schema.rb`.

roxxx33:
@carpodaster I have made the necessary changes. Could you please review it.

app/views/conferences/_form.html.slim
```diff
@@ -16,6 +16,7 @@
   = f.input :flights
   = f.input :round
   = f.input :accomodation
+  = f.input :lightningtalkslots, input_html: {value: nil}, label: "Lightning Talk Slots"
```
carpodaster: The `value: nil` is still here. See my previous comment on that line for the reasoning behind it.

app/views/conferences/index.html.slim
```diff
@@ -21,6 +22,7 @@ table.table.table-striped.table-bordered.table-condensed.conferences
       td = format_conference_date(conference.starts_on, conference.ends_on)
       td = conference.attendees.any? ? links_to_attendances(conference).join(', ').html_safe : '-'
       td = conference.tickets_left
+      td = conference.lightningtalkslots
```
carpodaster: I've just realised: this would display a `true` / `false` (or nothing at all) to the user. Could you improve on that by displaying a »yes« (or »no«, respectively)?

roxxx33:
@carpodaster 
I have changed the view files views/conferences/show.html.slim and views/conferences/index.html.slim. So, it display "yes" for "true" value in the field and "no" otherwise.

carpodaster:
@roxxx33 just the `value: nil` now (see comment) and we're good to merge :)

roxxx33:
@carpodaster removed "value: nil".

carpodaster:
Merged – thanks @roxxx33 !

Issue #414, opened 2016-03-13
Applicants should not be able to add more than 3 conferences


Nattivolk:
Conferences, it says an applicant should only add max 3 conferences. I could add many more conferences. If 3 is a hard limit, maybe we should not allow adding more?

Nattivolk:
Probably on hold due to https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/411
But important as soon as we open the conferences again afterwards

alicetragedy:
:+1:

alicetragedy:
For now I've set it so that editing the conference stuff in the profile is not visible until the season starts (July 1st) so we have a little bit of time to get it fixed. thanks for noticing & creating the issue!

F3PiX:
Copied from PR 414
I worked on a solution that:
- sets a limit to the number of permitted attendances (in the `accepts_nested_attributes_for` method)
- Hides the 'add conference' button for users who has already 3 attendances saved to their profile   
- Adds an flash alert when users wants to save their profile with more than 3 conferences.
The count of the attendances is updated only when the parent is saved. This means an user is able to add 4 conferences in the Edit Profile form, and will not be notified that 3 = max until after submit.
This is not super user friendly.

I stopped working on it, because but I am having second thoughts about the server side. It is making the code less flexible and more rigid, for no good reason.
Now I think it would be better to solve this at the client side with some nifty javascript. The current 'Add button' is probably built in by the nested_form gem. It must be possible to make it check for the number of attendances as well.
 
Feedback and suggestions are very welcome.

alicetragedy:
Note: I've removed this from the application process as I think it's not super important for now. It's enough if this gets solved before the start of the program!

carpodaster:
Makes sense, thank you! 💟

klappradla:
Closing this as this is solved via introducing the `ConferencePreference` model.

<img width="1035" alt="screen shot 2017-08-20 at 16 30 10" src="https://user-images.githubusercontent.com/3491815/29495750-7d829cf2-85c5-11e7-9d31-cc4d5615d296.png">

carpodaster:
🎊

PR #453, opened 2016-04-10
Limiting the number of Conferences(Issue414)

pratyusha972:
Hey @carpodaster ,I have put a condition check on number of conferences limiting it to a maximum number of 3. A note will be displayed if an applicant tries to add more than 3 conferences.
conf3
This image shows the instance when there are 3 conferences.
2conf3
This image shows the instance when the new conference button is clicked when there are already 3 conferences, i.e a note is displayed and it is redirected to the same page.

### DIFF
--- a/app/controllers/conferences_controller.rb
+++ b/app/controllers/conferences_controller.rb
@@ -1,6 +1,14 @@
 class ConferencesController < ApplicationController
   load_and_authorize_resource except: [:index, :show]
 
+  def new
+    if Conference.count == 3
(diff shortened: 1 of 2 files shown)


alicetragedy:
Hi @pratyusha972, we don't want to limit the number of conferences that can be added — we want to limit the number of conferences which attendees can apply to in each conference raffle round. Sorry if this wasn't clear from https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/414.

carpodaster:
What @alicetragedy said… Sorry that this wasn't clear from the issue description :cry:

PR #470, opened 2016-06-20
display status in teams index

michaelem:
Shows the same status icons as in the supervisor dashboard to admins in the /teams view.
The change is simple enough but is there any convention where to put shared partials?
Since they are used in another view now ```supervisor/dashboard/``` seems like an odd location for the performance indicating partials.


carpodaster:
> The change is simple enough but is there any convention where to put shared partials?

Not yet, no. Maybe add a `shared` subfolder within every namespace and put partials not related to a resource in there?

michaelem:
@carpodaster since that Is my usual approach I did that now.

Issue #256, opened 2015-07-09
Automatic reminder for daily log


anikalindtner:
As we talked about in our weekly organizer hangout yesterday, a lot of what the supervisors are doing is nagging teams to get stuff done. Which wears them down and shouldn't be the main task of our supervisors. So I thought about how to make at least *some* of the process automated to take it out of their hands. And one thing would be to send an email reminder at hour xy to the team if they haven't filled out their daily log. This way we nag them without having to do it ourselves and they get accustomed to doing it everyday from the beginning on. 

Exception would have to be weekends. 

@sareg0 was interested in helping to think about how we could implement something like that. 

Comments? Opinions?

carpodaster:
Yeah, we have talked about this enhancement to the minilog feature before and I think it's a good one.

Possible gotcha: The reminder should take the team's timezone into account.

alicetragedy:
@carpodaster yep, I mentioned the thing with the timezone briefly during the call when this was discussed :) but I also think that it would be a great way to annoy everyone to have them do their log so +1 on this

idengager:
This is a very good idea. :+1:

sareg0:
yeah, I really like this idea :) +1 from me as possible new feature. I think it would be super cool to work on, but if it happens to be something the team would like to work on, that's also cool :)

F3PiX:
I was thinking: with the new Supervisor Dashboard, it would be quite easy to add an automatic mailer function/button. The supervisor would be the one to initiate the action, and the mail itself could be sent with the orga-team as the sender. Ideally, it would count the warnings as well.
Wdyt?

sareg0:
it sounds interesting for sure, particularly since the supervisor can have a bit of a running record of how many reminders have been sent.
Re. the warnings though, the orga team makes a decision on sending those, so it feels like a different concern to being in the  supervisor's dashboard.

F3PiX:
Aha, I see: there are reminders and there are warnings. So only the reminder should be sent by the supervisor, cc orga.

anikalindtner:
@michaelem and i will try to tackle this. would be helpful to have a data dump for that. @alicetragedy could you help on this?

michaelem:
I would put this in a rake task that could run once a day and checks all the teams, sends emails to those who did not put out an update. Do we already have any tasks that run on a schedule?

Stackoverflow says we should use this: https://elements.heroku.com/addons/scheduler
[...]

PR #471, opened 2016-06-20
FIX: performance evaluation should not change when called multiple times

michaelem:
This fixes a bug (that was probably not occurring in the app directly) where each time evaluation was called the resulting score would be added to the last score (so after a few calls it would be higher than it should be). Since it's mostly your code @F3PiX, does this look fine to you?

(I found this while building the log update reminder in #256 which I'll add to this class as well, that is the reason why I also moved the self.buffer_days? method to a class method.)

### DIFF
--- a/app/models/team_performance.rb
+++ b/app/models/team_performance.rb
@@ -31,30 +31,34 @@ class TeamPerformance
 
   def comments_score
     latest_comment = @team.comments.ordered.first
-    if @team.comments.empty?
-      @score += 3 unless self.class.buffer_days?
-    elsif latest_comment.created_at <= Time.now-5.days
-      @score += 2
-    elsif latest_comment.created_at <= Time.now-2.days
-      @score += 1
-    elsif latest_comment.created_at > Time.now-2.days
-      @score += 0
-    else
-      @score += 1
+    if !self.class.buffer_days?
+      if @team.comments.empty?
+        @score += 3
+      elsif latest_comment.created_at <= Time.now-5.days
+        @score += 2
+      elsif latest_comment.created_at <= Time.now-2.days
+        @score += 1
+      elsif latest_comment.created_at > Time.now-2.days
+        @score += 0
+      else
+        @score += 1
+      end
     end
   end
 
   def activity_score
-    if @team.activities.empty?
-      @score += 3 unless self.class.buffer_days?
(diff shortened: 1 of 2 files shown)


carpodaster:
@michaelem I know there hasn't been a test to begin with, but would you mind adding one?

michaelem:
yes, will do so today (THB I was just too tired yesterday 🙈)

michaelem:
Added a spec for the class, I'm not that handy with rspec yet (I am more used to minitest), if anyone has suggestions for improvements please let me know.

klappradla:
Hey @michaelem, from a quick look at your tests (I might be wrong), it seems like you're using the *should*-syntax as opposed to the *expect*-syntax RSpec [introduced some time ago](http://rspec.info/blog/2012/06/rspecs-new-expectation-syntax/) and recommends to use. It's quite easy to [switch to this syntax](https://github.com/rspec/rspec-expectations/blob/master/Should.md) though. If your browser is not used to it, googling for the syntax tends to bring you to the old docs.... [these](https://www.relishapp.com/rspec/rspec-expectations/docs/built-in-matchers) are the current ones.

michaelem:
@klappradla omg, yes, I did not realise I was using the old syntax. Shouldn't (haha) I get warnings by now for doing so? I'll dig out transpec tonight.

Additional though: I'd disable the should syntax entirely?

klappradla:
Well, they just *encourage* you to use the expect-syntax, not **force**, since there seem to be some cases where you could prefer it... so maybe that's why there are no deprecation warnings. It will at some point maybe extract into a gem - so they kinda seem to want to get rid of it 😉 

Disabling it entirely, don't know if really necessary, but you could give it a try:
```ruby
RSpec.configure do |config|
  config.expect_with :rspec do |c|
    c.syntax = :expect             # disables `should`
    # or
    c.syntax = :should             # disables `expect`
    # or
    c.syntax = [:should, :expect]  # default, enables both `should` and `expect`
  end
end
```

michaelem:
Changed to the new syntax (and amended to the old commit).

F3PiX:
I am VERY happy that you improved this. 🙇  And even more that you added tests...

carpodaster:
Looks good! ✨

PR #478, opened 2016-06-29
More and better seeding

F3PiX:
Issue #225 
Adds seeds for users, teams and status updates.
First try for Application Drafts and Projects.

- [x] An unaccepted team
- [x] An accepted team
- [x] An accepted, but voluntary team
- [x]  Teams from last season
- [x] Status updates per team  
- [x] An application draft, unsubmitted
- [x] An application draft, completed
- [x] A proposed project
- [x] An accepted project
- [x] A rejected project

### DIFF
--- a/app/models/season.rb
+++ b/app/models/season.rb
@@ -13,6 +13,10 @@ class Season < ActiveRecord::Base
       find_or_create_by(name: (Date.today.year+1).to_s)
     end
 
+    def prev
+      find_or_create_by(name: (Date.today.year-1).to_s)
+    end
+
     # Project proposals open early. This predicate tells us if
     # we are in the transition phase after the current season's
     # coding period but before the new year (i.d. 'Season') begun.
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -1,3 +1,7 @@
+FactoryGirl.create_list(:team, 5, :current_season, kind: "sponsored")
+FactoryGirl.create(:team, :current_season, kind: "voluntary")
+
+
 FactoryGirl.create_list(:student, 12)
 FactoryGirl.create_list(:coach, 3)
 FactoryGirl.create_list(:organizer, 2)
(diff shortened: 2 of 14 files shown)


F3PiX:
Question. The application-drafts and projects doesn't seem to show up in dev-env. I used existing factories, and in the console, both the applications and the projects render data. Am I doing something wrong here?

db/seeds.rb
```diff
@@ -1,11 +1,25 @@
...
 FactoryGirl.create(:supervisor)
 
 # To explore use cases where user has no role yet
-# NB These are not listed in the Community listing
-FactoryGirl.create_list(:user, 6)
+FactoryGirl.create_list(:user, 3)
+
+#Create status updates for different teams
+5.times do
+FactoryGirl.create(:status_update, published_at: Time.now, team_id: (0..Team.all.count).to_a.sample)
```
carpodaster: Can you use `create_list` here as well?

carpodaster: Assignment of a random team: you can add a Team instance as well (instead of assigning it by id): `team: Team.all.sample`

F3PiX: As far as I know, a list will create 5 status_updates for one team. I opted for 5 status updates for different teams. Okay?

db/seeds.rb
```diff
@@ -1,11 +1,25 @@
+FactoryGirl.create_list(:team, 5, :current_season, kind: "sponsored")
+FactoryGirl.create(:team, :current_season, kind: "voluntary")
+
```
carpodaster: Idea: adding a sponsored/voluntary team from the last season?

db/seeds.rb
carpodaster: Oh, you might be right on this one!

carpodaster:
> The application-drafts and projects doesn't seem to show up in dev-env

Where did you check?

app/models/season.rb
```diff
@@ -13,6 +13,10 @@ def succ
       find_or_create_by(name: (Date.today.year+1).to_s)
     end
 
+    def prev
```
F3PiX: I don't like this solution very much. Wdyt @carpodaster?

F3PiX:
Sorry for the mess with other commits. I take it, as there are no conflicts with the base branch that it is not too bad. Or shall I close this one and start over?

carpodaster:
Good morning @F3PiX – tbh it confuses me why all the old commits show up in the diff. This PR is based against master, and even if you merged in a recent version of the master, the diff shouldn't be that noisy. It would be cool if you could start a fresh PR, yeah. Sorry for the inconvience.

Issue #476, opened 2016-06-29
Mailer content confuses roles


F3PiX:
When I assign myself as a Supervisor, a confirmation mail is sent by the mailer. In the body of the mail it says correctly that I am a Supervisor, but in the second part it adresses me as a coach.
This part of the mail is supposed to show only if role == "coach".
I don't see why it is showing up with the supervisor role as well. Anybody? Could it be misplaced slim syntax?

Mail: "You have been added to the 6-melting as a _supervisor_.

Visit the team at http://localhost:3000/teams/37
Please confirm that you are indeed willing to join the team as their _coach_. You can do so on the team's overview page listed above (make sure you are logged in using your Github account)."

This is the conditional:
https://github.com/rails-girls-summer-of-code/rgsoc-teams/blob/master/app/views/role_mailer/user_added_to_team.html.slim#L16
in this file:
rgsoc-teams/app/views/role_mailer/user_added_to_team.html.slim

carpodaster:
Hello @F3PiX, thanks for reporting this. The email I got had the same error. I just checked the code and couldn't find any apparent errors. I also cannot reproduce it in my local dev setup.

My initial idea was what the roles have mistakingly been added as coach-roles first (which would trigger the after create callback) and then later edited to supervisor or mentor. That's impossible, though, as the mail body correctly states that one has been added as a supervisor/mentor.

F3PiX:
By "the email you got" you mean the *real* email you received to confirm your *real* mentorship? Bummer. I hoped it only was a Factory-problem. Anyway, after your commit, the issue is still there when I create user seeds with FactoryGirl.

F3PiX:
Yeah, I hate Slim. I think it was an indentation problem.

PR #480, opened 2016-07-02
Improve confirmation email for non-coaches

F3PiX:
Supervisors and mentors were incorrectly adressed as a coach in the mailer that confirms their role.
  

Bug #476

### DIFF
--- a/app/views/role_mailer/user_added_to_team.html.slim
+++ b/app/views/role_mailer/user_added_to_team.html.slim
@@ -9,12 +9,12 @@ html
       | }
   body
     p
-      ' You have been added to the
+      ' You have been added to Team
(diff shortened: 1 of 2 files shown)


carpodaster:
Hah! I just realized what the bug related to #476 is! I completely ignored the fact that we have two mail view templates: one for text-only and another for HTML mails.

That's what I had the coach-paragraph in my mailbox (I have a [text-only mail client](http://www.mutt.org)) but could reproduce it using the mailer previews (which default to HTML).

Could you change the text-only template so that it also includes the `if @role.name == "coach"`  condition? That should totally fix #476

carpodaster:
Thank you, @F3PiX !

PR #486, opened 2016-07-06
Seeding

F3PiX:
A fresh and decluttered PR

Issue #225 
Adds seeds for users, teams and status updates; application-drafts and projects.

- [x] An unaccepted team
- [x] An accepted team
- [x] An accepted, but voluntary team
- [x]  Teams from last season
- [x] Status updates per team  
- [x] An application draft, unsubmitted
- [x] An application draft, completed
- [x] A proposed project
- [x] An accepted project
- [x] A rejected project

### DIFF
--- a/app/models/season.rb
+++ b/app/models/season.rb
@@ -13,10 +13,6 @@ class Season < ActiveRecord::Base
       find_or_create_by(name: (Date.today.year+1).to_s)
     end
 
-    def prev
-      find_or_create_by(name: (Date.today.year-1).to_s)
-    end
-
     # Project proposals open early. This predicate tells us if
     # we are in the transition phase after the current season's
     # coding period but before the new year (i.d. 'Season') begun.
--- a/app/models/season.rb
+++ b/app/models/season.rb
@@ -13,6 +13,10 @@ class Season < ActiveRecord::Base
       find_or_create_by(name: (Date.today.year+1).to_s)
     end
 
+    def prev
+      find_or_create_by(name: (Date.today.year-1).to_s)
+    end
+
     # Project proposals open early. This predicate tells us if
     # we are in the transition phase after the current season's
     # coding period but before the new year (i.d. 'Season') begun.
--- a/spec/factories/teams.rb
+++ b/spec/factories/teams.rb
@@ -11,6 +11,10 @@ FactoryGirl.define do
       season { Season.current }
     end
 
+    trait :last_season do
+      season { Season.prev }
+    end
+
     trait :supervise do
       name 'supervise'
     end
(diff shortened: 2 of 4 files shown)


app/models/season.rb
```diff
@@ -13,6 +13,10 @@ def succ
       find_or_create_by(name: (Date.today.year+1).to_s)
     end
 
+    def prev
+      find_or_create_by(name: (Date.today.year-1).to_s)
+    end
```
F3PiX: I don't like this solution very much. Wdyt @carpodaster?

carpodaster: I thought we already had a `Season.prev`. Since it's only used in a factory, I suggest to remove the class method again and create the season for the previous year directly in the factory.

spec/factories/teams.rb
```diff
@@ -11,6 +11,10 @@
       season { Season.current }
     end
 
+    trait :last_season do
+      season { Season.prev }
```
carpodaster: Does this work:
```ruby
season { create :season, name: Date.today.year - 1 }
```

app/models/season.rb
F3PiX: I tried that first, and didn't find a neat way to do that. 
In the teams factory, there is a trait :current_season, which calls the class method. I thought it will get very messy if I have to add a :last_season trait there, if it is one without a call to the class method. That's why I chose the option with a newly added class method.
The other solution I explored is to add a :last_season factory to the Season factory, but that would mean (AFAIK) that I would have to make associations from team factory to season factory. Apart from that I don't want to get into that, I think that would mess up all the RSpecs as well. 
Shall I go for the 'messy' :last_season trait in the team factory then?

spec/factories/teams.rb
F3PiX: I don't understand why, but it doesn't.  If I do it like that ^ or (that^).to_s, it raises a validation error when I reset the db. If I add a :prev_season to the Season factory and create or call that one from the teams factory, it is not recognized (either not registered or unknown constant.) 
???

carpodaster: I'll check it out tonight

carpodaster: Hey @F3PiX I was able to create a trait like this:

```ruby
trait :last_season do
  season { Season.find_or_create_by name: (Date.today.year - 1) }
end
```

F3PiX: 🙇

carpodaster:
Very cool! Thanks @F3PiX ! ✨

F3PiX:
Thanks for the lesson 💃

Issue #484, opened 2016-07-05
Remove the database column and all references to bank account info


anasofiapinho:
This year we don't have the bank account info on the students profile so it's better we remove the database column and all references to bank account info (words by @carpodaster), so that we don't get the "Bank info not filled out" note.

izuroxx:
@carpodaster looking into this one, if that's ok :)

PR #488, opened 2016-07-07
removed banking_info column from users table, views and controllers

izuroxx:
#484 removed banking_info @carpodaster @ramonh be gentle ;p

### DIFF
--- a/db/migrate/20160707112342_remove_banking_info_from_users.rb
+++ b/db/migrate/20160707112342_remove_banking_info_from_users.rb
@@ -1,0 +1,5 @@
+class RemoveBankingInfoFromUsers < ActiveRecord::Migration
+  def change
+  	remove_column :users, :banking_info, :text
+  end
+end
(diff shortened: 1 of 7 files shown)


ramonh:
Looks good to me! 👍 Nice one

db/migrate/20160707112342_remove_banking_info_from_users.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveBankingInfoFromUsers < ActiveRecord::Migration
+  def change
+  	remove_column :users, :banking_info, :text
```
carpodaster: Purely cosmetic stuff: the convention for indenting Ruby code is 2 spaces, no tabs. Your text editor can be set to enforce that, but the configuration options depend on the editor.

carpodaster:
Yes, looks good! Thanks @izuroxx !

db/migrate/20160707112342_remove_banking_info_from_users.rb
izuroxx: sorry! i'll keep that in mind! :)

Issue #455, opened 2016-04-12
Bugs in the Conferences Area


roxxx33:
@carpodaster 
![display_individual_conference](https://cloud.githubusercontent.com/assets/18331427/14476707/13213a7c-0127-11e6-8b50-5a4f4894cb56.png)
* [ ] In the display page for a single conference, the information is shown in an uneven format in cases such as when a field is left blank by a user that made the conference resulting in 2 fields appearing in the same line. 
* [x] A new conference can be added without a name, which can later not be viewed due to name field being blank in the display page for list of all conferences.

carpodaster:
Hi @roxxx33 – thank you for bringing this to our attention! Item no. 2 is done.

izuroxx:
@carpodaster while looking at this one, I've noticed that new conference saves without filling the tickets value - but it brakes conference/conferences show  (tickets_left method). would you like the tickets field to be required/ with default tickets value set to 0? :)

ramonh:
0 and required makes sense to me, thank you @izuroxx!

carpodaster:
Requiring it for new records makes perfect sense, yeah. As for the default value, I'm not so sure about that – mostly because it could obfuscate the presence validation in cases were entering a ticket number was simply forgotten.

The buggy show view for inconsistent records could probably be fixed by calling `to_i` on the the tickets value in the view template. Try entering `nil.to_i`  in `irb` 😉

ramonh:
Did not know that, cool!

agnik5:
I could continue working on this issue, if it's ok for you.

agnik5:
@carpodaster I guess I found another thing to repair: I was able to add a conference with "Starts on" and "Ends on" dates like: 20 Jul 16 - 07 Jul 16. The form shouldn't permit user to add such dates and should show error message. Am I right?

agnik5:
@carpodaster And after switching to season phase "Coding summer" it's possible to add a conference which took part in a period before. Though I'm not sure if I'm switching the right way.

F3PiX:
AFAIK, _adding_ a conference is not season-phase dependent in the current set-up. Not sure if it should be.

carpodaster:
@izuroxx @agnik5 Could you check the other missing validations that led to the @roxxx33 report of a buggy view? We can close this issue then.

As for the additional bugs regarding the conference area: could you please create a new issue?

agnik5:
@carpodaster ok!

[...]

PR #490, opened 2016-07-08
Fix tickets_left method and test

agnik5:
Ref. to #455 
tickets_left fixed by adding .to_i method to tickets
test in conference_spec.rb changed to reflect that


carpodaster:
Excellent! Thanks, @agnik5 !

F3PiX:
Gogo RubyCats!!!

Issue #491, opened 2016-07-08
'Real time' in season switch is broken 


F3PiX:
Issue: When switching the season switch to 'real time', we timetravel into 2015. That is definitively not realtime
I checked the season switch because of @agnik5 's comment on #455 (thanks @agnik5) 

This only happens in dev_env.

I think what happened is:
- In the season switch, we don't set the dates for real time, but we [destroy](https://github.com/rails-girls-summer-of-code/rgsoc-teams/blob/master/app/controllers/orga/seasons_controller.rb#L59) the current season
- In #486 , we introduce a second season in the seeds, ie last year's season (2015)
=> Now, when switching phases from a fake phase to real time, the season is destroyed , I think we fall back to the remaining season 2015.

Proposed fix:
- replace @season.destroy with a method that ensure we stay in real time.
- OR... ?

carpodaster:
@F3PiX I'm not sure if I can reproduce it correctly. I can confirm that immediately after the current season is destroyed and we are routed back to orga/seasons#index, the 2016 season is gone. However, the first call to `current_season` anywhere will automatically re-created it.

I've tried the following change locally but can you apply it on your end and see if it fixes things? Thanks! 😄 

```patch
index c644440..67823f1 100644
--- a/app/controllers/orga/seasons_controller.rb
+++ b/app/controllers/orga/seasons_controller.rb
@@ -58,6 +58,7 @@ class Orga::SeasonsController < Orga::BaseController
         @season.fake_coding_phase
       when 'RealTime'
         @season.destroy
+        Season.current # ensures a current Season is created right away
       end
     redirect_to orga_seasons_path, notice: "We time travelled into the #{params[:option]} phase"
   end
```

F3PiX:
@carpodaster Thanks! 
This solves one problem: now in the season's switch view, we DO see the current season. That is far better.

There still is a problem with the seeds. After switching to another phase and switching back to Real Time, the table in the Teams view is still empty. I *think* because there are no teams in the seeds with the proper (= new) season id.  The teams in the seeds are set only once, when they are created with db:reset.   

Not sure how to proceed. 
What I am thinking: in the seeds, there are only 2 season id's. Maybe in the 'Real Time' case, I could write something to update the season id's for the relevant seeds teams? 
Or?

carpodaster:
[...]

PR #492, opened 2016-07-10
Solve Season Switch error

F3PiX:
Solves #491
Before, the season switch to real time did season.destroy. This broke the references, especially from the seeds (only in dev env). Now the season switch updates the current season, and the references from the seeds work again.


carpodaster:
Looks good. Thanks Maud!

Issue #485, opened 2016-07-06
Only show selected teams in the drop down menu


anikalindtner:
Hey all, I've recently added some members to teams and it's really confusing and hard, because we have a seemingly endless list of teams to choose from - not just this year's teams, but also teams from last year and teams who applied but didn't get selected:

![screen shot 2016-07-06 at 08 47 22](https://cloud.githubusercontent.com/assets/1711357/16609690/51eeb8f8-4356-11e6-8fd0-0ed7f6a5ac60.png)

Who can change that? @F3PiX do you want to take that on?

carpodaster:
It's a valid point! Problem: we need to make the teams list for the select box Season phase-aware.

To get it fixed fast, I suggest we differentiate between "after rejection letters have been sent and coding phase has not yet ended" and "rest of the season" only. In the former case: feed the select box selected teams only.

F3PiX:
Cool, I will have a look.

F3PiX:
Where does this select-box live? Is it somewhere in orga(name)space?

F3PiX:
I will do the quick fix first. Then again, the workflow here looks a bit cumbersome to me for the beloved orga's. :-) Maybe look into this for a future Orga Dashboard?

F3PiX:
O sorry. Do we have a policy that only the opener of the issue can close? Or can I close this when it is merged.

carpodaster:
> Do we have a policy that only the opener of the issue can close? 

We do not have this set in stone, no. I usually ping the author of the issue via a comment when I think I fixed something and assign it back to them (only possible for project collaborators), but I do close issues myself if there hasn't been a response in some time. The original author will get a notification anyway and can re-open if necessary.

PR #493, opened 2016-07-12
Show relevant teams only in dropdown menu

F3PiX:
Solves #485 

From the day the acceptance letters are sent through the coding summer, only selected teams will turn up in the user profile editor (for orga's only). 
Plus some refactored Team.scopes

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -82,7 +82,7 @@ class UsersController < ApplicationController
 
     def teams
       all_teams = Team.all.order(:name)
-      selected_teams = Team.current_season.selected.order(:name)
+      selected_teams = Team.in_current_season.selected.order(:name)
       current_season.active? ? selected_teams : all_teams
     end
     helper_method :teams
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -82,7 +82,7 @@ class UsersController < ApplicationController
 
     def teams
       all_teams = Team.all.order(:name)
-      selected_teams = Team.active_teams.order(:name)
+      selected_teams = Team.current_season.selected.order(:name)
       current_season.active? ? selected_teams : all_teams
     end
     helper_method :teams
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -82,7 +82,7 @@ class UsersController < ApplicationController
 
     def teams
       all_teams = Team.all.order(:name)
-      selected_teams = Team.in_current_season.selected.order(:name)
+      selected_teams = Team.current_season.selected.order(:name)
       current_season.active? ? selected_teams : all_teams
     end
     helper_method :teams
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -82,7 +82,7 @@ class UsersController < ApplicationController
 
     def teams
       all_teams = Team.all.order(:name)
-      selected_teams = Team.current_season.selected.order(:name)
+      selected_teams = Team.current.selected.order(:name)
       current_season.active? ? selected_teams : all_teams
     end
     helper_method :teams
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -81,9 +81,9 @@ class UsersController < ApplicationController
     helper_method :conferences
 
     def teams
-      all_teams = Team.all
-      active_teams = Team.where(season: Season.current, kind: %w(sponsored voluntary)).order(:name)
-      current_season.active? ? active_teams : all_teams
+      all_teams = Team.all.order(:name)
+      selected_teams = Team.active_teams.order(:name)
+      current_season.active? ? selected_teams : all_teams
     end
     helper_method :teams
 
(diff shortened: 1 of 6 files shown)


app/controllers/users_controller.rb
```diff
@@ -81,6 +80,13 @@ def conferences
     end
     helper_method :conferences
 
+    def teams
+      all_teams = Team.all
+      active_teams = Team.where(season: Season.current, kind: %w(sponsored voluntary)).order(:name)
```
F3PiX: Wouldn't it be better to move the active_teams query to the teams model? Especially because there already lives a :by_season_phase scope. https://github.com/rails-girls-summer-of-code/rgsoc-teams/blob/master/app/models/team.rb#L49
I think I can't use that as is, because of the team visibility thingy in there.

app/models/team.rb
```diff
@@ -53,6 +53,10 @@ def by_season_phase
         Team.visible.where season: Season.current
       end
     end
+
+    def active_teams
+      Team.where(season: Season.current, kind: %w(sponsored voluntary)).order(:name)
```
F3PiX: I can't use the :by_season_scope, because of the visibility thingy in there. Hence the newly added scope here for the team query, and a season query in the season model.

carpodaster: I would rename the scope to `current_selected` or something like that. I wouldn't duplicate the term "team" in the scope since we're already telling `Team` to return teams.

Another idea: maybe modularize the scopes so that there is a `current` one (with all teams for the current season) and `selected` with (`kind: %w(sponsored voluntary)`). The required scope can then be composed with `current.selected` (and the other scopes can be refactored/DRYed to re-use the new scopes).

F3PiX: I was hoping you would say something like that :-) .  Maybe I better open a new PR for that, wdyt?

carpodaster: Nah, we can keep it in this PR

app/models/team.rb
```diff
@@ -45,18 +45,23 @@ def visible
     end
 
     # Returns a base scope reflecting the relevant teams depending on what
-    # phase of the running season we're currently in.
+    # phase of the running season we're currentin_ly in.
```
carpodaster: Typo 🎅

app/models/team.rb
```diff
@@ -45,18 +45,23 @@ def visible
...
+        Team.in_current_season.selected
       else
-        Team.visible.where season: Season.current
+        Team.in_current_season.visible
       end
     end
 
-    def active_teams
-      Team.where(season: Season.current, kind: %w(sponsored voluntary)).order(:name)
+    def in_current_season
```
carpodaster: Can you rename it to just `current`? We have similar scopes in other models and sticking with the same name keeps it consistent.

F3PiX: I will, sure, but I did it on purpose. I like a little more verbosity to makes things more clear and distinctive. And hopefully more beginner friendly. From my experience from last year, I found the fact that different things have similar names the most confusing. But it is probably a Rails convention, isn't it?

carpodaster: The concrete name »current« is not a Rails convention, but re-using method names on different classes (when the expected behaviour is the same or at least similar) is a _Ruby_ convention. Ruby focusses heavily on the idea of _duck typing_ – meaning it doesn't really care _what_ it is as long as it responds to a certain method.

But I understand what you mean, and the name `current` is totally up for debate.

What I suggest: let's stick with `current` in this PR. A follow-up PR can refactor all definitions of `current` (and its relevant factory traits) to `in_current_season`. It's a longer method name, but it totally reveals its intention. And that's way better! What do you think?

app/models/team.rb
```diff
@@ -45,16 +45,16 @@ def visible
...
-        Team.in_current_season.selected
+        Team.current_season.selected
       else
-        Team.in_current_season.visible
+        Team.current_season.visible
       end
     end
 
-    def in_current_season
+    def current_season
```
carpodaster: See my other comment. Let's stick to just `current` in this PR

app/models/team.rb
F3PiX: I like it very much! And thanks A LOT for the explanation. 

For the follow-up PR: I checked the factories, I didn't find a factory that uses the season scope of the Team class. The current_season traits all call the class method Season.current. 
You don't mean that the factory trait should have the Team scope instead, do you? 
And we are talking only about renaming current_season to in_current_season? Not current_other_things. 
 
If I understand it well: just renaming 'current_season' scope in Team class, to in_current_season scope: that is a tiny job, I can do it right away.
Otherwise, please merge and I will open a new PR and figure it out.

carpodaster: There are traits that are also referenced by `:current`. They should be called `:in_current_season` as well. But let's discuss the details in the new PR.

As for the merge: see my most recent comment below. If that's done, I'll merge in an instant :)

carpodaster: There are traits that are also referenced by `:current`. They should be called `:in_current_season` as well. But let's discuss the details in the new PR.

As for the merge: see my most recent comment below. If that's done, I'll merge in an instant :)

app/models/team.rb
F3PiX: Okay! I missed this while I was writing my answer. On it now

carpodaster:
Superb! Thanks @F3PiX ! ✨

Issue #482, opened 2016-07-05
Remove conferences 2015 from the User profile edit form


mkalininait:
User profile -> Edit

![screen shot 2016-07-05 at 12 48 27](https://cloud.githubusercontent.com/assets/1745735/16580826/f84fc50a-42ae-11e6-885c-bec4e66ccf21.png)

I see conferences from 2015 in the dropdown. Can we show only this year conferences or upcoming conferences?

F3PiX:
See discussion in #473

mkalininait:
@F3PiX I've seen it, thanks. I seems to me that this is a different bug. #473 is about conferences 2015 see till displaying for the supervisor who was a student, whereas this issue is about the possibility to select a 2015 conference on the user profile edit form.

F3PiX:
AHA you are riiiiiight. Sorry

F3PiX:
I made it up to you @mkalininait ;-)

carpodaster:
@mkalininait This was indeed fixed by @F3PiX – is it corrent that we don't have any conferences for 2016 in the Teams App yet?

mkalininait:
@carpodaster Yes, this is correct, we don't have conferences 2016 yet.
@F3PiX Thank you! 
I think we can close this issue. If something doesn't work with conferences 2016 (I will test once we have them), we will make another issue.

PR #494, opened 2016-07-12
Show only current Conferences in user profile

F3PiX:
Solves #482
Ignore conferences from previous seasons in selection dropdown.


carpodaster:
Looks good, let's give it a shot! :)

PR #498, opened 2016-07-13
Proxy images through camo

### LINKED ISSUE #479: Activity Stream: Proxy http:// images through camo
Currently images embedded / referenced to the activity stream ("minilog" feature) that do not use the https protocol scheme cause browsers to display a mixed content warning.

It's more of a nuisance than a real security threat of course, but it still doesn't feel right.

There is a service called [camo](https://github.com/atmos/camo) that proxies insecure assets through a TLS connection. It has to be self-hosted. It's a nice thing to play around with, I believe, but otoh, it adds complexity and takes some time to implement.

Is it worth it?

carpodaster:
This fixes #479

I tested it locally and it works. It was easier than I thought but I'm a bit sceptical if we want this at all: we need another Heroku app (the one that runs the camo app). Otoh, it just keeps everything as is when `CAMO_HOST` and `CAMO_ENV` are not configured.


carpodaster:
yolomerge!

alicetragedy:
and this merge on a Friday afternoon! That's so YOLO of you!

carpodaster:
😂

PR #501, opened 2016-07-14
Fix conference styling #455

agnik5:
Files modified:

**app/assets/stylesheets/global.sass**

Deleted `.conference` from
  ```
.profile, .mailing,
    dl
      // margin-bottom: 1em
      @include clearfix
    dt
      float: left
      width: 100px
      font-weight: normal
```

Added
``` 
.conference-row, .season-row
    width: 100%
    dt, dd
      display: inline-block
    dt
      font-weight: 200%
  
 
.season-row
    dt
      width: 10%

  .conference-row
    dt
      width: 20%
```

**app/views/conferences/show.html.slim**

Added divs of class `.conference-row` like
```    
.conference-row
      dt Location
      dd = conference.location
```

**app/views/orga/seasons/show.html.slim**

Added divs of class `.season-row` like
```   
 .season-row
      dt Year
      dd = @season.name
```
due to distinguish seasons styling from conferences styling (both had `.conference` class before) and make the code compatible with above mentioned changes

Result:
screenshot from 2016-07-14 14-59-20
screenshot from 2016-07-14 14-59-25

### DIFF
--- a/app/assets/stylesheets/global.sass
+++ b/app/assets/stylesheets/global.sass
@@ -184,8 +184,11 @@ html, body
       width: 100px
       font-weight: normal
   .conference
+    dl
+      width: 100%
     dt
-      width: 120px
+      width: 20%
+      text-align: left
 
   .status_update
     dl
--- a/app/views/conferences/show.html.slim
+++ b/app/views/conferences/show.html.slim
@@ -13,7 +13,7 @@ h1 class="conference" = conference.name
   p
 
 .conference
-  dl
+  dl.dl-horizontal
     dt Location
     dd = conference.location
     dt Date
--- a/app/views/orga/seasons/show.html.slim
+++ b/app/views/orga/seasons/show.html.slim
@@ -1,7 +1,7 @@
 h1 Season #{@season.name}
 
 .conference
-  dl
+  dl.dl-horizontal
     dt Year
     dd = @season.name
     dt Start Date

### CODE REVIEW

**Comment by carpodaster** [comment]
Hello @agnik5 and thank you for this PR :) 

My CSS skills are virtually nonexistent but I noticed a repeated `conference-row` / `season-row` class. Wouldn't it be better to define the styles for the outer `dl` element?

Could you add screenshots of how it looked before? (Sure, I could fire up a server, create a conf and see for myself. It's just easier to review a PR with a screenshot attached).

I wonder if we could collapse the styling for season and conference; they only differ by a 10% vs. 20% width. Does the season styling look weird/funny/bad with 20% width? If it can be generalized, a more general class name would be good – alas, I can't think of one right now 😿 

Maybe @alicetragedy can lend some CSS/styling advice?

Oh, and you don't have to state the changes manually. It's quite some work to style all the code blocks and GH already [does that for us](https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/501/files).

**Comment by alicetragedy** [comment]
Using very specific rules can get confusing fairly quickly, so I'm generally of the opinion to try and make a single class work for several sections (e.g. the `season` and `conference` views in this case). 
If we could stick to a single class (`.conference` or something less general) if would be 👍  
I'm going on holiday tomorrow ☀️ but I will try and see if I can take a better look at this later.

**Comment by ramonh** [comment]
Hey @agnik5, thanks for the PR!

I agree with @carpodaster and @alicetragedy. I think it makes sense to consolidate both views' styles for consistency. I'm struggling to come up with an overall class name, though. Maybe something like `.description`, since we're using `dt` and `dl` tags?

**Comment by agnik5** [comment]
Hi everybody, thanks for your comments.

My intention was to distinguish styling referring to different `show.html.slim` files, also keeping in mind probable "Refactoring of styling - global.sass" proposed in #504. I thought such naming would be more intuitive.

But of course there's no problem of changing the code as you wish with the use of one class instead of two :) 

@carpodaster Don't worry, your CSS skills are much better than mine :D I'm not sure if it's possible to define the styles for the outer `dl` element as I guess the core issue was to make rows containing `dd` and `dt` elements  displayed as block elements. Or maybe I misunderstood your above comment.
Before it looked like in the roxxx33's screenshot in the #455 thread https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/455. Seasons don't look very bad with 20% width, just names there are shorter than in conferences, so 10% is just a bit better /smaller space between columns/.

@alicetragedy Have a great holiday!
@alicetragedy @ramonh maybe sth like "block-row"?

It's also possible to fix it with `dt width: auto` + `dt margin: XX px`, but it looks better with width set in `%` as names of `dt` have different length.

Just to ask in case of myself learning: does such repeated code for different classes harm app's performance or is it just a case of good practice? I'd like to know your point of view to learn more and apply this knowledge in future.

**Comment by agnik5** [comment]
Screenshot of seasons with 20% width
screenshot from 2016-07-19 14-26-31

**Comment by ramonh** [comment]
Hey @agnik5!

I think the reason is for maintaining the styles. Since we have very similar styles for conferences and seasons, it makes the most sense to try and combine them into one general style, so that if we make any changes we can do them in one fell swoop. 

Furthermore, Having the repeated row classes makes things a little brittle — it could happen that we'd forget to use it.

Removing the row div altogether and trying to make the styles for conferences and seasons the same makes the html a little easier to read and the styles less error prone.

I hope that makes sense and that I got it right!

**Comment by ramonh** [comment]
Ahem, please excuse the broken sentence. I'm awful at typing on mobile 😅

**Comment by agnik5** [comment]
Thanks @ramonh. If I understood you correctly, you mean not only having one class for styling both conferences and seasons, but also undoing making any new divs in slim files, whatever class they are?

**Comment by ramonh** [comment]
That's right! Having those repeated row divs isn't fully necessary when we can leave the styling to the concrete tags inside, or the containers.

Generally it makes sense not to create a lot of divs that we can do without, as this also makes the html dom structure harder to read or debug.

Hope this helps 😄

**Comment by agnik5** [comment]
OK so what do you think about changing `global.sass`
```sass
  .profile, .mailing, .conference
    dl
      // margin-bottom: 1em
      @include clearfix
    dt
      float: left
      width: 100px
      font-weight: normal
  .conference
    dt
      width: 120px
```
into
```sass
  .profile, .mailing, .conference
    dl
      // margin-bottom: 1em
      @include clearfix
    dt
      float: left
      width: 100px
      font-weight: normal
  .conference
    dl
      width: 100%
    dt
      width: 15%
      text-align: left
```
with also changing `dl` to `dl.dl-horizontal` in both `show.html.slim` files for seasons and conference views?

screenshot from 2016-07-20 13-29-37
screenshot from 2016-07-20 13-30-01

ramonh:
That looks great, thanks @agnik5! The `.dl-horizontal` would make them responsive, is that right?

agnik5:
Yes, to some extent. But not for very small screens I guess. I changed the width to 20% and in my opinion it's the biggest width that will look ok on larger screens. PR or improvements?

ramonh:
Great stuff, thank you @agnik5! 

Hitting the merge-omatic 🎉

PR #507, opened 2016-07-19
Add option to show all public acitivities

ramonh:
screen shot 2016-07-19 at 21 31 32

This adds a default option to show all public acitivity on the landing page. This way, one can also revert their choice if one of the two non-"All" options are selected.


ramonh:
Thank you so much @katrin-k for your help with this!

carpodaster:
🎊

Issue #379, opened 2016-02-25
Feature request: SEARCH! 


alicetragedy:
The community area (/users) tends to become quite full every year. Even though we have pagination in place, finding someone in there is not 100% user friendly. I'd love for us to have a search in there in order to find someone quickly (mostly by name/username).

carpodaster:
Morning @alicetragedy! Agreed, a search would be very helpful. I'm not too sure about searching for a (user)name, though. Ideally, you'd search for people in your vicinity, but that requires geocoding all users based on their location and may yield funny results if someone could not be geocoded (e.g. because they decided not to give any location info).

F3PiX:
I thought that I was able to do that, so I tried and I COULD :tada: 
@carpodaster: I know it is not the BEST reason to add a feature :-) . WDYT?
See PR #382

alicetragedy:
thanks @F3PiX! :sparkles: 
@carpodaster you're right, for a lot of users, searching by location would make a lot of sense. In my case though, I often look up if people already exist in our community list and do so by name. Anyhow, I'm leaving this open, as I think we might want to think about ways to improve the search that Maud has already added (ie. adding location).

F3PiX:
In that case: maybe searching team names is also a nice future improvement.

F3PiX:
Good one, I will add that. And very glad that you will take care of the styling 🍹

F3PiX:
I need some help, please. 
First option is a reset link, that is disabled until a search is committed. This works, but needs tender love and caring in the css. 
![active link](https://cloud.githubusercontent.com/assets/6314687/16994596/51628f42-4ea8-11e6-8f02-83696e3727f7.png)
![inactive link](https://cloud.githubusercontent.com/assets/6314687/16994607/5f0fe96e-4ea8-11e6-8c3c-8d266ce72ff5.png)

Second try was with .js script on the link, to hide the link, and show after GO is clicked. The script works, but the link immediately disappears when the page is refreshing. Which makes sense, because after refreshing the hidden setting is activated.
Third try was similar script on the GO button. Same effect.
This is my first js script ever, so please don't think I know what I am talking about ;-) 
 
What to do? Quick fix is PR with option 1 (screenshots). 
Or I put the JS script in the PR, and someone takes over?

katrin-k:
I'd say the first option is fine so far. 
I can take over CSS for the link. 

[...]

PR #508, opened 2016-07-21
Clear Search Results

F3PiX:
#379 
Add link to clear search results, disabled until there is a search submitted.
- [x] add link to clear current Search results
- [x] add styling to link
- [ ] ~~hide link until search results on display~~

### DIFF
--- a/app/views/users/_search.html.slim
+++ b/app/views/users/_search.html.slim
@@ -1,0 +1,4 @@
+= form_tag users_path, method: :get, id: "search-form" do |f|
+  = text_field_tag :search, params[:search], placeholder: "Search by name"
+  = submit_tag "Go", class: "btn btn-default"
+  = link_to_if(params[:search], "Clear search results", users_path)
--- a/app/views/users/_search.html.slim
+++ b/app/views/users/_search.html.slim
@@ -1,4 +1,4 @@
 = form_tag users_path, method: :get, id: "search-form" do |f|
   = text_field_tag :search, params[:search], placeholder: "Search by name"
   = submit_tag "Go", class: "btn btn-default"
-  = link_to_if(params[:search], "Clear search results", users_path)
+  == link_to_if(params[:search].present?, "Clear search results", users_path)
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -18,7 +18,7 @@ class UsersController < ApplicationController
     @users = @users.with_assigned_roles if Time.now.utc > (current_season.starts_at || Date.new)
     @users = @users.with_role(params[:role]) if params[:role].present? && params[:role] != 'all'
     @users = @users.with_interest(params[:interest]) if params[:interest].present? && params[:interest] != 'all'
-    @users = @users.search(params[:search]) if params[:search]
+    @users = @users.search(params[:search]) if params[:search].present?
   end
 
   def show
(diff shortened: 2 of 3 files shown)


F3PiX:
@katrin-k The stage is yours ;-)

katrin-k:
@F3PiX I just wanted to take care of the link, but it doesn't render as a link when I run it locally. 
The output misses the HTML:
```
<form …>
  <input name="search" …>
  <input name="commit" …>
  Clear search results
</form>
```

Did some code snippet play lemming?

Re: the 3rd list item in this PR (hide the link): I'd rather put this into a separate + new issue to not delay the merge of this PR for too long. (Plus, I'm not 100% convinced this is really necessary given the Ajax effort…)

Re: the X icon: I'd rather not use it this way. From a UX pattern/user perspective, the X inside an input field first and foremost removes the entry done by the user, not the result. Usually, the result is then reloaded (more or less) seamlessly. Which again brings us to Ajax… And since we reload the page I'd go for a standard button. Wdyt?

F3PiX:
Haha, that got me confused. Turns out, I wrote this in html.erb. I am going to slim it up right now.

F3PiX:
Totally agree with everything you say @katrin-k  
I hope the .erb file was the problem, couldn't reproduce it myself.

katrin-k:
@F3PiX strangely, .erb was not the problem. Slim also doesn't render the HTML here… :crying_cat_face: 

Could anyone else take a look, plz? 
Tl;dr On my machine, `link_to_if` does not render the HTML. Works fine for Maud, though.

app/views/users/_search.html.slim
```diff
@@ -0,0 +1,4 @@
+= form_tag users_path, method: :get, id: "search-form" do |f|
+  = text_field_tag :search, params[:search], placeholder: "Search by name"
+  = submit_tag "Go", class: "btn btn-default"
+  = link_to_if(params[:search], "Clear search results", users_path)
```
carpodaster: `params[:search].present?` would be better in case the param got carried over as `…?search=`, i.e. w/o a value that could be interpreted as an empty string. An empty string is still a truthy value in Ruby.

F3PiX:
This is taking html escaping to a whole new level.

carpodaster:
@F3PiX is this still WIP? Looks good to merge.

That does the `==` do in Slim?

F3PiX:
Good morning karpermeester. It is still WIP, working with katrin_k on the link.
== is output without html escaping. Jury is still out on this solution.

katrin-k:
Upsi, I just realized I made a mistake in describing the problem. When there's no search param, the link text is displayed without any HTML tags (but it's displayed, which it shouldn't). The link properly renders when a search param is there. `params[:search].present?` doesn't make a difference. 

What does the trick is defining an else. 
```
= link_to_if(params[:search].present?, "Clear search results", users_path, class: "btn btn-default")  do
 a
end
```

However, the `end` is then rendered as HTML tags: `<end></end>`. Removing the `end` helps in the way that there's only an empty a link tag when there's no search param, but I'll keep it since I have a hunch this is better Ruby syntax and thus can't interfere with anything else. 

@F3PiX I'll also add the `params[:search].present?` to my commit.

katrin-k:
Close due to duplicate in #511.

PR #509, opened 2016-07-21
Search for team names too

F3PiX:
#379 
The search form now returns team names too.
- [x] Query for searching team names
- [x] Search result with both users with matching names and users with matching team names
- [x] Make class method work from outside controller
- [x] Write tests

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -108,6 +108,10 @@ class User < ActiveRecord::Base
       joins(:roles).where('roles.id IS NOT NULL')
     end
 
+    def with_teams
+      joins(:teams).references(:teams)
+    end
+
     def with_team_kind(kind)
       joins(:teams).where('teams.kind' => kind)
     end
@@ -154,7 +158,9 @@ class User < ActiveRecord::Base
   end
 
   def self.search(search)
-    where("users.name ILIKE ? OR teams.name ILIKE ?", "%#{search}%", "%#{search}%")
+    q_user_names = User.where("users.name ILIKE ?", "%#{search}%")
+    q_team_names = User.with_teams.where("teams.name ILIKE ?", "%#{search}%")
+    q_user_names + q_team_names
   end
 
   private
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -154,8 +154,7 @@ class User < ActiveRecord::Base
   end
 
   def self.search(search)
-    where("users.name ILIKE ?", "%#{search}%")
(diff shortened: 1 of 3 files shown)


app/models/user.rb
```diff
@@ -155,6 +155,7 @@ def current_student?
 
   def self.search(search)
     where("users.name ILIKE ?", "%#{search}%")
+    where("teams.name ILIKE ?", "%#{search}%")
```
carpodaster: Hey @F3PiX, I think chaining `where` statements constructs `AND` queries in SQL. Furthermore, I believe the teams association has to be added with `includes(:teams).references(:teams)`, too.

F3PiX: Thank you! That helps.

F3PiX:
The good news is: it works. 
The query returns both user names and team names. 
find users and teams
I don't understand WHY it works with `teams.name`. (Could it be because of the `has_many`-relations? I trial-and-errored with `roles.name` and that works as well. )
And if it DOES work, is it an acceptable way as well?

carpodaster:
Morning @F3PiX 

>  I don't understand WHY it works with teams.name.

It works in the controller, because line 16 states `with_all_associations_joined`. Try firing `User.search "cheesy"` in your rails console to reproduce the error I was expecting in an earlier comment.

Since `User.search` is part of the public API of that model and therefore can be called out of the controller context, it should bring its own include/join statements. We should probably have a unit test for the class method as well.

F3PiX:
Aha. 
I need some time figuring out how to put the associations into the combined query. I would like to, it's something I want to find out and I have an action plan. 🚀

app/views/users/index.html.slim
```diff
@@ -41,5 +41,5 @@ table.users.table.table-striped.table-bordered.table-condensed.table-hover
         / td = links_to_conferences(user.conferences.current).join(', ').html_safe
 
 p.pagination-info
-  = page_entries_info @users
-  = paginate @users, theme: 'twitter-bootstrap-3', pagination_class: 'pagination-sm'
+  = page_entries_info @users unless params[:search].present?
```
F3PiX: Tips for better solution than this?

F3PiX:
Hi @ramonh. Could you take a look please?

app/views/users/index.html.slim
F3PiX: I found a better solution. Following in next pr.

ramonh:
Looks great, thank you @F3PiX! ✌🏽

Issue #413, opened 2016-03-13
Information pulled from github should not be editable/deleted


Nattivolk:
- [ ] github handle and email have a star i guess for indicating a mandatory field, but i can save without an email address. This happend, because I deleted the email. Might happen, if a candidate wants a different email, deletes it and then forgets to add the new one. 
- [x] I'm not sure, if this is wanted, but I can also change the github handle. As I signed in with a specific account I don't know if we want to allow to change this field?
  Same here, I just edited the github account to a different one and could save the application

Possible fix: Github handle used for sign up should not be editable and email should be verified.

carpodaster:
Hey @Nattivolk - thanks!

The reason behind 1) is that a `User` record is indeed created w/o an email address at first: the rest gets pull from the GH API asynchronously. We should improve that, though. A validation context should fix that: the easy win would be to allow a `User` to _created_ w/o an email address, just not _updated_

2) Making the GH-handle of a created `User` immutable is sounds like a really good idea!

izuroxx:
hi @carpodaster! is this what you have in mind? (github_handle taken out of the params, field disabled; validation on email) 
![gh_greyedout_email_validation](https://cloud.githubusercontent.com/assets/14924179/16730923/c6e5ebae-4775-11e6-81f2-9aba53197636.jpg)

carpodaster:
Hey @izuroxx. Yes, looks good! The email validation should only kick it when _updating_ the record though (for reasons stated above). But we can discuss / check on that in a PR.

izuroxx:
@carpodaster there's one more thing :) - while finding all occurrences of Github_handle fields, I also found them in /teams/:id/edit, and also editable - you can change team members through changing the g_h field - would you like those fields to be disabled, too?

ramonh:
@izuroxx I think these should be left as editable so that students for example can add their team members. I hope I understood correctly

F3PiX:
I think Ramon is right, as far as teams in current season are concerned. Then again, I can edit _everything_ in my team from _last_ year. Maybe we should 'freeze' team info from previous seasons. Or is that a problem, @anikalindtner @carpodaster ?

carpodaster:
@izuroxx fixed the github_handle mutability! 🎊

F3PiX:
Can we extract a new issue, with only the first checkbox? @carpodaster

carpodaster:
@f3pix let‘s do it!

[...]

PR #514, opened 2016-07-27
WIP: Immutable GitHub handle

izuroxx:
#413 
I've added custom validation in users model - immutable_github_handle & added test to it. it should not let the gh handle to be changed on update. before you could freely change it. (commit 61e43fa)
immutable_gh_before_disabling_field

in the end i've disabled the filed (commit 78326fc)
immutable gh

Now, few test are failing: 
spec/controllers/users_controller_spec.rb
spec/models/creates_application_from_draft_spec.rb
spec/controllers/application_drafts_controller_spec.rb
spec/models/application_draft_spec.rb

@ramonh i've tried to fix this, tired to exclude github_handle f.ex. in application_draft factory, but I'm a bit lost in tracing the problem. Would you give my some kind of a hint, direction, feedback about the method, tests?

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -49,7 +49,7 @@ class UsersController < ApplicationController
 
   def update
     respond_to do |format|
-      if @user.update_attributes(restricted_user_params)
+      if @user.update_attributes(user_params.except(User.non_updateable_params))
         format.html { redirect_to params[:redirect_to] || @user, notice: 'User was successfully updated.' }
         format.json { head :no_content }
       else
@@ -66,7 +66,6 @@ class UsersController < ApplicationController
       format.json { head :no_content }
     end
   end
-
 
   private
 
@@ -107,9 +106,4 @@ class UsersController < ApplicationController
     def normalize_params
       params[:role] = 'all' if params[:role].blank?
     end
-
-    def restricted_user_params
-      user_params.except(:github_handle)
-    end
-
 end
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -49,7 +49,7 @@ class UsersController < ApplicationController
 
   def update
     respond_to do |format|
-      if @user.update_attributes(user_params.except(User.non_updateable_params))
+      if @user.update_attributes(user_params)
         format.html { redirect_to params[:redirect_to] || @user, notice: 'User was successfully updated.' }
         format.json { head :no_content }
       else
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -49,7 +49,7 @@ class UsersController < ApplicationController
 
   def update
     respond_to do |format|
-      if @user.update_attributes(user_params)
+      if @user.update_attributes(restricted_user_params)
         format.html { redirect_to params[:redirect_to] || @user, notice: 'User was successfully updated.' }
         format.json { head :no_content }
       else
@@ -107,4 +107,9 @@ class UsersController < ApplicationController
     def normalize_params
       params[:role] = 'all' if params[:role].blank?
     end
+
+    def restricted_user_params
+      user_params.except(:github_handle)
+    end
+
 end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -128,7 +128,7 @@ class User < ActiveRecord::Base
     end
 
     def non_updateable_params
-      [:github_handle]
+      :github_handle
     end
   end # class << self
 
@@ -166,11 +166,6 @@ class User < ActiveRecord::Base
     q_user_names = User.where("users.name ILIKE ?", "%#{search}%")
     q_team_names = User.with_teams.where("teams.name ILIKE ?", "%#{search}%")
     (q_user_names + q_team_names).uniq
-  end
-
-  def delete_keys_for_update(keys)
-    hash = self.dup
-    keys.each {|key|hash.delete(key)} 
   end
 
   private
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -127,7 +127,7 @@ class User < ActiveRecord::Base
       where(":interest = ANY(interested_in)", interest: interest)
     end
 
-    def non_updateable_params
+    def immutable_params
       :github_handle
     end
   end # class << self
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -126,7 +126,11 @@ class User < ActiveRecord::Base
     def with_interest(interest)
       where(":interest = ANY(interested_in)", interest: interest)
     end
-  end
+
+    def non_updateable_params
+      :github_handle
+    end
+  end # class << self
 
   def rating(type = :mean, options = {})
     Rating::Calc.new(self, type, options).calc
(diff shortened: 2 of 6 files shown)


spec/models/user_spec.rb
```diff
@@ -17,6 +17,25 @@
   it { expect(subject).to allow_value('https://example.com').for(:homepage) }
   it { expect(subject).to_not allow_value('example.com').for(:homepage) }
 
+  describe 'immutable github handle validation' do
+    context 'when creating a new user' do
+      let(:new_user) { build(:user, github_handle: 'github_handle') }
+
+      it 'allows creating a user with given github handle' do
+        expect(new_user.valid?).to eq(true)
```
ramonh: Minor thing here you might like. You can instead write:
`expect(new_user).to be_valid`

izuroxx:
@ramonh i hope it all makes some sense :) test pass this time. there's also a email validation to this issue, which i've already started, but wanted to focus on this github_handle first

spec/factories/application_drafts.rb
```diff
@@ -14,7 +14,8 @@
 
       after(:create) do |draft|
         draft.students.each do |student|
-          student.update(attributes_for :student, :applicant)
+          student_update_attributes = (attributes_for :student, :applicant).except(:github_handle)
+          student.update(student_update_attributes)
```
ramonh: Great catch!

I wonder if there would be a way to clean this up so as to not have the application draft know about `github_handle`s. What do you think?

izuroxx: yes, that would be great :) challenge accepted :+1:

ramonh: Awesome! :v:

carpodaster:
The changes made to `User.non_updateable_params` in 06c22b8 were alright. I don't think it works without the class method returning an `Array` and `Hash#except` being called with the splat'ed list of immutable attributes.

I think the reason why the tests failed is because `spec/controllers/users_controller_spec.rb:6` incorrectly removes github_**id** where it should be github_**handle**. You apply the changes in 06c22b8 again (only without `delete_keys_for_update` since it is not needed) and modify the spec as stated above? Do the tests still break?

app/models/user.rb
```diff
@@ -125,7 +126,11 @@ def with_all_associations_joined
     def with_interest(interest)
       where(":interest = ANY(interested_in)", interest: interest)
     end
-  end
+
+    def non_updateable_params
```
carpodaster: Naming suggestion: `immutable_attributes`

app/controllers/users_controller.rb
```diff
@@ -49,7 +49,7 @@ def create
 
   def update
     respond_to do |format|
-      if @user.update_attributes(user_params)
+      if @user.update_attributes(user_params.except(User.non_updateable_params))
```
carpodaster: I'm thinking: isn't the validation introduced below enough? If a user changes their GH handle, the validation will trigger and the record won't save. It will also display a meaningful error message.

izuroxx: understood :) i've deleted the except method. the github_handle field in a form is disabled, i've added disabled: true in the edit template (commit 78326fc). But when not disabled it displays github_handle can't be changed when trying to change it. shall i leave it disabled?

izuroxx:
c8965d3 - changed `update attributes` in update action in user controller `immutable_github_handle` to` immutable_params`, deleted changed update attributes in update action in user controller
0a6f1fd - added `User.immutable_params.to_s` in user_controller_spec

app/models/user.rb
```diff
@@ -127,7 +127,7 @@ def with_interest(interest)
       where(":interest = ANY(interested_in)", interest: interest)
     end
 
-    def non_updateable_params
+    def immutable_params
       :github_handle
```
carpodaster: This should be an Array

app/models/user.rb
izuroxx: the `immutable_params` where created at first for future-proofing if there might be some changes with github_handle or more params added and also in order to call `User.immutable_params` in application_draft factory (+ user controller in update action - deleted in last commits). 
Application draft factory needed `User.immutable_params` to return symbol, but i wanted to test how to pass more symbols, and that's why i made an array and a splat.
As for failing tests - everything passes since	200f387 commit, now it does too.
As for `spec/controllers/users_controller_spec.rb:6` - "github_id" and avatar_info were in the test before I started to fix it. I guess it passed because i've had filtered Update_attributes in update action. But since i've deleted it -  I added `User.immutable_params.to_s` in the controller test to exclude github_handle. Hope it all makes some sense, if not please let me know how bad i've messed up this time ;p I will try to understad and fix it again :)

spec/controllers/users_controller_spec.rb
```diff
@@ -3,7 +3,7 @@
 describe UsersController do
   render_views
 
-  let(:valid_attributes) { build(:user).attributes.except('github_id', 'avatar_url') }
+  let(:valid_attributes) { build(:user).attributes.except('github_id', 'avatar_url', User.immutable_params.to_s) }
```
carpodaster: If the class new method returns an array, it needs to be called with a `*`: 

```ruby
let(:valid_attributes) { build(:user).attributes.except('github_id', 'avatar_url', *User.immutable_params.map(&:to_s)) }

# Or:

let(:immutable_attributes) { User.immutable_params.map(&:to_s) + %w(github_id avatar_url) }
let(:valid_attributes) { build(:user).attributes.except(*immutable_attributes) }
```

spec/factories/application_drafts.rb
```diff
@@ -14,7 +14,7 @@
 
       after(:create) do |draft|
         draft.students.each do |student|
-          student_update_attributes = (attributes_for :student, :applicant).except(User.non_updateable_params)
+          student_update_attributes = (attributes_for :student, :applicant).except(User.immutable_params)
```
carpodaster: Needs a splat (`*`) - see above

app/controllers/users_controller.rb
izuroxx: @carpodaster 
i've added disabled: true to the form in edit user template, so user can't access it  - shall i leave it that way or let it display error when changed

app/models/user.rb
```diff
@@ -125,7 +126,11 @@ def with_all_associations_joined
     def with_interest(interest)
       where(":interest = ANY(interested_in)", interest: interest)
     end
-  end
+
+    def immutable_params
```
carpodaster: As suggested before: could you name this `immutable_attributes` please? Rationale: in the context of this AR model, attributes is more appropriate. "params" hints at a controller context.

carpodaster: The `def  … ` isn't properly indented :)

app/controllers/users_controller.rb
carpodaster: Leaving it disabled is perfect. It doesn't obsolete the validation, though: remember all form data can be forged so it's not a sufficient to rely on the `disabled` HTML attribute: it's very easy to remove the disabled flag with your DOM inspector / editor of choice. The validation will catch those forged request.

TL;DR: it's all good with the form 👍

app/models/user.rb
izuroxx: :sob: true, it was suppose to be attributes, sorry for that

indentation - `def immutable_attributes` is in class << self block

carpodaster: About the indentation: ooops! My bad. Sorry 😞

carpodaster:
Looks superb now! Let's merge this.

ramonh:
YAY! Nice one @izuroxx!

izuroxx:
Thank You! :bow:

PR #515, opened 2016-07-31
FIX: count feed_entries as log updates

### LINKED ISSUE #512: Daily log reminder even when daily log is present
My team fills out their daily logs on their blog and it gets fed into the teams app. I can read it on the same day it's "due" on the teams app, but they get an email reminder every day nonetheless.

The easy solution would be that they just fill it out on the teams app. I think it would be a huge shame that they abandon their blog (which they can continue to use even after the summer to track their progress) and even more of a shame if we are the ones to encourage them to do so.

The other solution would be that they just copy paste it off their blog, which is a hassle. 

I think the best solution would to find out why the teams app doesn't recognize this feed and sends them the email even though the log is there, on time, on the app, available for everyone to read 😄

michaelem:
This changes the without_recent_log_update scope in order to count feed_entries as well since some teams actually use these instead of status_updates. This fixes #512, using the suggestion from @F3PiX .

### DIFF
--- a/app/models/team.rb
+++ b/app/models/team.rb
@@ -32,7 +32,7 @@ class Team < ActiveRecord::Base
   before_save :set_last_checked, if: :checked
 
   scope :without_recent_log_update, -> {
-    where.not(id: Activity.where(kind: 'status_update').where("created_at > ?", 26.hours.ago).pluck(:team_id))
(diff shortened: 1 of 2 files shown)


ramonh:
Nice one, thank you @michaelem!

PR #516, opened 2016-07-31
Fix voor Kaminari paging

F3PiX:
Kaminari provides an array wrapper. This solves 2 things: the ugly checks in views if search params are present, PLUS the empty pages at the end of the user list.

Before:
before
After
after

Very happy that this longlasting issue is solved... (https://github.com/TeamCheesy/rgsoc-teams/issues/16)


carpodaster:
I've checked it out locally and clicked around a bit: looks good! Thanks a lot, Maud 👍

PR #518, opened 2016-08-09
Style filter on users_info/index

katrin-k:
Just a minor thing, but a bit annoying whenever I looked at this page. 

Screenshot: (It now looks the same as on users/index)
pr selection box on users-info

I also wanted to add the search input, but failed. 
- Adding `= render partial: 'users/search'` won't help since it requests the wrong page.
- Pasting the code snippet and adjusting the paths to `users_info_path` did help, but the search results would still show all entries. 
Any ideas?

### DIFF
--- a/app/views/users_info/index.html.slim
+++ b/app/views/users_info/index.html.slim
@@ -1,17 +1,18 @@
 h1.header User info
 
-form.filter action=request.url
-  h5 Role
-  - (['all'] + Role::ROLES).each do |role|
-    label.radio
-      = radio_button_tag :role, role == 'all' ? '' : role, params[:role] == role
-      = role.capitalize
+.selection-box
+  form.filter.form-inline action=request.url
+    h5 Role
+    - (['all'] + Role::ROLES).each do |role|
+      label.radio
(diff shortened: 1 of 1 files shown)


F3PiX:
'users/search' gets the params through the users_controller. If you don't go through that controller, the search field will send a 'no filter'-param to the search method.
That may explain why you can't cut and paste, but offers no solution. I think  `render partial:  'views/users/search'  ` would request the right _page_ but I expect that it is not solving your question.
I just learned :-) how to show a form on a different page with js and ajax. Maybe we could go that way?

F3PiX:
But first things first: looks a ton better.

katrin-k:
Aah, so the users_info_controller needs to be prepared, too? I had a hunch there was sth. like this necessary. 

I'd be interested in switching to Ajax/JS for loading the results :). However, I'd suggest to open a new issue for it.

F3PiX:
This is ready for review/merge, right, @katrin-k ? Or do you want a search form added first?

katrin-k:
@F3PiX naaah. I think it can live without one for now.

So yeah, good to merge.

alicetragedy:
cool, then let's merge this \o/

Issue #352, opened 2015-12-07
Reposition icon beside team name (due to mobile issue)


alicetragedy:
On top of the nav issue already mentioned in https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/335, I've noticed a couple of glitches with ie. Headers being cut off when the browser is resized. In the example below, the start is cut off when width of browser is between 992 and 1200px.

<img width="594" alt="screen shot 2015-12-07 at 15 15 13" src="https://cloud.githubusercontent.com/assets/3143348/11629033/de0566b4-9cf5-11e5-8607-067de9e25983.png">

This is just a minor issue with media-queries that needs fixing.

katrin-k:
@alicetragedy can you plz check if this issue can be closed?

alicetragedy:
@katrin-k the glitch is definitely not there anymore (thank you for fixing!) but there are still some things that aren't optimal: for example, the menu with buttons under the team name could be changed, so that the buttons aren't strangely stacked together. 
Maybe I can go through all that needs improving and add that to my original comment as a checklist, so that we can monitor what has been done and what still needs doing. wdyt?
Alternatively we can close this issue and open a new one specifically for the buttons (or any other improvement). Let me know what you prefer.

katrin-k:
@alicetragedy I was pondering over this, too… Looking at several pages, I have a strong urge to doublecheck and, if appropriate, overwork quite a few things, e.g. button/color patterns (including those button groups), the way forms are styled, all those tables that overflow in mobile views (Anika already made a suggestion to style them in a more appealing way) … I also have a few ideas inspired by the new design for the rorganize.it platform. 

Yeah, let's close this and open up new issue(s) with specific stuff. If you do find the time to go through things, that'd be super helpful!  

I'll also rename this issue to cover just the team name icon, so it can be recognized better later on.

PR #522, opened 2016-08-09
352-Optimize mobile view for team/show

katrin-k:
(New, 2nd PR)

On the team/show page, the star left of the team name was at certain points overflowing. The inital problem showed up on a larger screen, but it basically always shows up around the media breakpoints. The star-icon was positioned outside of the main container. 
My proposal: The star is now part of the container and thus follows the overall breakpoints. 

Screenshot:
352-reposition-star

Resolves issue #352.

### DIFF
--- a/app/assets/stylesheets/global.sass
+++ b/app/assets/stylesheets/global.sass
@@ -309,10 +309,6 @@ html, body
     white-space: nowrap
 
 .teams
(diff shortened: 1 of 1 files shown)


ramonh:
Looks great, thank you @katrin-k! Merging now

PR #528, opened 2016-08-16
WIP: Add comments to student's logs

F3PiX:
To do

- [x] Add basic comments setup to status_updates only (Maud)
- [ ] ~~Extend to feed_entries (instead of @status_update) (Maud)~~
- [x] Solve routing error in redirection to PATH_PARENTS after comment is saved (Maud through: :michael)
- [ ] Add javascript to show 'new comment'-form on activities index page (Katrin & Maud)
- [ ] Styling on index page (Katrin)
- [ ] 'Add a comment' button or link


F3PiX:
Step 1
first comment

F3PiX:
@carpodaster Could you have a look please? The failing tests are to do with the changes in the routing. Before I adjust the tests, could you check if I am on the right path? 
I had troubles with the redirecting to the PATH_PARENTS in the comments_controller after saving a comment. 
I added an alias to get the right route name. 
Also, do you object against moving the status_updates#show method into the activities controller?

carpodaster:
Hej Maud 👋 

You moved the `status_updates#show` to the ActivitiesController. I would leave it where it is – it isolates the commenting feature to just activities of kind `status_update`, otherwise it interferes with feed entries.

The `status_updates/show` template can hold all the comment code. You can look at `projects/show` for inspiration (with the exception that we cannot lock comments for status updates … yet). 

The problems with `PATH_PARENTS` / redirection: I'm thinking it tries to redirect to the activities_controller, right? In that case: we probably need a special comments controller that defines `PATH_PARENTS` differently but ideally, we can get around that.

@michaelem, do you have a hint on how to deal with PATH_PARENTS and/or the CommentsController subclassing here?

carpodaster:
> Extend to feed_entries (instead of @status_update) (Maud)

Before we do this, we should investigate how many feed updates are posted and if they are posted instead of a mini log update. Feed updates come from full blown blog sites and they already have a sophisticated comment system. The feed update show partial should maybe make the link to the original page more visible so that people willing to comment can do that on the origin page.

Wdyt?

F3PiX:
Hi  👋  
Aha, the comments on feed_entries were the reason I was looking for a solution to move the #show method out of the statusupdates controller. 
AFAIK there is only one team currently using the feed instead of the status update input. 
If I can leave the #show method in the status updates controller, that may solve the redirecting at once (as far as I remember that may be the case, but I'll have to check it tonight).
In that case, we definitively start with status_updates comments only. :-) 
Keep you posted.

F3PiX:
I moved the comments back to the status_updates_controller. I had to pull the [same trick](https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/528/commits/acbd1222caa895bd0aa0252fda052de46e89b212#diff-21497849d8f00507c9c8dcaf6288b136R23) with the routes to make the redirecting working.
(The failing tests are about the routing in the atom feed looking for the non-aliased route.)

So, what to do? Is this ok? Or add a comments_controller and tackle PATH_PARENTS ? 
Or something else?
( @carpodaster, @michaelem  )

app/views/status_updates/show.html.slim
```diff
@@ -24,12 +24,12 @@ h2#comments.header
       .form-inputs.clearfix
         = f.input :commentable_id, as: :hidden
         = f.input :commentable_type, as: :hidden
-        = f.input :text, as: :text, label: "Comment on this project", input_html: { rows: 5 }, hint: 'You can use basic <a href="http://en.wikipedia.org/wiki/Markdown" target="_blank">Markdown</a> here.'.html_safe, required: true
+        = f.input :text, as: :text, label: "Add a comment", input_html: { rows: 5 }, hint: 'You can use basic <ahref="http://en.wikipedia.org/wiki/Markdown" target="_blank">Markdown</a> here.'.html_safe, required: true
```
michaelem: there is a typo here, missing space in 'ahref'

michaelem:
@carpodaster @F3PiX PATH_PARENTS is intended to provide a way to prepend a namespace if one subclasses the comments controller (compare [here](https://github.com/rails-girls-summer-of-code/rgsoc-teams/blob/master/app/controllers/rating/comments_controller.rb)).

The issue at hand is that we get to the comments controller which redirects back to the object that the comment was created on and that object is an Activity. Thus redirect_to redirects to the activity_path(activity). That we have differnt kinds of activity is nothing that rails knows about.

So the current fix works because it creates an activity_path, but it would fall apart as soon as we have another commentable to which we want to redirect to within the root namespace.
Another possible solution would be to subclass the comments controller and have something like a StatusUpdates::CommentsController? This Controller then could live under /status_updates/comments and always redirect to a status update.

Kinda thinking out loud here, but does that sound reasonable?

michaelem:
@F3PiX other than the redirect thing (which is more of an issue with the existing codebase than your pull-request) this looks very nice. If it helps I can try to explain what is going on there in a more real time medium (slack?) tomorrow.

F3PiX:
@michaelem Sounds very interesting and I would love to learn more about it. I will ping you on Slack, thanks!

F3PiX:
Here is a different solution. With a lot of guidance of @michaelem 🙇 , and to and fro the PATH_Parents.
Is this is a better solution than the previous one?

michaelem:
I think so but I'm a bit biased 😅

app/controllers/status_updates/comments_controller.rb
```diff
@@ -0,0 +1,20 @@
+class StatusUpdates::CommentsController < CommentsController
+
+  def create
+    comment = Comment.new(comment_params)
+    status_update = comment.commentable
+
+    if (comment.text.present? && comment.save)
+      anchor = ActionView::RecordIdentifier.dom_id(comment)
```
carpodaster: This anchor isn't used.

config/routes.rb
```diff
@@ -20,7 +20,12 @@
   resources :conferences
   resources :attendances
   resources :contributors, only: :index
+
   resources :status_updates, only: :show
+  namespace :status_updates do
```
carpodaster: I'm not too sure about introducting this namespace. It doesn't really group anything and only holds a single controller.

Of course it not communicated anywhere, not is it set in stone, but I like to think the namespaces serve the purpose of grouping authorization for the different user roles in the app.

What do you think about a top-level controller named `StatusUpdateCommentsController` ?

F3PiX: I think the `StatusUpdateCommentsController` is a very good idea. This probably was a leftover from the path_parents-path I travelled ;-) I will have a look tomorrow.

app/controllers/status_update_comments_controller.rb
```diff
@@ -4,8 +4,8 @@ def create
     comment = Comment.new(comment_params)
     status_update = comment.commentable
 
-    if (comment.text.present? && comment.save)
-      anchor = ActionView::RecordIdentifier.dom_id(comment)
+    if comment.text.present?
+      comment.save
```
F3PiX: Let me try to fix this with a validation

app/models/comment.rb
```diff
@@ -8,6 +11,10 @@ class Comment < ActiveRecord::Base
   before_save :set_checked
   after_commit :notify!
 
+  def is_supervisor_check?
+    commentable.is_a? Team
+  end
+
   def for_application?
```
F3PiX: Should I refactor the `is_supervisor_check?` and `for_application?` check in one method that checks the commentables?

app/controllers/activities_controller.rb
```diff
@@ -7,6 +7,7 @@ class ActivitiesController < ApplicationController
 
   def index
     @team = Team.find_by id: params[:team_id] if params[:team_id]
+    flash.now[:notice] = "You can add comments now! Click on the title of a post and see what emerges"
```
F3PiX: Temp solution until we have a 'Add a comment' feature on the index page.

carpodaster: Does an 'Add comment' link on the index page even work? A comment always references a single record so we can only really offer that option for a show action.

I would remove the flash. It's good to advertise a new feature but the flash system is not the right channel. I've announced the last major changes via twitter; @anasofiapinho retweeted it and I hope that at least a few students interested in the Teams App caught the news. If not – they'll learn about it by some other means eventually 😄

app/models/comment.rb
```diff
@@ -1,4 +1,7 @@
 class Comment < ActiveRecord::Base
+
+  validates :text, presence: true, unless: :is_supervisor_check?
```
carpodaster: I don't know what side-effects this may cause. Sure, it only fires when the `is_supervisor_check?` is true, but I feel like validating on the two dimensions "by whom" (e.g. supervisor, public user, …) and "what kind of commentable" violates the idea of polymorphic commentable.

An idea would be to have something like a StatusUpdateComment [form object](https://medium.com/@ryakh/rails-form-objects-84b6849c886e) that encapsulates the validation, but I think that's too much for this PR.

It's not ideal, either, but I would revert to the `comment.text.present? && coment.save` check for the time being and remove the validation.

app/controllers/activities_controller.rb
F3PiX: > Does an 'Add comment' link on the index page even work? 
Don't know yet :-)

F3PiX:
I was thinking: what if we think the other way around? Say we had the validation for `presence` of `comment.text` in place, and polymorphic comments, before we had Supervisor checks and comments? Would we move the validation to the child comments controllers, like it is now? Or would we solve it by changing the flow after the Supervisor clicks the `Checked` button? 
Is this worth exploring?

carpodaster:
Sorry Maud for not getting back to you earlier. I would leave the current validation (or lack thereof) as it is and deal with it at another time. I don't know where to go with this myself and probably need to meditate on this with my text editor of choice ⛩

Looking at the check boxes above, what else is missing from your perspective?

F3PiX:
No sorries please. :-) 

Okay on the validation meditation. 

Katrin-k and me want to pair on adding some javascript to improve UX.  That will not happen before next week. If we want to have comments before the summer is over :-) we can merge this one as is.  We can open a new PR for the remaining tasks.

Shall I have a look where the conflicts come from? I can do that tonight. Other than that, to me it is ready for merging.

F3PiX:
I can't find anything obvious in this PR that is in conflict with Master, so I think it is something that changed in Master after this PR started.

carpodaster:
Hey Maud, you can resolve the merge conflicts when you expand the "command line instructions"  (only step 1 though) in the greyed-out merge area below; it should be displayed for you as well since you're a member of the org.

When you do the git pull in step no. 1 you should get the merge conflict (in `routes.rb`). You can resolve it with `git mergetool`. When you've resolved the mc, commit what's already been staged push it to your remote branch again. This should update the PR.

F3PiX:
This looks weird.

F3PiX:
Moved to #536.

Issue #534, opened 2016-09-06
Preview function for daily student updates


izuroxx:
@carpodaster @ramonh 
During the past weekend I've created preview function for student updates -> based on preview function for projects, which I recently finished, wanted to practice off-hours what i did in #530 .
I wanted to share this with you and ask if you would like to add this kind of feature to student updates? 

[video demo here](https://vimeo.com/181628590)

There are a few things in my current version ([my branch](https://github.com/izuroxx/rgsoc-teams/tree/student_updates_preview)), that still need fixing (like f.ex. "Back" button under preview for edit student update, that will confuse everybody), which I will fix asap.

carpodaster:
Hi @izuroxx – that's amazing! Looking forward to it :)

ramonh:
Super neat, thanks @izuroxx! I look forward to see how it looks! 🚀

F3PiX:
Wow that is super cool!!!

PR #535, opened 2016-09-08
WIP: Student updates preview

izuroxx:
ref #534 
[new demo video here](https://vimeo.com/181903066) :headphones: 

@carpodaster @ramonh 
I wanted to not interfere much with existing views and partials and not rearrange much, so came up with a way to deal with `Back` button in edit template (it was confusing while on preview tab). I've created `student_updates_helper` with `links_to_student_updates` method that checks if it's `edit` action and displays the button if so. What do you think about this?
Also I'm a bit concerned about another thing: while creating a new status update & also while previewing it, we have a `Previous updates` table of all the status updates below (photo).  Maybe while on preview tab, only the preview of student update should be visible?
Please let me know what you think.
Thank you :bow: 

student_updates


app/views/projects/new.html.slim
```diff
@@ -6,15 +6,15 @@ nav.page
 
 p.well Please make sure you have read our #{link_to "mentor guide", 'http://railsgirlssummerofcode.org/guide/projects/'} before you submit your Open Source project for the upcoming Rails Girls Summer of Code.
 
-= simple_nested_form_for @project, data: { js: "project-form" } do |f|
+= simple_nested_form_for @project, data: { js: "form" } do |f|
```
carpodaster: Is it better to name it `"form-with-preview"` instead of just `"form"`? I find it to be less generic.

app/helpers/status_updates_helper.rb
```diff
@@ -0,0 +1,7 @@
+module StatusUpdatesHelper
```
carpodaster: I wouldn't define this in a module since you're only using it once. You can easily write this inline in `app/views/students/status_updates/_form.html.slim`

app/helpers/status_updates_helper.rb
```diff
@@ -0,0 +1,7 @@
...
+  def links_to_status_updates
+    if params[:action]=="edit"
```
carpodaster: I _think_ there's also a helper method called `controller_action`. It's a bit more semantical than querying the params hash.

F3PiX:
I am not totally convinced about the table layout for the previous status updates. Is it a necessary in combination with the preview?

carpodaster:
@F3PiX The preview function just swaps the form with a rendered preview. Everything else on the page remains the same, including the table view of past status updates.

carpodaster:
Let's do this – thanks a lot @izuroxx !

F3PiX:
Haha nevertheless, I am still not convinced about the table view :-) But the preview is TOO cool. Love it.

izuroxx:
Thank you!!!! :bow: :tada:

Issue #542, opened 2016-09-29
Bug: the 'conferences' page is broken


mkalininait:
Can't open the 'Conferences' page: https://teams.railsgirlssummerofcode.org/conferences
![screen shot 2016-09-28 at 21 24 42](https://cloud.githubusercontent.com/assets/1745735/18952110/c728ad1a-8648-11e6-8f23-28b801a69e2b.png)

Yesterday this page worked; then I assigned some conferences to the students and added a new one (GitHub Universe). After that I found the 'Conferences' page broken.

F3PiX:
I have only 10 minutes to do something about it, and this is what I found.
In the [middle of a method in applications_helper](https://github.com/rails-girls-summer-of-code/rgsoc-teams/blob/master/app/helpers/application_helper.rb#L129), `user` becomes `nil`.
I can do a quick fix like this:
Add `if user.present?` to 
```link_to(user.name.present? ? user.name : user.github_handle, user, class: attendance.confirmed? ? 'confirmed' : '')```

I checked locally, and it will show the Conferences list.
Would this help for now?

ramonh:
Ooh! Thanks for the feedback @F3PiX. Could it be that these users you assigned the conferences to no longer exist? Hm.

ramonh:
Seems to be fixed after merging #543. Thank you @F3PiX! Could you please confirm, @mkalininait?

mkalininait:
@ramonh Works now! Thank you @F3PiX ! I hope I don't break it again :)

PR #543, opened 2016-09-29
TEMP fix in conferences

F3PiX:
Not even sure if it solves the functionality :-( 
But it will show the conferences list.
#542


ramonh:
Looks great to me, @F3PiX. Thank you very much! Will merge now.

Issue #254, opened 2015-07-08
Contributor JSON should only show orga + members of accepted teams, no?


carpodaster:
It currently also shows coaches and mentors of teams that couldn't be accepted. Do we want to filter?

Ping @sareg0 @anikalindtner @alicetragedy

alicetragedy:
I have to +1 with you on this one :+1:

anikalindtner:
yay for filtering them out!

sareg0:
This is in the process of being fixed. Pull request submitted, but not ready for merge til test is written

PR #545, opened 2016-10-10
only show contributors whose team is accepted

alicetragedy:
contributors.json should only return contributors whose team is volunteer or sponsored (i.e. not nil). Hopefully closes https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/254!

### DIFF
--- a/app/controllers/contributors_controller.rb
+++ b/app/controllers/contributors_controller.rb
@@ -5,7 +5,7 @@ class ContributorsController < ApplicationController
 
   def index
     @contributors = User.with_role(Role::CONTRIBUTOR_ROLES).includes(roles: :team).
-      where("teams.season_id IS NULL OR teams.season_id = ?", Season.current).
+      where("teams.season_id IS NULL OR teams.season_id = ? and teams.kind IS NOT NULL", Season.current).
       references(:teams).uniq
     respond_with @contributors.as_json(json_params)
   end
--- a/spec/controllers/contributors_controller_spec.rb
+++ b/spec/controllers/contributors_controller_spec.rb
@@ -7,8 +7,14 @@ describe ContributorsController do
     let!(:user)      { create(:user) }
     let!(:coach)     { create(:coach) }
     let!(:organizer) { create(:organizer) }
-    
+
     it 'returns json representation of user with relevant roles' do
+      get :index, format: :json
+      expect(assigns(:contributors).sort).to eq [coach, organizer].sort
+    end
+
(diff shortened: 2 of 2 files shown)


alicetragedy:
not sure if there is a better way to chain these sql queries. @ramonh @carpodaster can you take a look when you can? :)

app/controllers/contributors_controller.rb
```diff
@@ -5,7 +5,7 @@ class ContributorsController < ApplicationController
 
   def index
     @contributors = User.with_role(Role::CONTRIBUTOR_ROLES).includes(roles: :team).
-      where("teams.season_id IS NULL OR teams.season_id = ?", Season.current).
+      where("teams.season_id IS NULL OR teams.season_id = ? and teams.kind IS NOT NULL", Season.current).
```
carpodaster: I think that will exclude all orgas since they are not part of a team. I don't have the time to verify it at the moment, but can come back to this tomorrow.

alicetragedy: Also, if this was the case, wouldn't they already be excluded them in the current version? 🙈

F3PiX: I think they are in teams: https://github.com/rails-girls-summer-of-code/rgsoc-teams/blob/master/app/models/team.rb#L115

carpodaster: Hm, you may have found a source code artifact, @F3PiX – I don't recall having assigned anyone to a team with these special names. The _role_ is what makes an organizer:

```ruby
carpmeister.teams.pluck(:name)
# => ["joda", "RubyCats"]

carpmeister.roles.pluck(:name)
# => ["developer", "organizer", "mentor", "coach", "coach", "mentor"]

F3PiX: :ghost:

carpodaster:
@alicetragedy the breaking spec is fixx0rd in `master`. Could you merge it back in or rebase it?

alicetragedy:
@carpodaster awesome! yes, will do \o/

carpodaster:
@alicetragedy I took the liberty of adding the spec that I used locally to verify the inclusion of teamleas orgas. Good news: it works! 👍  It must be something about the operator precedence of `AND` and `OR` in the SQL statement. When I tried to split it into two chained `where` clauses, the newly added test failed. /shrug.

Let's wait for Travis to pick up the change and then meeeeerrrrrge!

alicetragedy:
🎉🎉🎉

Issue #503, opened 2016-07-15
Conferences - small improvements for adding new


agnik5:
- [x] 1/ When trying to add a conference without its name, conference is not being added, but the view is broken. I guess there should be for example some alert like 'fill in the name' or a view informing that a conference haven't been added with "Back to index" option.
  ![screenshot from 2016-07-15 16-00-16](https://cloud.githubusercontent.com/assets/12581810/16878048/48267e62-4aab-11e6-9587-5cc5e6dca753.png)
- [ ] 2/ It's possible to add a conference with end date earlier than start date.
  ![screenshot from 2016-07-15 15-42-21](https://cloud.githubusercontent.com/assets/12581810/16878147/b3c63a5e-4aab-11e6-81bf-13452c3fdc60.png)
  ![screenshot from 2016-07-15 15-44-37](https://cloud.githubusercontent.com/assets/12581810/16878146/b3c58140-4aab-11e6-8532-f02b4fa393a7.png)
- [x] 3/ When adding a conference it's possible to choose not round 1 or 2, but 16, 21, -11 etc. You get a view like conference has been created /but it's not on the list/. There's no alert about incorrect round chosen. I think there should be only two values to chose from /1 or 2/.
- [x] 4/ Similar problem in tickets and flights. You can create a conf with -50 tickets and -40 flight and it gets on the list.
  ![screenshot from 2016-07-15 16-55-35](https://cloud.githubusercontent.com/assets/12581810/16878696/d01a15f2-4aad-11e6-8382-6b627d5a779f.png)
  ![screenshot from 2016-07-15 17-00-40](https://cloud.githubusercontent.com/assets/12581810/16878695/d0173968-4aad-11e6-8a46-524eb38fb678.png)

These are not crucial improvements, as conferences are being added by Orga Team and so risk of mistake is close to zero. Only suggesting.

carpodaster:
Hey @agnik5, 👍 to everything you said 😄 . The numbered items fall all nicely into place as a priority list. I think no. 1 should _really_ be fixed. The `save!`  call raises an exception if the record is invalid. This would have to be changed to a normal `save` (w/o the bang). The form has to be re-rendered when there was an error and redirected in case everything went well.

F3PiX:
There is something strange with the date format as well....
[...]

Issue #463, opened 2016-05-18
Move conferences CRUD to orga namespace


carpodaster:

F3PiX:
I'll gladly take this one. Best to wait until  #512 is finished? To avoid complications.

carpodaster:
Hey @F3PiX, did you link the correct issue? #512 seems unrelated.

F3PiX:
thnx, It is #503.

F3PiX:
Closed by #546

PR #546, opened 2016-10-11
WIP Move conferences CRUD to Orga namespace

F3PiX:
#463

Steps:
- [x]  Add conferences_controller to orga namespace
- [x] Remove CRUD actions from conferences_controller
- [x] Add _table partial for shared views
- [x] Render table from both index views
- [x] Restore sortable columns in table (translate from slim)
- [x] show
- [x] other views
- [x] Check redirects from orga/conferences to conferences
- [x] Check abilities and authorisations
- [x] What to do with routing tests? Destroy :trollface: 

Other todo's:
- [x] Translate first round of conferences from slim to erb and add toggle  
- [x] small issues from #503
  - [x] redirect after failing validation
  - [x] restrict numbers in selection fields
- [x] Add tests for authorisations???
- [x] Make 'back to all'-link staying in relevant namespace

Left-overs
- link back from Calendar view
- restrict end dates that are before the start date #503 -2
- adjust js toggle to show the upcoming round of conferences

### DIFF
--- a/app/views/conferences/_table.html.erb
+++ b/app/views/conferences/_table.html.erb
@@ -27,5 +27,30 @@
 <br>
 <p><%= link_to "View as Calendar", calendar_path %>
 </p>
-<!--TODO first rounds-->
+<!--TODO translate first round to erb-->
 <p><%= link_to "Show/Hide First Round Conferences", "#", onclick: "$('.first-conferences').toggle()" %></p>
+
+<!--.first-conferences style='display:none;'-->
+<!--h1.header-->
+<!--/ = icon('group')-->
+<!--span Conferences (1st round)-->
+<!--table.table.table-striped.table-bordered.table-condensed.conferences-->
+<!--tr-->
+<!--th = sortable 'name', 'Conference'-->
+<!--th = sortable 'location', 'Location'-->
+<!--th = sortable 'starts_on', 'Date'-->
+<!--th Attendees-->
+<!--th Tickets left-->
+<!--th Lightning Talk Slots-->
+
+<!-- - conferences.where(round: 1).each do |conference|-->
+<!--tr-->
+<!--td = link_to conference.name, conference-->
+<!--td = conference.location-->
+<!--td = format_conference_date(conference.starts_on, conference.ends_on)-->
+<!--td = conference.attendees.any? ? links_to_attendances(conference).join(', ').html_safe : '-'-->
+<!--td = conference.tickets_left-->
+<!-- - if conference.lightningtalkslots?-->
+<!--td = "yes"-->
+<!-- - else-->
+<!--td = "no"-->
--- a/app/views/conferences/_table.html.erb
+++ b/app/views/conferences/_table.html.erb
@@ -27,14 +27,11 @@
 <br>
 <p><%= link_to "View as Calendar", calendar_path %>
 </p>
-<!--TODO translate first round to erb-->
+
 <p><%= link_to "Show/Hide First Round Conferences", "#", onclick: "$('.first-conferences').toggle()" %></p>
 
 <div class="first-conferences" style='display:none;'>
-<h1 class=" header">
-<!--/ = icon('group')-->
-<!--span -->
-Conferences (1st round)</h1>
+<h1 class=" header">Conferences (1st round)</h1>
 
 <table class="table-striped table-bordered table-condensed conferences">
   <thead>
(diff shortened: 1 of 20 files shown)


spec/routing/conferences_routing_spec.rb
```diff
@@ -10,19 +10,19 @@
       expect(get('/conferences/1')).to route_to('conferences#show', id: '1')
     end
 
-    it 'routes to #edit' do
+    xit 'routes to #edit' do
```
F3PiX: What should I do with this routing tests? I can move them to an spec/routing/orga/conferences_routing_spec.  But do we need routing_specs in the regular controller as well?

F3PiX:
In order to finish this, I have a few questions:
1) Can I delete the routing tests? Or what do I do with them?
2) When I type `/conferences/new` url, I expect one of two things: either a message that the route doesn't exist, or a message that acces is denied. However, I get a message:  `Couldn't find conference with id=new`.  Is this okay, or did I do something wrong with either authorization or the routing?
3) Because all the CRUD is behind the orga base controller, I removed all `can? :crud, Conference`-checks in the CRUD views. okay, right?
4) I tried to [add a validation](https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/546/commits/50d659c38edb806d58e94752aac9eb80e682e891) to ensure that the end_date will be after the start_date. (issue #503 nr 2) But... it would not allow for ANY end date at all, so I removed it in a later commit. Any suggestions?

(ping @carpodaster , @michaelem , @ramonh )

app/views/conferences/_table.html.erb
```diff
@@ -0,0 +1,64 @@
...
+    <td><%= conference.lightningtalkslots? ? "yes" : "no" %></td>
+  </tr>
+      <% end %>
+    </tbody>
+</table>
+
+<br>
+<p><%= link_to "View as Calendar", calendar_path %>
+</p>
+<!--TODO translate first round to erb-->
```
F3PiX: I will remove this.

spec/routing/conferences_routing_spec.rb
carpodaster: Personally, I don't write routing spec for "normal", ie. restful controllers. I find them to be useful when you do something fancy with parameter parsing, routing contraints, or generally diverging from the expected pathname for a given controller.

The conferences controller is pretty standard. I'd have no problem with deleting these examples 😱

carpodaster:
Hej Maud!

re: 2) You are right, you should get a routing error. Can you verify that there is no `new.html` template in the original view folder? I see it was deleted from the repo in your PR, but maybe it's still hanging around on your computer?

re: 3) Yes, simplifying the authorization is one of the premier goals of separating all the things into different namespaces. I'll shoot from the hip and say: yep, `can? :foo` checks (along with the definitions in `ability.rb` begone! 🔫 

re: 4) Let's concentrate on just moving the files into the orga namespace in this PR. Always be wearing [just one hat at a time](http://moleseyhill.com/blog/2009/05/18/kent-becks-four-hats/).

F3PiX:
Thanks Carsten! 
re: 2) Computer says: no `new.html` in `conferences`. The funny thing is, the `/new`-part is read as if it is a param, so the `show`-view is called.  I think now that it has to do with the conference helper method (here)[https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/546/files#diff-7fbe00a3a9ebba1c66f31e2057134f9aR12], but I did not find a way to solve it. 

re:3) Missed a view things ;-)

re: 4) I am going to call you LinkMeister from now on.  And I will watch my hats. Thnx

F3PiX:
re: 1) I found that the `/users/new`-route behaves exactly the same. => 404 instead of a neat and informative redirect. (TIL: load_and_authorize_resource doesn't work on non-existing routes)
One solution I came up with: restore all `conferences` routes, and add a redirect_to orga_conferences-path in the actions. So, the orga::base_controller takes care of access rights and messages. Wdyt, @carpodaster?

F3PiX:
@carpodaster I thought this was about ready. What do you think?

carpodaster:
@F3PiX thank you! I have a busy weekend ahead but will look into it next Monday

carpodaster:
Thanks a lot @F3PiX - looks good! Let's merge this one :smile:

F3PiX:
:tada: :tulip:

PR #552, opened 2016-11-12
FIX Add date validator to Conferences 

### LINKED ISSUE #503: Conferences - small improvements for adding new

F3PiX:
Closes #503


app/models/conference.rb
```diff
@@ -4,7 +4,9 @@ class Conference < ActiveRecord::Base
 
   has_many :attendances
   has_many :attendees, through: :attendances, source: :user
+
   validates :name, :round, presence: true
+  validate :chronology # custom validator
```
carpodaster: I would go for a more "speaking name". How do you like ` :must_end_after_it_started` or something like this?

app/models/conference.rb
```diff
@@ -15,4 +17,9 @@ def tickets_left
     confirmed_attendances = attendances.select { |attendance| attendance.confirmed }
     tickets.to_i - confirmed_attendances.size
   end
+
+  def chronology
```
carpodaster: This method should be made private; it's just an internal implementation detail / helper that should not be made available as part of the public api.

app/models/conference.rb
```diff
@@ -15,4 +17,9 @@ def tickets_left
...
+    return unless starts_on.present?
```
carpodaster: You should also take an empty `ends_on` into consideration

spec/models/conference_spec.rb
```diff
@@ -23,6 +23,27 @@
...
+    context 'when its end date is before start date' do
+      it 'is invalid' do
+        subject.starts_on = '2016-07-15'
+        subject.ends_on = '2016-07-07'
+        expect(subject).not_to be_valid
+      end
+    end
+  end
+
+
```
carpodaster: An unneeded empty line here :)

app/models/conference.rb
F3PiX: I see what you mean. I tried to find something with more analogy to `validates :name` . Can you live with: `validate :interval` or `validate :time_span` ? If not, I go with your suggestion.

app/models/conference.rb
F3PiX: I thought: the form is taking care of all this. In the form, a value will always be set (today's date). That is where the non-valid dates come from most of the time: when the start date is set in the future, the end date's default will still be today. 
Is it enough if the form handles this? Or shall I still add it to the model?

app/models/conference.rb
carpodaster: Ah, I see what you mean. The mean difference I see, though, is that `name` is a model attribute whereas the custom validation we're adding here isn't. But I'm not super invested into the naming of this helper so you decide! :smile:

app/models/conference.rb
```diff
@@ -21,6 +21,7 @@ def tickets_left
   private
 
     def date_interval
+      return unless starts_on.present?
```
F3PiX: If I delete this line (in previous commit) - because it is not needed for the app to run properly - there are 4 test that fail.
```rspec ./spec/controllers/orga/conferences_controller_spec.rb:18 # Orga::ConferencesController with admin logged in POST create creates a record
rspec ./spec/controllers/orga/conferences_controller_spec.rb:25 # Orga::ConferencesController with admin logged in POST create sets the current season
rspec ./spec/models/conference_spec.rb:9 # Conference should require round to be set
rspec ./spec/models/conference_spec.rb:8 # Conference should require name to be set
```
I can't figure out why?

spec/factories/conferences.rb
```diff
@@ -1,11 +1,12 @@
 FactoryGirl.define do
-  factory :conference do
-    name { [FFaker::CheesyLingo.title, 'Conf'].join ' ' }
-    tickets 2
+    factory :conference do
+      name [FFaker::CheesyLingo.title, 'Conf'].join(' ')
+      starts_on { Time.utc(Date.today.year, 7, 7) }
```
F3PiX: I removed the tickets, because they are not a requirement for validation. Plus, the only test that refers to the ticket attributes, overrides it.

spec/models/conference_spec.rb
```diff
@@ -3,10 +3,15 @@
...
-  it { is_expected.to validate_presence_of(:round) }
+  describe 'dependencies' do
+    subject { FactoryGirl.build(:conference) }
+
+    it { is_expected.to have_many(:attendances) }
+    it { is_expected.to have_many(:attendees) }
+    it { is_expected.to validate_presence_of(:name) }
+    it { is_expected.to validate_presence_of(:round) }
+    # The custom date validator is tested in test 'chronology', below
+  end
```
F3PiX: It didn't work without the wrapper. I also tested the opposite:
`it { is_expected.not_to validate_presence_of(:name) }` and that fails. 👍  
Still not sure if this is the proper solution.

app/controllers/orga/conferences_controller.rb
```diff
@@ -38,9 +38,8 @@ def conference
 
     def conference_params
       params[:conference] ? params.require(:conference).permit(
-        :name, :url, :location, :twitter, :tickets, :flights, :accomodation,
-        :'starts_on(1i)', :'starts_on(2i)', :'starts_on(3i)', :round,
-        :'ends_on(1i)', :'ends_on(2i)', :'ends_on(3i)', :lightningtalkslots,
+        :name, :url, :location, :twitter, :tickets, :flights, :accomodation, :starts_on, :ends_on,
+        :lightningtalkslots,
```
F3PiX: The tests kept failing with the old params, where the dates were split into components. It looks like the components are not needed. (I think Rails converts them automatically?) All tests pass, also no problems with adding and validating dates locally.

F3PiX:
@carpodaster Back to you.

spec/models/conference_spec.rb
```diff
@@ -23,6 +28,26 @@
...
+        subject.ends_on = '2016-11-21'
+        expect(subject).to be_valid
+      end
+    end
+
+    context 'when its end date is before start date' do
+      it 'is invalid' do
+        subject.starts_on = '2016-07-15'
+        subject.ends_on = '2016-07-07'
+        expect(subject).not_to be_valid
```
carpodaster: I think it's `shoulda-matchers` that provides a better check: 
```ruby
expect(subject).to 

- have(1).error_on(:ends_on)

```
(or some such, I haven't checked the API). It's better because the model can be invalid for other reasons than the start/end date order.

spec/models/conference_spec.rb
```diff
@@ -23,6 +28,26 @@
...
   end
 
+  describe 'chronology' do
+    subject { FactoryGirl.build(:conference) }
+
+    context 'when the dates are in the right order' do
+      it 'is valid' do
+        subject.starts_on = '2016-11-20'
+        subject.ends_on = '2016-11-21'
+        expect(subject).to be_valid
```
carpodaster: I think it's `shoulda-matchers` that provides a better check: 
```ruby
expect(subject).to have(:no).errors_on(:ends_on)
```
(or some such, I haven't checked the API). It's better because the model can be valid for other reasons than the start/end date order.

spec/models/conference_spec.rb
carpodaster: Why did it fail? I have an idea: the `data_interval` does a `starts_on > ends_on`. It will fail if either is `nil`.  See comment below.

spec/models/conference_spec.rb
```diff
@@ -23,6 +28,26 @@
...
+      end
+    end
+
+    context 'when its end date is before start date' do
+      it 'is invalid' do
+        subject.starts_on = '2016-07-15'
+        subject.ends_on = '2016-07-07'
+        expect(subject).not_to be_valid
+      end
+    end
```
carpodaster: We need two more tests to be complete: where either `starts_on` and `ends_on` is not set.

app/controllers/orga/conferences_controller.rb
carpodaster: Cool! Maybe that changed with Rails 5.

app/models/conference.rb
```diff
@@ -15,4 +17,10 @@ def tickets_left
     confirmed_attendances = attendances.select { |attendance| attendance.confirmed }
     tickets.to_i - confirmed_attendances.size
   end
+
+  private
+
+    def date_interval
+      errors.add(:end_date, 'must be later than start date') if starts_on > ends_on
```
carpodaster: I would add the error to either `:base` or an existing attribute that's concerned with the validation; e.g. `starts_on` or `ends_on`.

F3PiX:
@carpodaster Can you take a look please? 
It is now like this:  both dates can be left empty, but if one date is set, the other must be set too. 
I think requiring the dates per se is too much. However, if one of the dates is set, the other date must be set too.

app/models/conference.rb
```diff
@@ -15,4 +17,15 @@ def tickets_left
     confirmed_attendances = attendances.select { |attendance| attendance.confirmed }
     tickets.to_i - confirmed_attendances.size
   end
+
+  private
+
+    def date_interval
+      if [starts_on, ends_on].count(&:nil?) == 1
+        errors.add(:base, 'Set both dates or none')
+      elsif starts_on.to_s > ends_on.to_s
```
carpodaster: `starts_on` should be less than `ends_on`, right?

Comparing strings lexically is a hack that sure makes it work. But in that case, it should also work if either is `nil`, no?

What about this:
```ruby
unless [starts_on, ends_on].all?(:present?) && (starts_on < ends_on)
  errors.add(:ends_on, 'must be later than start date')
end
```

spec/controllers/orga/conferences_controller_spec.rb
```diff
@@ -19,7 +21,7 @@
         expect {
           post :create, conference: attributes_for(:conference)
         }.to change { Conference.count }.by 1
-        expect(response).to redirect_to [:orga, Conference.last]
+         expect(response).to redirect_to [:orga, Conference.first]
```
carpodaster: Whoops? Why is it the other way round now?

Also: indentation :wink:

app/models/conference.rb
```diff
@@ -15,4 +17,15 @@ def tickets_left
     confirmed_attendances = attendances.select { |attendance| attendance.confirmed }
     tickets.to_i - confirmed_attendances.size
   end
+
+  private
+
+    def date_interval
+      if [starts_on, ends_on].count(&:nil?) == 1
```
F3PiX: I believe that the reasons for presence validations on the dates itself, is only to make our dev life easier, and not per se helping either the user or data integrity. And, on the other hand, I can imagine that orga's life can become harder if it is never ever allowed to add a conference without dates. So, with those considerations, I opted for this solution. But I am not per se against the presence validations, if you think that other considerations are more important?

app/models/conference.rb
F3PiX: 🙈  Far better

F3PiX: O, wait. I am not sure if that is what I meant. 
I suppose you mean:  `[starts_on, ends_on].all?` (ie "both dates are set, none is nil").

I'll wait for your answer in the previous comment, to be sure that we are aiming for the same thing. 🎯

F3PiX:
Sorry, rebase gone bad. I'll open an new PR.

carpodaster:
Doesn't look bad to me

F3PiX:
Is that new, that merge conflicts can directly be resolved at GitHub?

F3PiX:
Looks like git is going to win this one 🤣  .
@carpodaster If we finish the conversation [about the validation](https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/552#discussion_r94916349) here, then I will open a new PR with the next commit.

app/models/conference.rb
F3PiX: @klappradla We need a decision here about the validation of the dates for a conference. 
The main object for this PR is to not allow a `end date` input to be before the `start date`. 
If you can add your vote for either 'presence validation' or not, I can make the PR ready for review.

klappradla: Hard to say what makes more sense. If you skip the presence validation, you still need to check for the presence of both fields in order to perform the validation on "a before b"...

What's the problem you want to solve after all? Conferences have "invalid" dates, meaning someone basically made a typo when inputting them? Or they don't have dates at all and therefore cannot be processes / used?

F3PiX: The conf dates input is with a date picker, so that is covered. 
The problem I want to solve: it is currently possible to input an end date that is earlier than the start date. 
That is basically solved (I want to implement it differently than I did in this PR).
The next problem that came up: should the dates be `presence validate`d.  
TL;DR~My vote would be to leave out the presence validation. In theory orga's could enter a conf before the dates are set (RailsConf 2019).  I can imagine use cases for that. 

OTOH The data picker will set (both) the dates to today automatically on the client side, so why wouldn't we add the presence validation? It would always pass automatically, and It would make the chronology validation easier.  
Never heard orga's complain about the dates set upfront.~
:rubberduck: Conclusion: I add the presence validation.

F3PiX:
Closing and fresh start in #747

PR #557, opened 2016-11-22
Soft landing for project proposals

carpodaster:
What do you think about how it looks like so far?

bildschirmfoto 2016-11-22 um 23 55 32


carpodaster:
I'm done, functionally. Suggestions for wording+styling welcome (you're wholeheartedly invited to just push to this PR)

PR #559, opened 2016-11-27
Remove redundant strong params for conferences

F3PiX:
The permitted params are set in the orga/conferences_controller. In the non-orga controller, they are never called.

Leftover from #546

### DIFF
--- a/app/controllers/conferences_controller.rb
+++ b/app/controllers/conferences_controller.rb
@@ -16,18 +16,9 @@ class ConferencesController < ApplicationController
     helper_method :conferences
 
     def conference
-       @conference ||= params[:id] ? Conference.find(params[:id]) : Conference.new(conference_params)
+       @conference ||= Conference.find(params[:id])
     end
     helper_method :conference
-
-    def conference_params
-     params[:conference] ? params.require(:conference).permit(
-         :name, :url, :location, :twitter, :tickets, :flights, :accomodation,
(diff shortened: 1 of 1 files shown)


carpodaster:
:confetti_ball: Thanks!

Is it maybe better to also remove the redirect and all the controller actions from `routes.rb` that are not :index and :show? Probably not as part of this PR, but in another one? WDYT?

F3PiX:
Yes, that was the original plan. But when I did that, an url like `.../conferences/new` would raise a 404: file with id: new not found. That is a bit clumsy, isn't it?
I found that you need the routes to be in place, or CanCanCan won't do the polite redirecting. 
See [this commit](https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/546/commits/7b2fb2a2e563a94131dcf5ea3a398b037bff4e89)
Do you agree, or...?

carpodaster:
I would let it fail with a 404 cause that's exactly what is: a resource not found. In production, it will display the proper [404 page](https://teams.railsgirlssummerofcode.org/404) instead of an exception + stack trace, so that's ok I think.

I'm leaning towards removing the two unused actions, but I let you decide. Merge at your leisure!

F3PiX:
Thanks :-)
To celebrate that this is a matter of opinion, I go for YOLO and merge. 🎉

PR #561, opened 2016-11-27
Max 3 conferences added in User Profile

F3PiX:
Solves #414
- Add limit to number of permitted attendances
- Hide 'add conference' button if User has 3 attendances: 
- Add flash alert when user wants to save their profile with more than 3 conferences.

One problem left:
The count of the attendances is updated only when the parent is saved. This means an user is able to add 4 conferences in the Edit Profile form, and will not be notified that 3 = max until after submit.
This is not super user friendly

Todo's
- [ ] Make limit for attendances only, no other attributes
- [ ] Add tests

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -4,6 +4,11 @@ class UsersController < ApplicationController
 
   load_and_authorize_resource except: [:index, :show]
 
+  # limit is currently set on conference attendances only
+  rescue_from ActiveRecord::NestedAttributes::TooManyRecords do |error|
+    redirect_to edit_user_path(@user), alert: error.message
+  end
+
   def index
     @filters = {
       all:        'All',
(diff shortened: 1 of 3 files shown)


F3PiX:
Closing.

alicetragedy:
@F3PiX I'm confused, why did you close this PR? Did something not work right?

F3PiX:
It works, but I am having second thoughts about the server side. It is making the code less flexible and more rigid, for no good reason.  
Now I think it would be better to solve this at the client side with some nifty javascript. The current 'Add button' is probably built in by the nested_form gem. It must be possible to make it check for the number of attendances as well.
(How is your Javascript?)

alicetragedy:
@F3PiX I understand, thanks for the explanation! It's nice to give a little bit of context when closing a PR if there's a reason behind it, that way we know what e.g. didn't work. I haven't looked at the original PR (or at the code) tbh. And I agree about what you said related to the user-friendliness (or rather unfriendliness) of not being notified that there are too many confs chosen until after submitting. This is something that should definitely be solved with JS, and unfortunately my JS is still pretty bad. 🙈 
We could move this discussion to the original issue (https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/414) to give more context of what we're trying to achieve, and see if someone could pair with you on this issue? wdyt?

PR #565, opened 2016-11-29
Javascript: choose max 3 attendances 

F3PiX:
Solves #414

💃

### DIFF
--- a/app/assets/javascripts/max-attendances.js
+++ b/app/assets/javascripts/max-attendances.js
@@ -1,0 +1,23 @@
+$(function() {
+    var fieldsCount,
+        maxFieldsCount = 2,
+        $addLink = $('a.add_nested_fields');
+
+    function toggleAddLink() {
+        $addLink.toggle(fieldsCount <= maxFieldsCount);
+    }
+
+    $(document).on('nested:fieldAdded:attendances', function() {
+        fieldsCount += 1;
+        toggleAddLink();
+    });
+
+    $(document).on('nested:fieldRemoved:attendances', function() {
+        fieldsCount -= 1;
+        toggleAddLink();
+    });
+
+    // count existing nested fields after page was loaded
+    fieldsCount = $('form .fields').length;
+    toggleAddLink();
+});
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -4,6 +4,11 @@ class UsersController < ApplicationController
 
   load_and_authorize_resource except: [:index, :show]
 
+  # limit is set on conference attendances
+  rescue_from ActiveRecord::NestedAttributes::TooManyRecords do |error|
+    redirect_to edit_user_path(@user), alert: error.message
+  end
+
   def index
     @filters = {
       all:        'All',
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -80,7 +80,7 @@ class User < ActiveRecord::Base
   validates :homepage, format: { with: URL_PREFIX_PATTERN }, allow_blank: true
   validate :immutable_github_handle
 
-  accepts_nested_attributes_for :attendances, allow_destroy: true
+  accepts_nested_attributes_for :attendances, allow_destroy: true, limit: 3
   accepts_nested_attributes_for :roles
 
   before_save :sanitize_location


F3PiX:
A bit more context:
With less than 3 conferences chosen you get a button to add one more:
with 2 conferences

With 3 conferences, the button disappears:
with 3 conferences

And it only happens with the `:attendances`, not with other nested attributes (like  number of members in a team.)

app/assets/javascripts/max-attendances.js
```diff
@@ -0,0 +1,23 @@
+$(function() {
+    var fieldsCount,
+        maxFieldsCount = 2,
+        $addLink = $('a.add_nested_fields');
```
carpodaster: JS / jQuery n00b question: the `$` only denotes that it's a jQuery DOM object, right? It's not some magic JS trick?

app/assets/javascripts/max-attendances.js
```diff
@@ -0,0 +1,23 @@
...
+
+    function toggleAddLink() {
```
carpodaster: This defines a named function in the global namespace, right? I don't know if that is cool. But then again, I am not a JS person :8ball: :sweat_smile:

Come to think of it: are the vars `fieldsCount`, `maxFieldsCount`  and `$addlink` global as well?

app/assets/javascripts/max-attendances.js
```diff
@@ -0,0 +1,23 @@
...
+        $addLink.toggle(fieldsCount <= maxFieldsCount);
+    }
+
+    $(document).on('nested:fieldAdded:attendances', function() {
```
carpodaster: This is the event fired by simple_form?

F3PiX:
And I am not sure enough to answer the JS questions. Do we know a JS wizzard who can take a look?

About the validations: I tried the client side validation first. That means even an orga can't decide to allow a 4th attendance. And I can think of scenario's where that may be wanted.

I added a validation, and put it in the user model, because everything around the `nested_attributes_for :attendances` are handled there. Okay? Or is it better to put it in the attendance model?

app/assets/javascripts/max-attendances.js
klappradla: Yes, the whole JS file is an adaption from an example in [their wiki](https://github.com/ryanb/nested_form/wiki/How-to:-limit-max-count-of-nested-fields) - so best practice I guess.

app/controllers/users_controller.rb
```diff
@@ -4,6 +4,11 @@ class UsersController < ApplicationController
 
   load_and_authorize_resource except: [:index, :show]
 
+  # limit is set on conference attendances
```
klappradla: It looks kinda weird to me to rescue these model-validation-related exceptions in a controller @F3PiX. I guess moving this closer to the User model and its validations would make more sense? 

You could either pack it into some `before_save` or whatever callbacks... Or you just don't use the `limit` option for the association but instead just a "normal validation.
```ruby
validates :attendances, length: { maximum: 3 }
```
should work as well I think.

However, since submitting an invalid number of conference attendances is already prevented on client side and then fully enforced on DB level (because save will never happen if the error is raised), I guess its also totally fine to give the client a 500. Because in this case *"something really went wrong"*, e.g. someone by hand manipulated the form, the JS is not working, etc. And these are probably errors we should not cover but act upon.

klappradla:
I haven't done fontend code in a while @F3PiX, so just some thoughts I have when looking at this PR, eventually not really relevant 🙈 

So first, I think it's weird, that this piece of code is **always** evaluated completely, even if the client does not show the respective form (80% of all cases I think). Also, you're right, the application code, e.g. having max 3 attendances, should not be tied as closely to the fontend code. I prefer to use data attributes here to just pass some data from Ruby to JS 😉 

So this would look aprox like this:
```javascript
$(function() {
    var fieldsCount
    var maxFieldsCount
    var $addLink
    var $form = $("[data-behavior='attendances-form']")

    function init() {
        maxFieldsCount = $form.data('max-fields');
        $addLink       = $("[data-behavior='add-link']")
        fieldsCount    = $form.fields.length
        _run()
    }

    function _run() {
        _toggleAddLink()

        $form.on('nested:fieldAdded:attendances', function() {
            fieldsCount += 1;
            _toggleAddLink();
        });

        $form.on('nested:fieldRemoved:attendances', function() {
            fieldsCount -= 1;
            _toggleAddLink();
        });
    }

    function _toggleAddLink() {
        $addLink.toggle(fieldsCount <= maxFieldsCount);
    }
      
    if ($form.exists()) init()
});
```

I did not run this code, just wrote it straight here in the editor - so it prob won't work 😉  Just added it to easier explain what I mean: init all the callbacks only if the respective form is actually there and get more data from the application.

app/controllers/users_controller.rb
F3PiX: > It looks kinda weird to me to rescue these model-validation-related exceptions in a controller @F3PiX. I guess moving this closer to the User model and its validations would make more sense?

I think so, yes.

F3PiX:
I don't know how to solve the javascript. I'd say: I close this for now, open a new PR with a validation of max 3 attendances on the model. And I open a new issue for finding a nice frontend solution.

okay?

PR #567, opened 2016-12-06
Upgrade to Rails 5

### LINKED ISSUE #549: Upgrade to Rails 5

ramonh:
This should hopefully resolve #549 

There's one deprecation warning I couldn't figure out, namely the following:
`DEPRECATION WARNING: ActiveRecord::Base.raise_in_transactional_callbacks= is deprecated, has no effect and will be removed without replacement. (called from <top (required)> at /Users/pudidis/coding/rgsoc-teams/config/environment.rb:5)`

This would appear to have been a stopgap between Rails 4.1 and 4.2 and will be removed soon. However, I couldn't find where we defined it.

I've also taken the opportunity to upgrade other gems, while we're at it.


ramonh:
Ack, hold on, something is broken with Travis. Will investigate

ramonh:
Phew! Seems we had to move the `codeclimate` running into `travis.yml`. I hope I've done this correctly!

Gemfile
```diff
@@ -2,7 +2,7 @@ source 'https://rubygems.org'
 
 ruby '2.3.1'
 
-gem 'rails', '~> 4.2.7.1'
+gem 'rails', '~> 5.0', '>= 5.0.0.1'
```
carpodaster: I'm not familiar with passing two version constraints. Shouldn't `~> 5.0`  be enough? I thought it would also match `5.0.0.1`

app/views/activities/_status_update.html.slim
```diff
@@ -2,8 +2,8 @@ li class="activity #{activity.kind}"
   div.row
     div.avatars.hidden-xs.col-sm-4.col-md-3
       - activity.students.each do |student|
-        = image_tag(student.avatar_url, class: "avatar")
-    
+        = image_tag(student.avatar_url, class: "avatar") if student.avatar_url
```
carpodaster: Ah, good catch! `if student.avatar_url.present?` would be even better.

pgaspar:
> There's one deprecation warning I couldn't figure out, namely the following:
`DEPRECATION WARNING: ActiveRecord::Base.raise_in_transactional_callbacks= is deprecated, has no effect and will be removed without replacement. (called from <top (required)> at /Users/pudidis/coding/rgsoc-teams/config/environment.rb:5)`

> This would appear to have been a stopgap between Rails 4.1 and 4.2 and will be removed soon. However, I couldn't find where we defined it.

Could this be it? https://github.com/rails-girls-summer-of-code/rgsoc-teams/blob/feature/rails-5/config/application.rb#L33

(great work by the way!)

ramonh:
Thank you for your feedback, @carpodaster and @pgaspar! 

You were right @pgaspar that did the trick! Great find. 😄 

I didn't know that it's possible to run an update from the command line, @carpodaster, so I actually did id manually. I actually can't seem to find a command for upgrading or updating rails.

carpodaster:
Excellent, @ramonh, thank you!

PR #570, opened 2016-12-08
Fix issue with Github ribbon overlapping Sign In

### LINKED ISSUE #569: Github ribbon blocks Sign In link in some window sizes
Noticed this by chance in development and confirmed it also happens in production. It happens for windows within the 1315px - 1415px range approximately.

github-login

The wrapper:

screen shot 2016-12-08 at 23 10 43

pgaspar:
Fixes #569. Removing the `div.wrapper` seemed to do the trick.

The media queries seem to work as before, but I may have missed something here...

Anyway, the result:

github-login-fix

### DIFF
--- a/app/assets/stylesheets/partials/_github-ribbon.sass
+++ b/app/assets/stylesheets/partials/_github-ribbon.sass
@@ -1,13 +1,5 @@
 // Responsive GitHub Ribbon
 
-.wrapper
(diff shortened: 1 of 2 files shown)


carpodaster:
Rock'n'Roll! Looks superb, thank you Pedro! :bowing_man:

PR #571, opened 2016-12-11
User impersonation (development only)

### LINKED ISSUE #548: Offer a "login as other" user to allow filling in application as team mate in development
In order locally test the entire application process, a 2nd user (ie. a team mate) is needed.

This feature should allow the currently logged in user to be swapped with a selected user from the listing at `users#index` – only in development.

pgaspar:
This PR addresses and hopefully closes #548. It allows users to impersonate other users in development. This is useful for testing the entire application process as both team members.

## How it looks & works

In the public-facing Community page (`users#index`), you'll see an orange button for every user in the list, except yourself.

impersonation

As soon as you hover your mouse over one of these buttons, you get a [bootstrap popover](http://getbootstrap.com/javascript/#popovers) with a small explanation. (this may be a bit over the top 😅 let me know what you think).

screen shot 2016-12-10 at 23 44 21

If you open a user's profile, you'll also see an orange "Impersonate" link. It also shows the same popover on hover.

screen shot 2016-12-10 at 23 44 59

Clicking any of these orange buttons will switch your `current_user` to be the impersonated user. I used a gem called [pretender](https://github.com/ankane/pretender) that also gives us a `true_user` method, which is the original user. This is how clicking the link looks like:

screen shot 2016-12-10 at 23 45 21

You'll notice that when in "impersonation mode", the Sign out link is swapped by a "Stop impersonation" one.

screen shot 2016-12-10 at 21 19 40

Clicking it will stop the impersonation make you _you_ again!

### Keeping this out of production

As you'll see in the code, I'm checking for the production environment at the route level and at the view level. Let me know if you feel I should add some controller-level protection, like a guard clause which is what's being used on the Switch Season Phase feature.

### Tests

I wanted to explore testing the route-level production protections and I (eventually) got it working 🙌 . At the end of it I had a feeling that this may be testing _too deeply_, so if I'm happy to remove/alter if you'd like.

### The orange buttons & other visible changes

I picked the `eye-open` and `eye-close` icons because it was the best I could find in the Font Awesome version we're using. Open to suggestions, but I'm not too worried since this is dev-only.

The Stop impersonation button isn't great - I wanted to make it "Stop Impersonation", but that would break the menu layout for supervisors (they have an extra link in the menu).

The popover that comes up when you hover the buttons may be a bit too much. I'm ok with removing it entirely if you want.

Oh, and the buttons are orange because the Switch Phase button (another development-only feature) was also orange 😄

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,10 +126,12 @@ E.g. `foreman run rails server` or `foreman run rails console`.
     ``` user = User.last``` #or ```user = User.find_by(github_handle: "yourgithubhandle") ```
     ``` user.roles.create(name: "organizer") ```
     You can assign yourself other roles in the same way. If however you assign
-    yourself a student role AND another role, that may lead to unexpected behavior in the app. In that case, remove the student role.    
+    yourself a student role AND another role, that may lead to unexpected behavior in the app. In that case, remove the student role. 
     - Refresh the browser to effectuate. You should see links for organizers. 
 - Once you are an `organizer`, you can add a season and switch between season's phases at
 http://localhost:3000/orga/seasons in your browser.
+- While in development, you are also able to impersonate other users to easily test the system
+as someone else. Go to http://localhost:3000/users while logged in to do that.
 - You are good to go now. Happy coding!
 
 ## Testing
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -20,6 +20,10 @@ class ApplicationController < ActionController::Base
 
   before_action :set_timezone
 
+  # Allow users to impersonate other uses in a non-production environment.
+  # See the `pretender` gem.
+  impersonates :user
+
   def after_sign_in_path_for(user)
     if user.just_created?
       request.env['omniauth.origin'] || edit_user_path(user, welcome: true)
--- a/app/views/users/_impersonate.html.slim
+++ b/app/views/users/_impersonate.html.slim
@@ -1,0 +1,9 @@
+= link_to icon('eye-open'), impersonate_user_path(user), class: 'btn btn-warning btn-sm',
+  method: :post,
+  title: "Impersonate User (for development only)",
+  data: { \
+    toggle: :popover,
+    content: "By clicking this you'll experience the app as if you were #{user.name}. You can use this to locally test the entire application process, by becoming the second team member.",
+    trigger: :hover,
+    placement: 'right auto',
+  }
--- a/Gemfile
+++ b/Gemfile
@@ -24,6 +24,7 @@ gem 'jquery-rails'
 gem 'jquery-ui-rails'
 gem 'bootstrap-sass'
 gem 'font-awesome-sass-rails'
+gem 'pretender'
 
 gem 'sass-rails'
 gem 'coffee-rails'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -186,6 +186,8 @@ GEM
       omniauth (~> 1.2)
     orm_adapter (0.5.0)
     pg (0.19.0)
+    pretender (0.2.1)
+      actionpack
     pry (0.10.4)
       coderay (~> 1.1.0)
       method_source (~> 0.8.1)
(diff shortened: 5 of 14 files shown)


alicetragedy:
This is an awesome feature @pgaspar! 🙌 
I think that since there is not really an icon that really fits with the concept, having the extra popover is definitely a good idea, as it explains what the button does (might be confusing otherwise, with an icon that does not 100% reflect this sort of complicated concept). So from my side, +1 on keeping that.

katrinkampfrath:
This sounds just wonderful. I can't assess the actual code implementation, but the UX appears to be well thought out and hell yeah… frontend work for all those many roles will be a lot easier.

app/controllers/application_controller.rb
```diff
@@ -20,6 +20,10 @@ class ApplicationController < ActionController::Base
 
   before_action :set_timezone
 
+  # Allow users to impersonate other uses in a non-production environment.
+  # See the `pretender` gem.
+  impersonates :user
```
carpodaster: Would it be good to guard this with a `unless Rails.env.production`?

app/views/users/_impersonate.html.slim
```diff
@@ -0,0 +1,9 @@
+= link_to icon('eye-open'), impersonate_user_path(user), class: 'btn btn-warning btn-sm',
+  method: :post,
```
carpodaster: Oh nice, I didn't know you could do these multi-line thingies with slim

app/controllers/application_controller.rb
carpodaster: Or is it enough to not route the impersonation in prod (ie. there is a guard in `routes.rb`) ?

carpodaster:
Oh, and about the popovers: those are the most beautiful usage hints the Teams App has ever seen! :heart:

app/controllers/application_controller.rb
pgaspar: We can certainly add more protection @carpodaster. I don't know if doing it here is the best idea though, because this call creates the `true_user` helpers (and a couple more).

For now, I think it would work because I think we're only using `true_user` after we check that we're not in production in the views. I'll try it out and see if it works!

pgaspar:
Thank you for the awesome comments @alicetragedy @katrinkampfrath @carpodaster 😸!

app/controllers/application_controller.rb
carpodaster: You're right – it's probably well protected as it is. Thanks for clearing it up!

carpodaster:
Merge at your own leisure! 😄

F3PiX:
This looks awesome! 
It would be great if you would add a line or two to the [Quick Start guide ](https://github.com/rails-girls-summer-of-code/rgsoc-teams#quick-start)as well? Maybe we can simplify the modus operandi as well.

pgaspar:
Great suggestion @F3PiX 👍 @carpodaster and others, check the text I added to the readme and let me know if it's ok so we can merge! 😺

carpodaster:
Looks great :+1:

README.md
```diff
@@ -126,10 +126,12 @@ E.g. `foreman run rails server` or `foreman run rails console`.
     ``` user = User.last``` #or ```user = User.find_by(github_handle: "yourgithubhandle") ```
     ``` user.roles.create(name: "organizer") ```
     You can assign yourself other roles in the same way. If however you assign
-    yourself a student role AND another role, that may lead to unexpected behavior in the app. In that case, remove the student role.    
+    yourself a student role AND another role, that may lead to unexpected behavior in the app. In that case, remove the student role.
```
F3PiX: I wonder: do we still need to add the roles, like student, to ourselves? If not, we can delete the lines 128 + 129.

pgaspar: I think that instruction is probably still valid. I guess it depends on what you want to do, and there can probably be a role that is not covered by our initial fake data, so there would be no one to impersonate for that role.

We can definitely improve this part of the readme in a separate PR if necessary. For now, I'm merging this one 🎉

alicetragedy:
I'd also be up for merging this — I think the info about roles in the readme should still be kept for now (I consider it to still be valid information) 👍

Issue #499, opened 2016-07-13
Refactor season switch into a separate class


carpodaster:
The season switch methods [add some complexity](https://codeclimate.com/github/rails-girls-summer-of-code/rgsoc-teams/Season?from_sha=17c93b66&to_sha=1b9c7d0b) to the `Season` class according to CodeClimate.

There are three steps:
- [ ] Extract default dates for a newly created season into some form of constant
- [ ] Re-use these constant default values in `set_application_dates` (where applicable) and `back_to_reality`
- [ ] Move the `fake_*_phase` and `back_to_reality` into a service class

F3PiX:
Yes, I saw the code climate score earlier this week and it made me very sad ;-) I would love to improve it.

carpodaster:
Sure Maud!

F3PiX:
Goal accomplished 🎉 
![season after refactoring](https://cloud.githubusercontent.com/assets/6314687/21317835/19c29970-c607-11e6-92ac-a3408bba2e5a.png)

PR #572, opened 2016-12-17
Move season phase switch out of model

F3PiX:
Take 1 on issue #499
- Store dates in constants
- Move switcher methods out of Season class

### DIFF
--- a/app/controllers/orga/seasons_controller.rb
+++ b/app/controllers/orga/seasons_controller.rb
@@ -48,16 +48,16 @@ class Orga::SeasonsController < Orga::BaseController
   # by opening and closing the corresponding links in the nav bar
   def switch_phase
     return if Rails.env.production?
-    @season = current_season
+    @switch = current_season.phase_switcher
       case params[:option]
       when 'Proposals'
-        @season.fake_proposals_phase
+        @switch.fake_proposals_phase
       when 'Application'
-        @season.fake_application_phase
+        @switch.fake_application_phase
       when 'CodingSummer'
-        @season.fake_coding_phase
+        @switch.fake_coding_phase
       when 'RealTime'
-        @season.back_to_reality
+        @switch.back_to_reality
       end
     redirect_to orga_seasons_path, notice: "We time travelled into the #{params[:option]} phase"
   end
--- a/app/controllers/orga/seasons_controller.rb
+++ b/app/controllers/orga/seasons_controller.rb
@@ -48,16 +48,15 @@ class Orga::SeasonsController < Orga::BaseController
   # by opening and closing the corresponding links in the nav bar
   def switch_phase
     return if Rails.env.production?
-    @switch = current_season.phase_switcher
       case params[:option]
       when 'Proposals'
-        @switch.fake_proposals_phase
+        Season::PhaseSwitcher.fake_proposals_phase
       when 'Application'
-        @switch.fake_application_phase
+        Season::PhaseSwitcher.fake_application_phase
       when 'CodingSummer'
-        @switch.fake_coding_phase
+        Season::PhaseSwitcher.fake_coding_phase
       when 'RealTime'
-        @switch.back_to_reality
+        Season::PhaseSwitcher.back_to_reality
       end
     redirect_to orga_seasons_path, notice: "We time travelled into the #{params[:option]} phase"
   end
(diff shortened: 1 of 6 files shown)


F3PiX:
@carpodaster Before I solve the tests: Is this the way to go?

app/models/season.rb
```diff
@@ -1,8 +1,17 @@
 class Season < ActiveRecord::Base
+  include SeasonPhaseSwitcher
+
   validates :name, presence: true, uniqueness: true
 
   before_validation :set_application_dates
 
+  SUMMER_OPEN   = '7, 1'
```
carpodaster: Pulling the default dates into a constant is a very good thing! :+1: 

I don't believe you can call `Time.utc` with a string parameter (see line 30). You probably want to use an array with 2 integers instead. But even then you'll have to splat the argument call like this:

```ruby
Time.utc(year, *SUMMER_OPEN) # 1 jul
```

A comment on what the values for the constants mean (= month, day) would be good for new devs or future versions of ourselves.

A different approach could be assigning Procs or Lambdas to the constants but that may get a little too verbose.

app/models/concerns/season_phase_switcher.rb
```diff
@@ -0,0 +1,50 @@
+module SeasonPhaseSwitcher
```
carpodaster: I would make this a class. Extracting instance methods into a module which is only used by that one class gains little in terms of true separation. On the contrary: it adds cognitive load since the functionality is not stated explicitly anymore. 

How about a class nested in the `Season` namespace?
```ruby
# app/models/season/phase_switcher.rb
class Season::PhaseSwitcher
  # …
end
```

The class can operate on `Season.current` directly.

F3PiX:
@carpodaster I don't understand how to implement so that "The class can operate on Season.current directly."  So I tried this.

carpodaster:
What I meant is that since the season switch only makes sense on the current season, the season switcher class doesn't necessarily be passed a season instance. It can grab it itself by calling `Season.current`. Your approach definitely looks cleaner, I was only thinking if passing a season instance to the constructor would suggest that it actually makes any sense to pass a different season.

app/controllers/orga/seasons_controller.rb
```diff
@@ -48,16 +48,16 @@ def destroy
   # by opening and closing the corresponding links in the nav bar
   def switch_phase
     return if Rails.env.production?
-    @season = current_season
+    @switch = current_season.phase_switcher
```
carpodaster: You can instantiate the class directly without calling a method on `current_season`. Now that there is a dedicated class for it, it doesn't have to be the `Season`'s concern to switch phases anymore.

app/models/season/phase_switcher.rb
```diff
@@ -1,8 +1,11 @@
-module SeasonPhaseSwitcher
-  extend ActiveSupport::Concern
+class Season::PhaseSwitcher
+
+  def initialize(season)
```
carpodaster: This is what I meant in my other comment: passing the season as an argument is not really necessary since the switcher can grab the current season from the DB itself.

app/controllers/orga/seasons_controller.rb
carpodaster: Ok, a bit of a crossroads here:

I would either

a) remove `Season#phase_switcher` from the the public interface, forward the `fake_*_phase` methods to a private switcher instance transparently and keep calling `fake_*_phase` on `current_season` 

or

b) leave `Season` completely out of the phase switching business by calling the class directly (see comment above). This would also mean moving the phase-switching examples into a separate spec file for `Season::PhaseSwitcher`.

I think b) is the better way to go, but I let you decide :)

app/models/season/phase_switcher.rb
carpodaster: Oh, and when the `@season` ivar is gone, it's questionable if we still need an _instance_ of `Season::PhaseSwitcher` at all; there is no state left and all the methods are self-contained. We can make all the methods class methods.

F3PiX:
Thanks @carpodaster for your help.
I chose option b) 
> b) leave Season completely out of the phase switching business by calling the class directly (see comment above). This would also mean moving the phase-switching examples into a separate spec file for Season::PhaseSwitcher.

carpodaster:
Awesome, Maud, thanks!

PR #560, opened 2016-11-27
Fix date formats in conference views

F3PiX:
(#503, [comment](https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/503#issuecomment-235365223))
  
This  date format with a comma:
image

.. is now without the comma.

schermafbeelding 2016-11-27 om 18 31 35

Checked all other usages of the conference_date_formatter, everything is fine. 🏎

PR #573, opened 2016-12-18
Fix conference format on different profile pages

F3PiX:
In the never ending story of comma's and spaces:

The conference texts have different formats on the teams profile page and the user profile page.
This fix solves a missing space on the user profile page when the conference location is given.

The formatting was [suspicious](https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/560#issuecomment-263961154) and turned out NOT to be okay:
Before:
user page

After:
user page after

(Apart from the leading space, that I couldn't solve. I'll leave that for now.)

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -135,8 +135,10 @@ module ApplicationHelper
 
   def links_to_conferences(conferences, extra_info = false)
     conferences.map do |conference|
-      text = conference.name
-      text += " (#{conference.location}#{format_conference_date(conference.starts_on, conference.ends_on)})" if extra_info
+      text = conference.name + " ("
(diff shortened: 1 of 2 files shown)


carpodaster:
> The conference texts have different formats on the teams profile page and the user profile page.

Would it make sense to remove the difference?

F3PiX:
It would be nice to DRY this up, but it requires quite a bit of [refactoring ](https://github.com/rails-girls-summer-of-code/rgsoc-teams/blob/master/app/helpers/application_helper.rb#L136). Which affects other helper methods etc.
So I decided to only do the bug fix.
I can open a new issue, and sum up the questions that will be raised by that change?

carpodaster:
Closing in favour of #590

PR #575, opened 2016-12-19
Limit user' s attendances to max 3

F3PiX:
Part 1 of #414 
This adds a simple limit to the number of conferences a user can attend.
In user profile (user/:id/edit):
* On submitting a form, the 4th conference will not be saved.
* On loading the form, when more than 2 conferences are saved before, the link to 'Add another conference' will not show up.

TODO after this PR:
- [ ] better UX (see #565)
- [ ] make the attendances aware of raffle rounds

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -59,6 +59,8 @@ class User < ActiveRecord::Base
     "over 60",
   ]
 
+  MAX_ATTENDANCES_PER_ROUND = 3
+
   include ActiveModel::ForbiddenAttributesProtection
   include Authentication::ActiveRecordHelpers
   include ProfilesHelper
@@ -103,11 +105,19 @@ class User < ActiveRecord::Base
 
   validates :name, :email, :country, :location, presence: true, unless: :github_import
 
-  accepts_nested_attributes_for :attendances, allow_destroy: true, limit: 3
+  accepts_nested_attributes_for :attendances, allow_destroy: true, reject_if: :too_many_attendances
+  # , reject_if: -> (attendance) {attendances.conferences Conference.in_current_season.include? attendance.conference}
   accepts_nested_attributes_for :roles, allow_destroy: true
 
   before_save :sanitize_location
   after_create :complete_from_github
+
+  def too_many_attendances
+    this_round_attendances = attendances.collect {
+      |attendance| attendance.conference.in?(Conference.in_current_season)
+    }.count
+    this_round_attendances > MAX_ATTENDANCES_PER_ROUND
+  end
 
   # This field is used to skip validations when creating
   # a preliminary user, e.g. when adding a non existant person
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -103,7 +103,7 @@ class User < ActiveRecord::Base
 
   validates :name, :email, :country, :location, presence: true, unless: :github_import
 
-  accepts_nested_attributes_for :attendances, allow_destroy: true
+  accepts_nested_attributes_for :attendances, allow_destroy: true, limit: 3
   accepts_nested_attributes_for :roles, allow_destroy: true
 
   before_save :sanitize_location
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -105,22 +105,23 @@ class User < ActiveRecord::Base
 
   validates :name, :email, :country, :location, presence: true, unless: :github_import
 
-  accepts_nested_attributes_for :attendances, allow_destroy: true, reject_if: :too_many_attendances
-  # , reject_if: -> (attendance) {attendances.conferences Conference.in_current_season.include? attendance.conference}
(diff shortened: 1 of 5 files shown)


bitboxer:
A test for this is missing here

app/models/user.rb
```diff
@@ -80,8 +80,8 @@ def student
   validates :homepage, format: { with: URL_PREFIX_PATTERN }, allow_blank: true
   validate :immutable_github_handle
 
-  accepts_nested_attributes_for :attendances, allow_destroy: true
-  accepts_nested_attributes_for :roles, allow_destroy: true
```
beanieboi: quick question. what is the reason to remove the `allow_destroy: true` for roles?

app/models/user.rb
```diff
@@ -104,8 +104,8 @@ def student
 
   validates :name, :email, :country, :location, presence: true, unless: :github_import
 
-  accepts_nested_attributes_for :attendances, allow_destroy: true
-  accepts_nested_attributes_for :roles, allow_destroy: true
+  accepts_nested_attributes_for :attendances, allow_destroy: true, limit: 3
+  accepts_nested_attributes_for :roles
```
beanieboi: quick question. what is the reason to remove the allow_destroy: true for roles?

F3PiX: Hi @beanieboi Thanks for your review!
The removal is not in my branch locally. 
Can it be that this happens with merging the master into the branch? IIRC I added the `allow_destroy: true` for roles in another PR. So can it be that this PR overrides the other one?

beanieboi: @F3PiX that makes total sense! I think we should update the PR so that we don't remove it. I also try to figure out the impact of this PR. if we have users in the database with more than 3 attendances, rails would think that the users are invalid and users can't update their profiles (e.g.) anymore. since I don't have access to the database, I also posted that question in slack and hopefully somebody can help us out there.

F3PiX:
@klappradla I tried to find a way to add the `in_current_season` scope to the `limit: 3`, but I don't know how. 
An option is to close this PR and leave it as is? Until someone takes the javascript solution on (as we discussed before). #576

klappradla:
Ok, I see @F3PiX. If I remember correctly, then what you need to archive the scoping is just a scoped association. Looks approx. like this:
```ruby
has_many :attendances, dependent: :destroy
has_many :conferences, through: :attendances
has_many :conferences_in_season,
  -> { where(conferences: { season_id: Season.current.id }) },
  through: :attendances,
  source: :conference
```

Then you need to adjust the calls to `accept_nested_attributes` and the form and your limit should work. Makes sense?

app/models/user.rb
klappradla: For scoping this down to the current season, it should be as I already wrote in the conversation. However, if that's all too complex, what about "just" a validation? I imagine that something like
```ruby
validates_length_of :conferences_in_season, maximum: 3
```
should already be sufficient. Don't you think?

F3PiX: The max of conferences is 3 _per raffle round_. That's why I wanted to set the limit on the `nested_attributes`: so it only applies on form `submit` for the relevant raffle round. I suspect that either solution above doesn't work on the form only?
I can use either the `limit` option or the `reject_if` option and add a Proc or a symbol. So I am thinking something like this: 
```
accepts_nested_attributes_for :attendances, allow_destroy: true, 
reject_if: :too_many_attendances

MAX_ATTENDANCES_PER_ROUND = 3

def too_many_attendances
    this_round_attendances = attendances.collect {
      |attendance| attendance.conference.in?(Conference.in_current_season)
    }.count
    this_round_attendances > MAX_ATTENDANCES_PER_ROUND
  end
```
 NB `round` is an attribute of `conference`. I didn't take that into account yet, just checking if `option + method` is a viable way.

Is this a viable way?

klappradla: Sure, sounds like it should do the job.

F3PiX:
Did I solve the merge conflict in rebase the wrong way? I am not sure where the :skip_validations came from. Same for the :skip_validations trait; it is too sophisticated to be written by me :-)

klappradla:
I don't know what exactly you are talking about @F3PiX, but I'm 99% sure that every `skip_validations` trait you'll find there was probably written by me 😉  It's something like my own *"pattern"*™ that I throw into basically every Rails codebase I'm on :v:

What's the problem with it?

F3PiX:
No problem with the trait at all! I am not sure that I solved the rebase conflicts right and if it should be there at all. Because in the same conflict solving that added this, I _removed_ a thing that _should_ be there. 
(...)
Think I know now what happened, and how to check. Thnks.

F3PiX:
Temporarily blocked, until we figured out how to incorporate the 'round' attribute (see #751 )

F3PiX:
Closed because we are reworking the Conferences functionality.

Issue #154, opened 2015-03-06
users/edit: A role will not be destroyed when you 'remove' + submit


carpodaster:
Computer says no when I try to remove @alicetragedy from Team Origami

![](http://wowalt.com/wp-content/uploads/2014/07/computer-says-no.png)

alicetragedy:
Okay. I just tried it, and it works if I remove myself (obviously not the point, right?)

F3PiX:
I couldn't remove my own roles. Would be great to solve this, it will make the life of next year's Rails Girls so much easier. Someone want to pair with me on this?

carpodaster:
Solved by #577

PR #577, opened 2016-12-19
Fix: remove role from user's profile

F3PiX:
Solves #154
Removing a role from a user's profile (as an admin), did not actually remove the role. Allowing 'destroy's on nested attributes fixes this.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -99,7 +99,7 @@ class UsersController < ApplicationController
         :application_location, :application_minimum_money, :application_money, :application_goals, :application_code_background,
         interested_in: [],
         attendances_attributes: [:id, :conference_id, :_destroy],
-        roles_attributes: [:id, :name, :team_id]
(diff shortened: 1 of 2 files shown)


F3PiX:
<iframe src="//giphy.com/embed/l4Ki01RIvdIQVFhqE" width="480" height="270" frameBorder="0" class="giphy-embed" allowFullScreen></iframe><p><a href="http://giphy.com/gifs/vice-spain-l4Ki01RIvdIQVFhqE">via GIPHY</a></p>
We don't _have_ Capybara. We are not going to set it up just to test this feature, are we?

I did some research, found that testing nested forms is notoriously complex, and I didn't find a solution how to do that other than with Capybara or Cucumber or another way of integration testing. 
Also, I didn't find an example in the Teams app repo with other nested forms, to reuse.

So. I really don't know how to test this. Tips?

F3PiX:
@carpodaster I am still confused about the Capybara testing. Can you explain please?

F3PiX:
We talked about it last year. One day..... ;-)
Thanks.

PR #585, opened 2017-01-02
Refactor season phases call

F3PiX:
Move phases to model. Get rid of the switch statement.

The switch statement has been bugging me since I created it, last year, as a Ruby Nuby. 
The new PhaseSwitcher class looks like the most appropriate place to translate the params to a method call.

### DIFF
--- a/app/controllers/orga/seasons_controller.rb
+++ b/app/controllers/orga/seasons_controller.rb
@@ -48,7 +48,7 @@ class Orga::SeasonsController < Orga::BaseController
   # by showing the corresponding links in the nav bar
   def switch_phase
     return if Rails.env.production?
-    Season::PhaseSwitcher.switcher(params[:option])
+    Season::PhaseSwitcher.distribute(params[:option])
     redirect_to orga_seasons_path, notice: "We time travelled into the #{params[:option]} phase"
   end
 
--- a/app/controllers/orga/seasons_controller.rb
+++ b/app/controllers/orga/seasons_controller.rb
@@ -48,7 +48,7 @@ class Orga::SeasonsController < Orga::BaseController
   # by showing the corresponding links in the nav bar
   def switch_phase
     return if Rails.env.production?
-    Season::PhaseSwitcher.distribute(params[:phase])
+    Season::PhaseSwitcher.destined(params[:phase])
     redirect_to orga_seasons_path, notice: "We time travelled into the #{params[:phase].humanize.titlecase}"
   end
 
--- a/app/controllers/orga/seasons_controller.rb
+++ b/app/controllers/orga/seasons_controller.rb
@@ -73,7 +73,7 @@ class Orga::SeasonsController < Orga::BaseController
   end
 
   def phase
-    params[:phase].to_sym
+    params[:phase].to_sym if phase.in? PHASES
   end
 
   def set_breadcrumbs
--- a/app/controllers/orga/seasons_controller.rb
+++ b/app/controllers/orga/seasons_controller.rb
@@ -45,22 +45,12 @@ class Orga::SeasonsController < Orga::BaseController
   end
 
   # # switch_phase: enables developers to easily switch between time dependent settings in views
-  # by opening and closing the corresponding links in the nav bar
+  # by showing the corresponding links in the nav bar
   def switch_phase
     return if Rails.env.production?
-      case params[:option]
-      when 'Proposals'
-        Season::PhaseSwitcher.fake_proposals_phase
-      when 'Application'
-        Season::PhaseSwitcher.fake_application_phase
-      when 'CodingSummer'
-        Season::PhaseSwitcher.fake_coding_phase
-      when 'RealTime'
-        Season::PhaseSwitcher.back_to_reality
-      end
+    Season::PhaseSwitcher.switcher(params[:option])
     redirect_to orga_seasons_path, notice: "We time travelled into the #{params[:option]} phase"
   end
-
 
   private
 
--- a/app/controllers/orga/seasons_controller.rb
+++ b/app/controllers/orga/seasons_controller.rb
@@ -48,8 +48,9 @@ class Orga::SeasonsController < Orga::BaseController
   # by showing the corresponding links in the nav bar
   def switch_phase
     return if Rails.env.production?
-    Season::PhaseSwitcher.distribute(params[:option])
-    redirect_to orga_seasons_path, notice: "We time travelled into the #{params[:option]} phase"
+    byebug
+    Season::PhaseSwitcher.distribute(params[:phase])
+    redirect_to orga_seasons_path, notice: "We time travelled into the #{params[:phase].humanize}"
   end
 
   private
--- a/app/controllers/orga/seasons_controller.rb
+++ b/app/controllers/orga/seasons_controller.rb
@@ -48,7 +48,7 @@ class Orga::SeasonsController < Orga::BaseController
   # by showing the corresponding links in the nav bar
   def switch_phase
     return if Rails.env.production?
-    Season::PhaseSwitcher.destined(params[:phase])
+    Season::PhaseSwitcher.destined(phase)
     redirect_to orga_seasons_path, notice: "We time travelled into the #{params[:phase].humanize.titlecase}"
   end
 
@@ -70,6 +70,10 @@ class Orga::SeasonsController < Orga::BaseController
       :project_proposals_open_at,
       :project_proposals_close_at
     )
+  end
+
+  def phase
+    params[:phase].to_sym
   end
 
   def set_breadcrumbs
--- a/app/controllers/orga/seasons_controller.rb
+++ b/app/controllers/orga/seasons_controller.rb
@@ -48,9 +48,8 @@ class Orga::SeasonsController < Orga::BaseController
   # by showing the corresponding links in the nav bar
   def switch_phase
     return if Rails.env.production?
-    byebug
     Season::PhaseSwitcher.distribute(params[:phase])
-    redirect_to orga_seasons_path, notice: "We time travelled into the #{params[:phase].humanize}"
+    redirect_to orga_seasons_path, notice: "We time travelled into the #{params[:phase].humanize.titlecase}"
   end
 
   private
--- a/app/controllers/orga/seasons_controller.rb
+++ b/app/controllers/orga/seasons_controller.rb
@@ -73,7 +73,7 @@ class Orga::SeasonsController < Orga::BaseController
   end
 
   def phase
-    params[:phase].to_sym if phase.in? PHASES
+    params[:phase].to_sym
   end
 
   def set_breadcrumbs
--- a/app/models/season/phase_switcher.rb
+++ b/app/models/season/phase_switcher.rb
@@ -60,7 +60,7 @@ class Season::PhaseSwitcher
   end
 
   def self.whitelisted_phases
-    @whitelist_phases ||= [
+    @whitelisted_phases ||= [
         :fake_proposals_phase,
         :fake_application_phase,
         :fake_coding_phase,
--- a/app/models/season/phase_switcher.rb
+++ b/app/models/season/phase_switcher.rb
@@ -7,62 +7,62 @@ class Season::PhaseSwitcher
     fake_application_phase
     fake_coding_phase
     back_to_reality
-    ).freeze
+  ).freeze
+
+  def self.destined(phase)
+    raise ArgumentError.new("#{phase} is not a valid phase") unless phase.in? PHASES
+    self.send(phase)
+  end
+
+  private
 
   def self.season
     Season.current
   end
 
-  def self.destined(phase)
-    raise ArgumentError.new("#{phase} is not a valid phase") unless phase.in? PHASES
-    self.public_send(phase)
-  end
-
-  private
-
   def self.fake_coding_phase
     season.update(
-       starts_at: 6.weeks.ago,
-       ends_at: 6.weeks.from_now,
-       applications_open_at: 4.months.ago,
-       applications_close_at: 3.months.ago,
-       acceptance_notification_at: 2.months.ago
+      starts_at: 6.weeks.ago,
+      ends_at: 6.weeks.from_now,
+      applications_open_at: 4.months.ago,
+      applications_close_at: 3.months.ago,
+      acceptance_notification_at: 2.months.ago
     )
   end
 
   def self.fake_proposals_phase
     season.update(
-       starts_at: 6.months.from_now,
-       ends_at: 9.months.from_now,
-       applications_open_at: 3.months.from_now,
-       applications_close_at: 4.months.from_now,
-       acceptance_notification_at: 5.months.from_now,
-       project_proposals_open_at: 4.weeks.ago,
-       project_proposals_close_at: 4.weeks.from_now
+      starts_at: 6.months.from_now,
+      ends_at: 9.months.from_now,
+      applications_open_at: 3.months.from_now,
+      applications_close_at: 4.months.from_now,
+      acceptance_notification_at: 5.months.from_now,
+      project_proposals_open_at: 4.weeks.ago,
+      project_proposals_close_at: 4.weeks.from_now
     )
   end
 
   def self.fake_application_phase
     season.update(
-       starts_at: 2.months.from_now,
-       ends_at: 5.months.from_now,
-       applications_open_at: 2.weeks.ago,
-       applications_close_at: 2.weeks.from_now,
-       acceptance_notification_at: 6.weeks.from_now
+      starts_at: 2.months.from_now,
+      ends_at: 5.months.from_now,
+      applications_open_at: 2.weeks.ago,
+      applications_close_at: 2.weeks.from_now,
+      acceptance_notification_at: 6.weeks.from_now
     )
   end
 
   def self.back_to_reality
     this_year = Date.today.year
     season.update(
-       name: this_year,
-       starts_at: Time.utc(this_year, *Season::SUMMER_OPEN),
-       ends_at: Time.utc(this_year, *Season::SUMMER_CLOSE),
-       applications_open_at: Time.utc(this_year, *Season::APPL_OPEN),
-       applications_close_at: Time.utc(this_year, *Season::APPL_CLOSE),
-       acceptance_notification_at: Time.utc(this_year, *Season::APPL_LETTER),
-       project_proposals_open_at: Time.utc(this_year-1, *Season::PROJECTS_OPEN),
-       project_proposals_close_at: Time.utc(this_year, *Season::PROJECTS_CLOSE),
+      name: this_year,
+      starts_at: Time.utc(this_year, *Season::SUMMER_OPEN),
+      ends_at: Time.utc(this_year, *Season::SUMMER_CLOSE),
+      applications_open_at: Time.utc(this_year, *Season::APPL_OPEN),
+      applications_close_at: Time.utc(this_year, *Season::APPL_CLOSE),
+      acceptance_notification_at: Time.utc(this_year, *Season::APPL_LETTER),
+      project_proposals_open_at: Time.utc(this_year-1, *Season::PROJECTS_OPEN),
+      project_proposals_close_at: Time.utc(this_year, *Season::PROJECTS_CLOSE),
     )
   end
 end
--- a/app/models/season/phase_switcher.rb
+++ b/app/models/season/phase_switcher.rb
@@ -5,7 +5,7 @@ class Season::PhaseSwitcher
   end
 
   def self.distribute(phase)
-    self.send phases[phase]
+    self.public_send(phase)
   end
 
   def self.fake_coding_phase
--- a/app/models/season/phase_switcher.rb
+++ b/app/models/season/phase_switcher.rb
@@ -4,6 +4,16 @@ class Season::PhaseSwitcher
     Season.current
   end
 
+  def self.switcher(phase)
+    @phases = {
+        'Proposals' => :fake_proposals_phase,
+        'Application' => :fake_application_phase,
+        'CodingSummer' => :fake_coding_phase,
+        'Real Time' => :back_to_reality
+    }
+     self.send @phases[phase]
+  end
+
   def self.fake_coding_phase
     season.update({
        starts_at: 6.weeks.ago,
--- a/app/models/season/phase_switcher.rb
+++ b/app/models/season/phase_switcher.rb
@@ -5,7 +5,7 @@ class Season::PhaseSwitcher
   end
 
   def self.destined(phase)
-    unless phase.to_sym.presence_in(whitelisted_phases)
+    unless phase.to_sym.in?(whitelisted_phases)
       raise ArgumentError
     else
       self.public_send(phase)
--- a/app/models/season/phase_switcher.rb
+++ b/app/models/season/phase_switcher.rb
@@ -5,9 +5,7 @@ class Season::PhaseSwitcher
   end
 
   def self.destined(phase)
-    unless phase.to_sym.in?(whitelisted_phases)
-      raise ArgumentError
-    else
+    if phase.to_sym.in?(whitelisted_phases)
       self.public_send(phase)
     end
   end
@@ -62,7 +60,7 @@ class Season::PhaseSwitcher
   end
 
   def self.whitelisted_phases
-    @phases ||= [
+    @whitelist_phases ||= [
         :fake_proposals_phase,
         :fake_application_phase,
         :fake_coding_phase,
--- a/app/models/season/phase_switcher.rb
+++ b/app/models/season/phase_switcher.rb
@@ -13,8 +13,8 @@ class Season::PhaseSwitcher
     Season.current
   end
 
-  def self.destined(phase: phase)
-    raise ArgumentError.new("#{phase} is not a valid phase") unless phase.in?(PHASES)
+  def self.destined(phase)
+    raise ArgumentError.new("#{phase} is not a valid phase") unless phase.in? PHASES
     self.public_send(phase)
   end
 
--- a/app/models/season/phase_switcher.rb
+++ b/app/models/season/phase_switcher.rb
@@ -4,14 +4,8 @@ class Season::PhaseSwitcher
     Season.current
   end
 
-  def self.switcher(phase)
-    phases = {
-        'Proposals' => :fake_proposals_phase,
-        'Application' => :fake_application_phase,
-        'CodingSummer' => :fake_coding_phase,
-        'Real Time' => :back_to_reality
-    }
-     self.send phases[phase]
+  def self.distribute(phase)
+    self.send phases[phase]
   end
 
   def self.fake_coding_phase
@@ -58,6 +52,17 @@ class Season::PhaseSwitcher
        project_proposals_open_at: Time.utc(this_year-1, *Season::PROJECTS_OPEN),
        project_proposals_close_at: Time.utc(this_year, *Season::PROJECTS_CLOSE),
     })
+  end
+
+  private
+
+  def self.phases
+    @phases ||= {
+        'Proposals' => :fake_proposals_phase,
+        'Application' => :fake_application_phase,
+        'CodingSummer' => :fake_coding_phase,
+        'Real Time' => :back_to_reality
+    }
   end
 
 end
--- a/app/models/season/phase_switcher.rb
+++ b/app/models/season/phase_switcher.rb
@@ -1,4 +1,3 @@
-# The following is not a dev comment but a magic string, with the compliments of Ruby 2.3
 # frozen_string_literal: true
 
 class Season::PhaseSwitcher
--- a/app/models/season/phase_switcher.rb
+++ b/app/models/season/phase_switcher.rb
@@ -4,7 +4,7 @@ class Season::PhaseSwitcher
     Season.current
   end
 
-  def self.distribute(phase)
+  def self.destined(phase)
     unless phase.to_sym.presence_in(whitelisted_phases)
       raise ArgumentError
     else
--- a/app/models/season/phase_switcher.rb
+++ b/app/models/season/phase_switcher.rb
@@ -1,3 +1,6 @@
+# The following is not a dev comment but a magic string, with the compliments of Ruby 2.3
+# frozen_string_literal: true
+
 class Season::PhaseSwitcher
 
   def self.season
@@ -5,7 +8,7 @@ class Season::PhaseSwitcher
   end
 
   def self.destined(phase)
-    if phase.to_sym.in?(whitelisted_phases)
+    if phase.in?(whitelisted_phases)
       self.public_send(phase)
     end
   end
@@ -60,11 +63,11 @@ class Season::PhaseSwitcher
   end
 
   def self.whitelisted_phases
-    @whitelisted_phases ||= [
-        :fake_proposals_phase,
-        :fake_application_phase,
-        :fake_coding_phase,
-        :back_to_reality
+    @whitelisted_phases ||= %w[
+        fake_proposals_phase
+        fake_application_phase
+        fake_coding_phase
+        back_to_reality
     ]
   end
 end
--- a/app/models/season/phase_switcher.rb
+++ b/app/models/season/phase_switcher.rb
@@ -5,13 +5,13 @@ class Season::PhaseSwitcher
   end
 
   def self.switcher(phase)
-    @phases = {
+    phases = {
         'Proposals' => :fake_proposals_phase,
         'Application' => :fake_application_phase,
         'CodingSummer' => :fake_coding_phase,
         'Real Time' => :back_to_reality
     }
-     self.send @phases[phase]
+     self.send phases[phase]
   end
 
   def self.fake_coding_phase
--- a/app/models/season/phase_switcher.rb
+++ b/app/models/season/phase_switcher.rb
@@ -2,31 +2,37 @@
 
 class Season::PhaseSwitcher
 
+  WHITELISTED_PHASES = %i[
+        fake_proposals_phase
+        fake_application_phase
+        fake_coding_phase
+        back_to_reality
+    ]
+
   def self.season
     Season.current
   end
 
-  def self.destined(phase)
-    if phase.in?(whitelisted_phases)
+  def self.destined(phase: phase)
+    if phase.in?(WHITELISTED_PHASES)
       self.public_send(phase)
     end
   end
 
-
   private
 
   def self.fake_coding_phase
-    season.update({
+    season.update(
        starts_at: 6.weeks.ago,
        ends_at: 6.weeks.from_now,
        applications_open_at: 4.months.ago,
        applications_close_at: 3.months.ago,
        acceptance_notification_at: 2.months.ago
-    })
+    )
   end
 
   def self.fake_proposals_phase
-    season.update({
+    season.update(
        starts_at: 6.months.from_now,
        ends_at: 9.months.from_now,
        applications_open_at: 3.months.from_now,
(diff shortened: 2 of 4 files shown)


carpodaster:
Hey Maud, thanks for the PR! Before I get into the review, could you please rebase the branch against the current master? It includes your fix from #584 and should make CI pass again.

F3PiX:
Sorry, I saw the red cross when I did not have a chance to work on it. 
Also, I did not add extra tests yet. I will after your review, if that is okay?

app/models/season/phase_switcher.rb
```diff
@@ -4,6 +4,16 @@ def self.season
     Season.current
   end
 
+  def self.switcher(phase)
+    @phases = {
+        'Proposals' => :fake_proposals_phase,
```
carpodaster: I'm thinking if we should drop the human readable mapping altogether and pass the method names around as strings instead. Reason: the magic values are human readable/understandable, but no one ever sees them. They are hidden in a select box and passed around as POST parameters.

If we decide to use the method names directly, it's imperative to whitelist the values that get passed to `send` / `public_send`. It would be a gaping security hole otherwise.

app/models/season/phase_switcher.rb
```diff
@@ -4,6 +4,16 @@ def self.season
...
+        'Application' => :fake_application_phase,
+        'CodingSummer' => :fake_coding_phase,
+        'Real Time' => :back_to_reality
```
carpodaster: If you want to stick with the "magic string" mapping, `Real Time` must be changed to `RealTime`

app/models/season/phase_switcher.rb
F3PiX: Would it make sense to move this traffic control plus the whitelisting back to the controller?

carpodaster: I wouldn't. Right now, the controller just acts as a YOLO-kind of messenger. It just passed whatever down to whatever owns the actual logic. Let that handle the input validation/sanitation.

app/controllers/orga/seasons_controller.rb
```diff
@@ -45,23 +45,13 @@ def destroy
...
-        Season::PhaseSwitcher.fake_proposals_phase
-      when 'Application'
-        Season::PhaseSwitcher.fake_application_phase
-      when 'CodingSummer'
-        Season::PhaseSwitcher.fake_coding_phase
-      when 'RealTime'
-        Season::PhaseSwitcher.back_to_reality
-      end
-    redirect_to orga_seasons_path, notice: "We time travelled into the #{params[:option]} phase"
+    Season::PhaseSwitcher.destined(params[:phase])
```
F3PiX: I opted for :destined, analogue to Avdi's UserAdder class with :addition method. What I like about :destined/:destination is that is has a 'going somewhere'-connotation. 
Other options: :destination, :chosen, :distribution, :direction 
Now I am totally into Naming Things, working on a blogpost :-)

app/models/season/phase_switcher.rb
```diff
@@ -50,4 +61,12 @@ def self.back_to_reality
     })
   end
 
+  def self.whitelisted_phases
```
F3PiX: Talking about naming things: `whitelisted_phases` or `allowed` phases?

app/views/orga/seasons/_switch.html.erb
```diff
@@ -3,13 +3,13 @@
...

+  <%= radio_button_tag :phase, :fake_coding_phase %>

+  <%= label_tag 'Coding Summer' %>

+  <%= radio_button_tag :phase, :back_to_reality %>
+  <%= label_tag 'Back to Real Time' %>
   <%= submit_tag "Switch phase", class: 'btn btn-sm btn-warning' %>
-<% end %>
+<% end %>
```
F3PiX: So much better, thanks!

app/models/season/phase_switcher.rb
```diff
@@ -4,6 +4,17 @@ def self.season
...
   end
 
+  def self.destined(phase)
+    unless phase.to_sym.in?(whitelisted_phases)
+      raise ArgumentError
+    else
+      self.public_send(phase)
+    end
+  end
+
```
F3PiX: - Can it be that the raising Error line slows down the switching? It looks like it took a whole 4ms more to switch after I added the `raise`.  
- Also, not sure if this is the right Error to raise.
- Before that I had a silent fail - not sending,  and not telling so.
I liked that, because a) in the controller we prevent access to the method in production env and b) if someone tries to do something malicious, a smoke screen serves them right LOL

app/controllers/orga/seasons_controller.rb
```diff
@@ -45,23 +45,13 @@ def destroy
...
+    redirect_to orga_seasons_path, notice: "We time travelled into the #{params[:phase].humanize.titlecase}"
```
carpodaster: I did not know `titlecase` existed. Thanks! :)

app/models/season/phase_switcher.rb
carpodaster: It's more the `unless` condition: converting a String to Symbal and iterating over an Array. Should that take 4ms? I don't know. It's not that we're running at webscale here so it doesn't really matter.

As for the raise: yeah, maybe the silent fail is indeed enough. You'd still need the whitelist check, but don't need to raise anything.

app/models/season/phase_switcher.rb
carpodaster: Since you named the memoized class variable `phases`, I would name the method to match that

spec/models/season/phase_switcher_spec.rb
```diff
@@ -26,9 +25,11 @@
...
 
   describe '#fake_application_phase' do
+
     it 'timeshifts the application phase to today' do
+      phase = :fake_application_phase
       Timecop.travel(Season.current.applications_close_at - 1.month) do
-        Season::PhaseSwitcher.fake_application_phase
+        Season::PhaseSwitcher.destined(phase)
         expect(subject).to be_application_period
       end
```
F3PiX: I thought we want to test the results of the incoming call (from `.destined`, instead of  :fake_*_phase). Is that right?

F3PiX:
@carpodaster Back to you. And thanks again for your explanantions 🙇  .

spec/models/season/phase_switcher_spec.rb
```diff
@@ -4,10 +4,32 @@
...
+    it 'addresses the requested phase' do
+      wrong_phase = :fake_proposals_phase
+      Season::PhaseSwitcher.destined(wrong_phase)
+      expect(subject).to_not be_application_period
+    end
+
+    it 'fails silently when it receives a non-whitelisted phase' do
+      phase = :bad_intentions
+      bad_call = Season::PhaseSwitcher.destined(phase)
+      expect( bad_call ).to be_nil
```
carpodaster: This doesn't test if `.destined` doesn't change things, it merely checks for the _return value_ of that method call. A better way is to make use of `expect { … }.not_to change { … }`. A good thing to check for in the `change` block could be `Season.current.updated_at`.

spec/models/season/phase_switcher_spec.rb
carpodaster: Yes and no. The describe block says you're testing the actual `fake_application_phase`, but that one's now private API. You shouldn't test private methods. It would stick with the changed method call but would update the describe block to a `context` in `describe '#destined' do … end` with something human readable

spec/models/season/phase_switcher_spec.rb
```diff
@@ -4,10 +4,32 @@
...
 
+  describe 'destined' do
+    it 'sends the params to a phase setting method' do
+      phase = :fake_application_phase
+      Season::PhaseSwitcher.destined(phase)
+      expect(subject).to_not be_nil
+    end
+
+    it 'addresses the requested phase' do
+      wrong_phase = :fake_proposals_phase
```
carpodaster: I don't understand this test. Why is it the wrong phase?

Also, see below for making use of `expect { }.to change { … }`.

app/models/season/phase_switcher.rb
```diff
@@ -5,9 +5,7 @@ def self.season
   end
 
   def self.destined(phase)
-    unless phase.to_sym.in?(whitelisted_phases)
-      raise ArgumentError
-    else
+    if phase.to_sym.in?(whitelisted_phases)
```
carpodaster: I just came to realize: `phase` is user input from the web. In older Ruby versions, converting arbitrary user input to symbols using `String#to_sym` could be exploited as Denial of Service attacks (and in some cases even privilege escalation … if memory serves?). I don't know if that is still the case now that symbols are garbage collected, but I would suggest to switch to (frozen) strings.

That is: you can add the Ruby 2.3 frozen string literal magic comment (check the MRI release notes), make `whitelisted_phases` a list of strings instead of symbols and skip the symbol conversion in this line alltogether.

F3PiX:
The season switch specs pass locally. I'll look into the failing check tomorrow.

F3PiX:
Yay! 💚

spec/models/season/phase_switcher_spec.rb
```diff
@@ -54,7 +54,7 @@
     it 'fails silently when it receives a non-whitelisted phase' do
       phase = 'bad_intentions'
       forbidden_call = Season::PhaseSwitcher.destined(phase)
-      expect { forbidden_call }.not_to change { Season.current.updated_at }
+      expect { forbidden_call }.not_to change { Season.current.updated_at.to_s }
     end
```
F3PiX: I got this error on Travis:
```when #doing_this_thing
     Failure/Error: expect { forbidden_call }.not_to change { Season.current.updated_at }
       expected result not to have changed, but did change from 2017-01-19 19:05:27.192892392 +0000 to 2017-01-19 19:05:27.192892000 +0000```
This has something to do with rounding. (Either that, or I am reaaaaally good at pointing the exact millinillisecond a few times )

Comment (carpodaster) [inline spec/models/season/phase_switcher_spec.rb]: I'm not entirely sure if that's really about rounding. The timestamp shouldn't change at all. Furthermore, the `forbidden_call` gets evaluated in line 56 whereas it has to be evaluated within the `expect { }` block. Can you update that one and we'll check the timestamp issue after that?

File: spec/models/season/phase_switcher_spec.rb [hunk 13]
```diff
@@ -53,9 +53,7 @@
 
     it 'fails silently when it receives a non-whitelisted phase' do
       phase = 'bad_intentions'
```
carpodaster: Could you add these two lines right before the `expect {} ` block?

```ruby
# Just to make sure we have our season in the test DB.
# It creates one if it didn't already exist though some other factory:
Season.current 
allow_any_instance_of(Season).to receive(:save).and_raise RuntimeError
```

If the test outputs a RuntimeError, something is updating the season where it shouldn't get an update. If the date error prevails, it's the rounding.

F3PiX:
Huh??  Now it passes?

carpodaster:
@F3PiX Yeah, that makes no sense at all. Will have to check in more detail – but not tonight I'm afraid :zzz:

F3PiX:
Thanks for your help! Very interesting, as always :-) 🙏

F3PiX:
I want to try some things, I'll ping you when I am ready, okay?

F3PiX:
Okay, I think I found why it _is_ the timestamps, I wrote it down here: https://medium.com/@mauddev/about-rspec-timestamps-and-postgresql-7bbb0e020a7b

And about the method stub: I tested it in the all the other switch specs. It works everywhere except in the last two. Rspec [says about stubbing the method:](https://www.relishapp.com/rspec/rspec-mocks/v/2-14/docs/method-stubs/stub-on-any-instance-of-a-class) " Nothing happens if the message is never received." And that is exactly what the silent fail does, isn't it? Redirecting without sending the message.  

Mystery solved?

F3PiX:
@carpodaster I *think* I solved the mystery ^. What do you think?

F3PiX:
Ping @klappradla

app/models/season/phase_switcher.rb
klappradla: Just outta curiosity @F3PiX, why did you chose a memorized class method over a constant?
I'd intuitively go with
```ruby
PHASES = %i(fake_proposal_phase ...)
```
Also I prefer symbols over strings 😉

app/models/season/phase_switcher.rb
```diff
@@ -1,9 +1,20 @@
+# frozen_string_literal: true
+
 class Season::PhaseSwitcher
 
   def self.season
     Season.current
   end
 
+  def self.destined(phase)
+    if phase.in?(whitelisted_phases)
```
klappradla: This is only personal preference, but rather then doing nothing on invalid input, I'd raise an error instead. I think this is more helpful in certain cases
```ruby
def self.destined(phase)
  raise ArgumentError.new("#{phase} is not a valid phase switch") unless whitelisted_phases.include?(phase)
  publis_send(phase)
end
```

app/models/season/phase_switcher.rb
```diff
@@ -1,9 +1,20 @@
...
+      self.public_send(phase)
+    end
+  end
+
```
klappradla: Again personal preference, but these double blank lines before `private` are kinda oldskool and hurt my eye 😉  - why not just a single blank line before and after?

app/models/season/phase_switcher.rb
```diff
@@ -1,9 +1,20 @@
...
+
+  private
+
   def self.fake_coding_phase
     season.update({
```
klappradla: Since we're on Ruby > 2.some, I'd vote for using *"real"* KWARGS everywhere. Plus, since it's only one param, this can also savely be done in one line I'd say:
```ruby
season.update(starts_at: 6.weeks.ago)
```

app/controllers/orga/seasons_controller.rb
klappradla: I'm not 100% sure, but doesn't the `phase` end up in the `params` as `string` **always**? So depending on your contract with `PhaseSwitcher` I'd suggest to do the necessary conversion **before** passing it on to the business logic. Could e.g. look like this:

```ruby
def switch_phase
  # ...
  Season::PhaseSwitcher.destined(phase)
end

private

def phase
  params[:phase].to_sym
end
```

spec/models/season/phase_switcher_spec.rb
```diff
@@ -4,32 +4,61 @@
...
   subject { Season.current }
 
-  describe '#fake_application_phase' do
-    it 'timeshifts the application phase to today' do
-      Timecop.travel(Season.current.applications_close_at - 1.month) do
-        Season::PhaseSwitcher.fake_application_phase
-        expect(subject).to be_application_period
+  describe 'destined' do
+
+    it 'sends the params to a phase setting method' do
```
klappradla: Sorry, I don't get what this spec is supposed to do. Why does calling the `PhaseSwitcher` have an impact on weather there is a current season or not?

spec/models/season/phase_switcher_spec.rb
```diff
@@ -4,32 +4,61 @@
...
       end
     end
+
+    it 'addresses the requested phase' do
+      another_phase = 'fake_proposals_phase'
+
+      Season::PhaseSwitcher.destined(another_phase)
+      # Just checking, here too the error is not raised
+      Season.current
+      allow_any_instance_of(Season).to receive(:save).and_raise RuntimeError
```
klappradla: Maybe I just don't understand the behavior of `Season.current`... But what is this line for? When should there be another call to `some_season.save` when calling `Season.current` again *(as in calling the subject)*?

spec/models/season/phase_switcher_spec.rb
```diff
@@ -4,32 +4,61 @@
...
 
-  describe '#fake_coding_phase' do
-    it 'timeshifts the coding phase to today' do
-      Timecop.travel(Season.current.ends_at - 1.week) do
-        Season::PhaseSwitcher.fake_coding_phase
-        expect(subject).to be_started
+    context 'when directed to the #fake_coding_phase' do
+      it 'sets dates so that the summer of code is currently happening' do
+        phase = 'fake_coding_phase'
+        Timecop.travel(Season.current.ends_at - 1.week) do
```
klappradla: Why is it necessary to do timetravel here *(this also applies to the following specs)*? The way I understand it, the underlying methods just update the certain values for the current `season`. So in theory it should be enough to do something along the lines of
```ruby
expect(Season)
  .to receive(:current)
  .and_return(season)
expect(season)
  .to receive(:update)
  .with(# all the things)
```
shouldn't it?

Besides that, I don't really understand why `Season.current` is the `subject` of your specs here. Isn't the thing that's being tested the `PhaseSwitcher`? So in my opinion, the `subject` should actually be
```ruby
described_class.destined(phase)
```
then you can of course also expect the changes on a season record with this. But I think this is outta scope for this test - the `Season` is already tested. I'd simply test if the respective methods get called on the season object.

spec/models/season/phase_switcher_spec.rb
```diff
@@ -4,32 +4,61 @@
...
+      another_phase = 'fake_proposals_phase'
+
+      Season::PhaseSwitcher.destined(another_phase)
+      # Just checking, here too the error is not raised
+      Season.current
+      allow_any_instance_of(Season).to receive(:save).and_raise RuntimeError
+      expect(subject).to_not be_application_period
+    end
+
+    it 'fails silently when it receives a non-whitelisted phase' do
```
klappradla: If you go with the `ArgumentError` approach, this needs to be updated as well of course 😉

F3PiX:
Thanks a lot, @klappradla!!! 
There is one more unanswered question: about the Travis build failing. See [this comment](https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/585/commits/b7c09e3688ef4a67bb8b6a268c2960f8a6692977#r97083178) for the error message. And [this one ](https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/585#issuecomment-275074458) where I _think_ I solved it. (That would be the subject of my blog post, BTW).

app/models/season/phase_switcher.rb
F3PiX: The silent failing is intentional, to diss mailicious input-ters. Do I miss a use case where it would be helpful? Can you think of one? See https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/585#discussion_r95774916

app/controllers/orga/seasons_controller.rb
F3PiX: I like this very much... but we skipped the magic strings in a previous review. So the phase-param is always a symbol. No conversion needed. Right?

spec/models/season/phase_switcher_spec.rb
F3PiX: This was a temporarily check because of the failing build. TBH I just copied Carsten's prompt. 😳

spec/models/season/phase_switcher_spec.rb
F3PiX: I am very bad at writing tests and writing Rspec, sorry. IIRC I _intended_ to test the interface only, and now I understand I did not. 
So, I tried your suggestions and I don't understand what to do. The `expect and receive` things looks to me as it is exactly what we need, I didn't know about that. But I got the strangest errors, never got past the first part where the Season is expected to receive :current. So, maybe you can give me hint in the right direction?
Or I can show a few things I tried? Whatever is easier for you. 
So, I moved to part 2, because it makes totally sense that the season should not be the subject. Here too: all kind of errors. I did a sanity check: 
```
it "is available as described_class" do
expect(described_class).to eq(PhaseSwitcher)
end
```
and that failed. 🤔  
Tips?

app/controllers/orga/seasons_controller.rb
klappradla: Hm, I need to try this out in real code tomorrow. I don't see how Rails could be able to distinguish between symbols and strings in the params hash. It's parsed out of the HTTP request and I cannot imagine how this should work then... But I'll try it out.

michaelem:
@F3PiX I just restarted the build and it went green. I think my recent ajaxification of the comments controller (https://github.com/rails-girls-summer-of-code/rgsoc-teams/commit/468100ab410be6086e8f0b527960a941e39387c9) is to blame here, although I'm not sure yet what exactly causes the test to flake.

app/controllers/orga/seasons_controller.rb
F3PiX: Aha, no, you are right ofcourse. I turned things topsy turvy. I'll update accordingly.

klappradla:
And for the _open question_ @F3PiX: I actually don't think that the microseconds can become a problem here. To me this looks more like the object has not been written to the DB before (e.g. via `let!()`). Aside that, if you just want to check weather or not the record has been updated, I'd go with something like this
```ruby
expect { ClassMethod.called }.not_to change { my_object.reload }
```
checking an already instantiated object won't show the changes.

spec/models/season/phase_switcher_spec.rb
klappradla: Ok. So I'm having a look at the respective files locally.

First of all, in order to be able to use `described_class`, the spec needs to be setup as a proper (model) spec. So start the file with:
```ruby
require 'spec_helper'

describe Season::PhaseSwitcher do
  context 'when switching phases' do
    # ...
```

Once you've done this, the expectation needs to include the `Season` module as well:
```ruby
it "is available as described_class" do
  expect(described_class).to eq(Season::PhaseSwitcher)
end
```
This is necessary because you're out of the `Season` "namespace" in the spec file. You could of course go into it by setting this up like this:
```ruby
require 'spec_helper'

module Season
  describe PhaseSwitcher do
# ...
```
However, I **strongly** advise not to do it! `Season` is not a module but a class and the `PhaseSwitcher` class not set up to be within the `Season` namespace. If it would, it would look like this:
```ruby
module Season
  class PhaseSwitcher
```
However, with Rails 💩 autoloading stuff, you'll run into troubles with this anyways 😉 

Once you got the spec files, namespaces, etc. sorted out, let me know how I can help with the tests themselves.

app/models/season/phase_switcher.rb
```diff
@@ -1,44 +1,61 @@
+# frozen_string_literal: true
+
 class Season::PhaseSwitcher
 
+  WHITELISTED_PHASES = %i[
```
klappradla: 👍  but please switch to 2-spaces wide soft tabs - the 4 spaces thing is Python only 😉 
It's just a Ruby Community best practise to go with 2 spaces.

Plus: I prefer the `i%(` over the `i[[` syntax, since it makes it more obvious that what's being invoked here is actually a function... Aside from that, it's also handy to freeze the constant afterwards:
```ruby
PHASES = %i(
    fake_proposals_phase
    fake_application_phase
    fake_coding_phase
    back_to_reality
).freeze
```

And on naming: I think it's perfectly fine to reduce this to `PHASES`, no need to prepend "whitelist" if there no such thing as general `PHASES` or "blacklisted" phases there.

app/models/season/phase_switcher.rb
```diff
@@ -1,44 +1,61 @@
...
+        fake_coding_phase
+        back_to_reality
+    ]
+
   def self.season
     Season.current
   end
 
+  def self.destined(phase: phase)
+    if phase.in?(WHITELISTED_PHASES)
```
klappradla: May slightly change if we opt into the raise-error-appraoch. But for now, again, just a community best practise, I'd vote for doing this in a single line instead, makes it a bit more expressive:
```ruby
self.public_send(phase) if phase.in?(PHASES)
```

app/models/season/phase_switcher.rb
klappradla: Ok, get it. However, it's a dev-env-only feature anyways, so if someone is trying to shoot down their server with invalid params... their problem. To me "silent failing" sounds like hiding potential problems and increasing technical dept. Therefore I'd vote for using the error approach, it's simply quicker feedback that there's something wrong.

app/controllers/orga/seasons_controller.rb
F3PiX: While I was changing this, I remembered [this conversation](https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/585#discussion_r96146778) about the DoS risk with converting user input strings to symbols (and the magic . Aren't we reintroducing the risk if we convert upfront?

F3PiX: ^
http://brakemanscanner.org/docs/warning_types/denial_of_service/

spec/models/season/phase_switcher_spec.rb
F3PiX: That made all the difference. 🙈 
Second try,  to only test the public interface. To understand what is happening, I needed to work on a lower abstraction level than your instructions. I can change that later, unless you are tired of it...
The `expect(Season).to receive(:current).and_return(season)`did not want to return `season`, so I just added a var. 
What I tried to accomplish: see if the season receives the call with the phase, by checking roughly if something in the output changed. 
And now I realised why I added the Timecop: when the test run during the same phase as the fake-phase. I solved it not by adding a test in another phase. 
wdyt?

app/controllers/orga/seasons_controller.rb
klappradla: Well, it's a dev-env-only feature... But anyways, Ruby 2.2 introduced GC for symbols (cc @carpodaster), so as long as you don't do a
```ruby
define_method(string.to_sym)
```
DoS from converting any strings to symbols should not be a problem.

However, if you prefer strings, feel free to switch to them. I just proposed symbols because I personally prefer them over strings. But the whole logic works with strings as good as with symbols (just need to switch `%i()` to `%w()` then).

app/controllers/orga/seasons_controller.rb
```diff
@@ -82,6 +72,10 @@ def season_params
     )
   end
 
+  def phase
+    params[:phase].to_sym if phase.in? PHASES
```
klappradla: As I said in a previous comment, since I think we can be sure to **always** be on a Ruby version > 2.2, we'll have GC for symbols. Therefore, no problem with DoS when converting strings to symbols in this case.

I'd vote for removing the `if` condition here. I also cannot imagine it working *(`PHASES` is defined in a different namespace, would be `Season::PhaseSwitcher::PHASES`)*... plus I don't think it makes huge sense to first check wheather the param is whitelisted, if not, pass on nil and then check again if it's whitelisted.

So, I'd suggest to just go with:
```ruby
def phase
  params[:phase].to_sym
end
```
and then to the whole whitelist stuff in the model where the actual business logic lies.

app/models/season/phase_switcher.rb
```diff
@@ -1,44 +1,60 @@
+# frozen_string_literal: true
+
 class Season::PhaseSwitcher
 
+  PHASES = %i(
+    fake_proposals_phase
+    fake_application_phase
+    fake_coding_phase
+    back_to_reality
+    ).freeze
```
klappradla: Almost :v: (sorry for being so picky, but since it's so close...): put this line on the same level as `PHASES` and it looks adorable 😻

app/models/season/phase_switcher.rb
```diff
@@ -1,44 +1,60 @@
...
   def self.season
     Season.current
   end
 
+  def self.destined(phase)
+    raise ArgumentError.new("#{phase} is not a valid phase") unless phase.in? PHASES
+    self.public_send(phase)
+  end
+
+  private
```
klappradla: Since you're only defining methods on `self` here and use `public_send` to invoke them, `private` does not have any effect here. I'd vote to either really make things private or just remove it in order to not irritate future developers.

If you decide to make things private and only expose the `destined` method to the outer world *(which would make sense, since it's the only method that get's tested and it's the only thing invoked from outside)*, you need to "open up" the class. This looks like this:
```ruby
class Season::PhaseSwitcher
  PHASES = %i(
    fake_proposals_phase
    fake_application_phase
    fake_coding_phase
    back_to_reality
  ).freeze

  class << self
    def destined(phase)
      raise ArgumentError.new("#{phase} is not a valid phase") unless phase.in? PHASES
      self.send(phase)
    end

    private

    def season
      Season.current
    end

    def fake_coding_phase
      season.update(
        starts_at: 6.weeks.ago,
        ends_at: 6.weeks.from_now,
        applications_open_at: 4.months.ago,
        applications_close_at: 3.months.ago,
        acceptance_notification_at: 2.months.ago
      )
    end

    def fake_proposals_phase
      season.update(
        starts_at: 6.months.from_now,
        ends_at: 9.months.from_now,
        applications_open_at: 3.months.from_now,
        applications_close_at: 4.months.from_now,
        acceptance_notification_at: 5.months.from_now,
        project_proposals_open_at: 4.weeks.ago,
        project_proposals_close_at: 4.weeks.from_now
      )
    end

    def fake_application_phase
      season.update(
        starts_at: 2.months.from_now,
        ends_at: 5.months.from_now,
        applications_open_at: 2.weeks.ago,
        applications_close_at: 2.weeks.from_now,
        acceptance_notification_at: 6.weeks.from_now
      )
    end

    def back_to_reality
      this_year = Date.today.year
      season.update(
        name: this_year,
        starts_at: Time.utc(this_year, *Season::SUMMER_OPEN),
        ends_at: Time.utc(this_year, *Season::SUMMER_CLOSE),
        applications_open_at: Time.utc(this_year, *Season::APPL_OPEN),
        applications_close_at: Time.utc(this_year, *Season::APPL_CLOSE),
        acceptance_notification_at: Time.utc(this_year, *Season::APPL_LETTER),
        project_proposals_open_at: Time.utc(this_year-1, *Season::PROJECTS_OPEN),
        project_proposals_close_at: Time.utc(this_year, *Season::PROJECTS_CLOSE),
      )
    end
  end
end
```

app/models/season/phase_switcher.rb
```diff
@@ -1,44 +1,60 @@
...
+
   def self.fake_coding_phase
-    season.update({
+    season.update(
        starts_at: 6.weeks.ago,
```
klappradla: All these lines are intended one space too deep 😉

app/models/season/phase_switcher.rb
```diff
@@ -1,44 +1,60 @@
...
        applications_close_at: 3.months.ago,
        acceptance_notification_at: 2.months.ago
-    })
+    )
   end
 
   def self.fake_proposals_phase
-    season.update({
+    season.update(
        starts_at: 6.months.from_now,
```
klappradla: All these lines are intended one space too deep 😉

app/models/season/phase_switcher.rb
```diff
@@ -1,44 +1,60 @@
...
        project_proposals_open_at: 4.weeks.ago,
        project_proposals_close_at: 4.weeks.from_now
-    })
+    )
   end
 
   def self.fake_application_phase
-    season.update({
+    season.update(
        starts_at: 2.months.from_now,
```
klappradla: All these lines are intended one space too deep 😉

app/models/season/phase_switcher.rb
```diff
@@ -1,44 +1,60 @@
...
        acceptance_notification_at: 6.weeks.from_now
-    })
+    )
   end
 
   def self.back_to_reality
     this_year = Date.today.year
-    season.update({
+    season.update(
        name: this_year,
```
klappradla: All these lines are intended one space too deep 😉

spec/models/season/phase_switcher_spec.rb
```diff
@@ -1,35 +1,43 @@
 require 'spec_helper'
 
+describe Season::PhaseSwitcher do
 
-context 'when switching phases' do
-  subject { Season.current }
+  it "is available as described class" do
```
klappradla: Since this is now sorted out, please remove this again, no need to test rspec's internals.

spec/models/season/phase_switcher_spec.rb
```diff
@@ -1,35 +1,43 @@
...
+    expect(described_class).to eq Season::PhaseSwitcher
+  end
+
+  context 'when switching phases' do
+    season = Season.current
```
klappradla: Outside of examples, it's best practice and way more understandable for every other developer if you'd use `let` here:
```ruby
let(:season) { Season.current }
let!(:season) { Season.current } # if you need to have it called *before* the first call to "season"
```

spec/models/season/phase_switcher_spec.rb
```diff
@@ -1,35 +1,43 @@
...
 
+describe Season::PhaseSwitcher do
 
-context 'when switching phases' do
-  subject { Season.current }
+  it "is available as described class" do
+    expect(described_class).to eq Season::PhaseSwitcher
+  end
+
+  context 'when switching phases' do
```
klappradla: Since what's happening here is basically testing a `PhaseSwitcher`'s class method, I'd suggest to change this line to:
```ruby
describe '.destined' do
```
this makes things more obvious for people who read the test.

spec/models/season/phase_switcher_spec.rb
```diff
@@ -1,35 +1,43 @@
...
-        expect(subject.project_proposals_close_at).to be > fake_time
+    context 'when the input is malicious' do
+      phase = :bad_intentions
+
+      it 'raises an error when it receives a non-whitelisted phase' do
+        expect {
+          described_class.destined(phase)
+        }.to raise_error(ArgumentError)
+      end
+      it 'does not change the season dates' do
```
klappradla: I'm not a huge fan of this `rescue` pattern. How about using a chained expectation instead?
Should look something along the lines of this if I recall correctly:
```ruby
RSpec::Matchers.define_negated_matcher :not_change, :change

expect { described_class.destined(phase) }
  .to raise_error(ArgumentError)
  .and not_change { Season.current.reload }
```

spec/models/season/phase_switcher_spec.rb
klappradla: Yes, makes sense. Thanx for explaining!

app/models/season/phase_switcher.rb
F3PiX: 😂

app/models/season/phase_switcher.rb
F3PiX: Junior caught in the cross fire 🎆  🚒  🤣 
Quote: 
> It's better to use public_send. Object#send will also call private methods which is ok (but arguably still dangerous) in certain meta-programming contexts. We want to enforce the public API here.

F3PiX: I really care about keeping `self.` instead of `class << self` 😬

F3PiX:
Thanks again Max for your help. Glad that the tests are now testing what they should test. Phew.

app/models/season/phase_switcher.rb
klappradla: Alright. Just out of curiosity, why do you want to have all methods on `self`? I could also have imagined this class being a service object or a "fake singleton" using `class << self`.

klappradla:
Good to 🚢  I'd say @F3PiX

Merging now will break the build because the deploy hook is currently not working, but I can deploy it by hand for you.

app/models/season/phase_switcher.rb
F3PiX: Then maybe I am wrong. I always thought that they were the same, just a matter of personal preference. I also thought this were a service object 🙈.
Well, I have homework to do. 
Lots of lesson learned by refactoring away a case statement ;-)

carpodaster:
🎊

app/models/season/phase_switcher.rb
klappradla: It's a classic. Katrina Owen has given whole talks about refactoring a case statement :v:

And yes, for public class methods it's effectively the same. However, if you want to have private ones, only using `private` is not enough, if I recall correctly, there would be some macros like `def_private_method` or some like this...
Therefore when I want to have private methods (basically all of the time) I use `class << self`.

And for service objects: they can be anything as long as you call them "service object". Oftentimes, in a Rails context, the may for instance look like this:
```ruby
class Service
  def initialize(val:)
    @val = val
  end

  def call
    perform_complex_logic
  end

  private

  attr_reader :val

  def perform_complex_logic
    # do stuff
  end
end
```

But they can essentially just be anything that suits your business logic needs 😉

carpodaster: Boom! It never occured to me that `attr_reader` can be made private as well `/mindblown` 🙇

PR #589, opened 2017-01-06
Avoid orphaned attendances

### LINKED ISSUE #586: New bug with the conferences....
The fix in #584 causes a new error locally (`NoMethodError at /users/34
undefined method 'name' for nil:NilClass`) ,  in the user/show view where the conferences are called: 
https://github.com/rails-girls-summer-of-code/rgsoc-teams/blob/master/app/views/users/show.html.slim#L53

It can be fixed by changing `@user.attendances.includes(:conference)` into `@user.attendances.joins(:conference)`  
I can submit this fix as well.
Or do we need another approach?

carpodaster:
Fix #586 – the bug described in the issue must have occurred due to a deleted conference.

This PR will delete attendances alongside its conference. It is possible that the attendance records were intentionally allowed to linger so that we may know of a student's attendance even after its corresponding conference record was deleted. I do not think we need it - knowing that there was some kind of conference attendance without actually knowing _which_ conference exactly hardly provides value. What do you think?

The PR leaves existing attendance orphans intact but avoids the error described in the #586 using the fix @F3PiX suggested.

### DIFF
--- a/app/models/conference.rb
+++ b/app/models/conference.rb
@@ -2,7 +2,7 @@ class Conference < ActiveRecord::Base
 
   include HasSeason
 
-  has_many :attendances
+  has_many :attendances, dependent: :destroy
   has_many :attendees, through: :attendances, source: :user
   validates :name, :round, presence: true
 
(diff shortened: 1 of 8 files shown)


alicetragedy:
+1 on the reasoning behind this. no point in having attendances stick around if the conference is removed, I think.

carpodaster:
Very valid points, Maud, thank you!

I've added both a `dependent: :destroy` for a user's attendances and a migration that sweeps all orphaned attendances in the system.

PR #590, opened 2017-01-06
Fix ninja commas or missing whitespace in conferences links … once and for all?

carpodaster:
Some experiments as part of my investigation for #573 led to this PR. I think it supersedes your PR, @F3PiX which is why I would appreciate if you could have a look at this.

I removed the slight difference in displaying conference attendances; they all use the helper logic now. I've also added a bucketload of test examples that will hopefully silence conference formatting error for good now :smile:

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -140,7 +140,7 @@ module ApplicationHelper
     conferences.map do |conference|
       text = conference.name
       if verbose
-        extra = [conference.location, format_conference_date(conference.starts_on, conference.ends_on)].reject(&:blank?).join(', ')
+        extra = [conference.location, format_conference_date(conference.starts_on, conference.ends_on)].reject(&:blank?).join(' – ')
         text += " (#{extra})" unless extra.blank?
       end
       link_to(text, conference)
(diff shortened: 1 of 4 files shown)


F3PiX:
Great that they all share the same helper method! I am not convinced about the new format and omitting the comma's entirely. Maybe you can add a screenshot to show how it looks? 

Anyway, if the next step would be to move everything over to a conference helper class, like [you suggested](https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/586#issuecomment-270620652), then we can reevaluate the formatting when everything is nice in place. Right?

carpodaster:
Should have included screenies right away, apologies! 

I replaced the comma with a `–` so that there aren't two commas for a location in the form of `City, Country`.

The main difference is that the entire string is a link now. 

### Before

users-show-before

---

### After
users-show-after

F3PiX:
Cool, that looks so much sharper! 
The comma/dash thing works when there is no conference location?

The link_to_everything is what I tried to avoid with my solution. I think it requires an extra step of refactoring to get the 'verbose details' out of the link. I would cheer for that, and leave it up to you :-)

carpodaster:
> The comma/dash thing works when there is no conference location?

[It does](https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/590/files#diff-31c50b16a4c9f23be71c49d30afcf7edR134) :smile: 

I'm not too fuzzed about the extra info being part of the link, at least it's all part of the same helper now and we have tests for it. It's easier to change now.

Thanks for your feedback!

Issue #366, opened 2016-02-03
Application process: normalize location fields with Google auto-complete


carpodaster:
We request people to answer »Where do you plan to live July-September?« as _City, Country_ which doesn't work so well in all cases. Google provides an autocompleter for location data that could remove some of the randomness in the entered data.

carpodaster:
If we add latitude / longitude fields to this we might even calculate the distance to a conference site. But this might be overdoing it?

izuroxx:
hey @carpodaster ! I've been looking at this issue - can i try to do this?

carpodaster:
Morning @izuroxx – but of course!

mkalininait:
I've just looked at how it is implemented in the User Profile: there are two fields Country and City, the first is a dropdown and the latter is a simple input field. Does it mean that we will have different implementations of location input on the Application Form and in the User Profile?

Another thing to think about is the [Additional Info form](https://github.com/rails-girls-summer-of-code/selection_process/issues/35). We probably want to enter location the same way there (and even be filled out automatically), don't we?

klappradla:
I'm taking a look at it. Feels kinda weird to write JS after such a long time 🙈 

Just one question: is there already a RGSoC project with a Google (Maps) API key @carpodaster ?

carpodaster:
@klappradla I don't know of any Google Maps keys, no. Not in front of a computer right now, but maybe `heroku config` has something?

klappradla:
Ah, good idea! Gonna look there.

I'm for now just playing around with the API. However, I guess it's not suggested to publish the API key in a repo *(especially given the fact that there are rate limits for the free plan)*. Still, this would again make the dev setup more complex 🙈

F3PiX:
I think the advantages of autocomplete address input are well worth the extra dev setup. I am a big fan of this feature. 💃

klappradla:
Hm, what I kinda don't like about it: you have to deliberately chose one of Google's suggestions. But ok, maybe that's what people are used to from using GoogleMaps anyways...

![autocomplete](https://cloud.githubusercontent.com/assets/3491815/21728760/28102f2a-d448-11e6-980e-dd3e467aa1db.gif)

alicetragedy:
I'm definitely working on setting up an account that we can link to the API key, will get that done together with @klappradla once I've talked to Anika about this. Also, awesome work and thanksssss!!! ✨

mkalininait:
[...]

PR #596, opened 2017-01-08
Some cosmic change / legacy things

klappradla:
Hey @carpodaster,

I just came across a few pre Rails5 era and therefore eventually no longer necessary things while looking into #366. Just thought, I'd share them in a separate PR, feel free to discard all of this if you where already aware of these things or if they just don't matter.

Tiny bit of background: on the way from Rails 3 to 4, they got rid of the `secret_token` initializer - I just remembered it was part of the [upgrade guide](http://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#config-secrets-yml). The Teams app already uses the new preferred way - the `secrets.yml` file, so the initializer is obsolete, all env vars are already set on Heroku.
*This means one could then also remove the `RAILS_SESSION_SECRET` from the Heroku config.*

Since January '15 Heroku also [recommends Puma](https://devcenter.heroku.com/changelog-items/594) as a webserver, due to some [severe problems with Unicorn](https://devcenter.heroku.com/articles/deploying-rails-applications-with-the-puma-web-server#slow-clients). I just added there default Puma config, it's approx. the same as the unicorn config before. Plus: I don't see a reason why not to use it in development as well.

As I said, feel free to discard these commits, I just wanted to not directly forget about these things when I saw them :v:


carpodaster:
Superb, thanks @klappradla!

PR #597, opened 2017-01-09
Clearify beginner wording

bitboxer:
When setup up the app this morning I had a little problem understanding a small part of the readme because of layout issues. I fixed those. And I removed a hard to understand word 👍

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,8 +132,10 @@ E.g. `foreman run rails server` or `foreman run rails console`.
       Instead of using the last created User, you can also search for a
       specific User with ```user = User.find_by(github_handle: "yourgithubhandle")```.
 
-      You can assign yourself other roles in the same way. If however you assign
-      yourself a student role AND another role, that may lead to unexpected behavior in the app. In that case, remove the student role. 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -115,19 +115,26 @@ directory the variables from `.env` will be loaded into the environment.
...

+    * In Rails Console:   
+      ```
+      user = User.last
+      user.roles.create(name: "organizer")
+      ```
+      Instead of using the last created User, you can also search for a
+      specific User with ```user = User.find_by(github_handle: "yourgithubhandle")```.
+
+      You can assign yourself other roles in the same way. If however you assign
```
F3PiX: 'the same way' is not so clear, with the sequence changed. Maybe something as : the same way as creating the organizer role.

F3PiX:
Nice!!!

README.md
bitboxer: fixed

carpodaster:
Merci! Documentation feedback and fixes are so incredibly valuable!

alicetragedy:
🎉

Issue #184, opened 2015-04-06
DRY up student/user fields


carpodaster:
The templates `app/views/users/edit.html.slim` and `app/views/application_drafts/_student_fields.html.slim` duplicate form fields. Let's find a way to [DRY](https://de.wikipedia.org/wiki/Don%E2%80%99t_repeat_yourself) things.

F3PiX:
Can I fav items I would like to work on in de RGSoC?

carpodaster:
You can't fav issues I don't think, only the repo. Just provide the link to the issue / the issue no. in the project plan, that's the best way to reference it. Thanks! :ok_hand: :smile:

carpodaster:
Things get a lot more interesting (read: difficult) with PR #422

katrin-k:
I'd like to look into this one. However, I'm not 100% sure, I get the issue completely. I see the duplicate fields and my 1st thought was: Hmm, why not remove the "Application-specific info" in the user profile? (And double-check, everything is then part of the application) But I suppose there is something preventing this?

carpodaster:
The application-specific fields are stored on the user record so that they always "belong" to a certain student. The knack with the application draft is: it's not too easy to be sure who's `student0` and who's `student1` when a student accesses the application draft form. The `ApplicationDraft#students` method attempts to make this deterministic, but I don't know what happens if a student changes their pair in the existing team – were the `studentX_*` fields stored on the ApplicationDraft record, this could result in possible privacy disclosure.

It's a difficult topic – I hope I could shine at least some light onto this? 🔦 

If anyone has a good solution for not storing the application specific data on the user record, please speak up 😄 For the time being I think it's best to DRY the form fields w/o touching the database columns.

F3PiX:
With all the action on the Application 2017, it would be really helpfull to solve this issue. I already encountered duplication issues with the very first item (#601). 
Not hindered by too much insight about the implications: can't we just add a shared form partial where we pass in the right student as a local var?

carpodaster:
I was thinking to remove it from `users#edit` alltogether – at least for the time being.

I _think_ my idea of including it there was so that applicants can already start filling out their info while they are still looking for a team mate. Is this is a use case? I don't know. For the sake of simplicity, I am willing to let that doubious feature fall.

F3PiX:
That is a very plausible use case! I think it is weird that you have to pair up before you can see the form at all.  The form contains useful hints about what is expected from a student. 
[...]

PR #604, opened 2017-01-10
Remove duplicate appl fields

F3PiX:
Solves #184 (at least, temporarily ;-) )

- [x] Delete application fields from User Profile Page
- [x] Delete hint in Application Field to User Profile Page
- [x] Check guides for references to the possibility of User Info entry on Profile Page. -> Found none.


carpodaster:
Thanks Maud!

PR #603, opened 2017-01-10
[#583] Application Form: add a community involvement section

bitboxer:
closes #583

PR #605, opened 2017-01-10
Application Form: Add character limits on application inputs

### LINKED ISSUE #578: Application Form: Set character limit to some fields
There are a few fields in the application form that are a bit too prone to applicants writing long texts, which we wish to avoid. A way to solve this would be to set a character limit on those fields.

**Implementation**
- The character limit should be set at the view level in the form
- The length should *probably* be validated at the model level as well (that would be `user` and `application_draft`)
- Each input field should contain a hint about the number of characters allowed
- *Bonus / nice to have*: character count going down as the user types (i.e. like you have on twitter) 

**The fields that need a character limit are the following:**

"Please tell us a little bit about yourself. What is your background?" (1000 characters)
"How did you get into programming?" (1000)
"Please give us a short summary of your programming skills" (1000)
"Tell us about your community involvement" (1000)
"Please summarize what you have been doing to learn programming since your first workshop" (1000)
"What do you want to achieve by the end of the summer?" (2000)
"Why are you applying to this program, and what do you think you can give back?" (2000)

carpodaster:
Fixes #578

It is currently deployed on [staging](https://rgsoc-teams-staging.herokuapp.com/) and can be tried out.

I did not go for the HTML `maxlength` attribute; I thought it would cause negative side-effects (ie. trunking) when pasting text from another source. The counter JS indicates that the limit has been exceeded. The draft can be saved, but the application can't be submitted.

The character limit is 2000 now which I find is not too much. It's _characters_ after all, not words.

Should the new "giving back" field (as introduced in #603) also be character-limited?


mkalininait:
@carpodaster Thank you Carsten! It looks good! I didn't try to submit the form yet though.

Re the "giving back" field: yes, I think it should be character-limited, and also look like a multiline textarea (like the other 'essay' fields).

F3PiX:
Wow, I never knew how much code was involved for this feature :-)
On staging, I find my own application form for last year. When I try to add and save text  > limit, it doesn't save, AND it doesn't tell me why. (I was only able to test in my own profile, where I can't save any of the changes, so maybe it is not related to the number of characters limit.)   
And in the Preview tab, I can't do the 'ready to submit check' either. I get a warning that I have to save my draft (which I tried) and that my draft was not ready to submit.

spec/inputs/character_limited_input_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+        'data-maxlength' => Student::CHARACTER_LIMIT,
+        'class'          => 'form-control character_limited required',
+        'name'           => 'foo[attr_name]'
+      }
+
+      expect(subject).to have_tag('textarea', with: tag_attributes)
+    end
+
+    it 'does not use the html maxlength attribute' do
+      expect(subject).not_to include 'maxlengh='
```
F3PiX: typo: maxlength

carpodaster:
@mkalininait "giving back" is a character-limited field now. In order for the styling to work, the _"x characters left"_ needs a hint/help-block (sorry, I don't know better CSS-wise). I had to come up with a hint for "giving back". Suggestions for a better one highly welcome ;)

@F3PiX that is a weird thing you report there. You shouldn't be able to see your past application draft at all. I also got some errors on staging stating `ActiveRecord::RecordInvalid: Validation failed: Github handle has already been taken`. Would you mind creating a new issue for this? This looks like something different, but something we should look into. Thanks!

mkalininait:
@carpodaster Thank you!
We could write something like: "Would you like to mentor newcomers in your local community? Become a RGSoC coach? Share your thoughts." WDYT?
Regarding that merging conflict: let's name the field "Why are you applying to this program?" with the hint: "Write a short essay explaining your motivation."

Issue #582, opened 2017-01-02
Application Form: Rename “Please link to pull requests”


alicetragedy:
The description and hint of this field should be more direct, so a suggested improvement would be to rename to “Please provide examples of your code” and add the following hint: “Links to completed coding challenge provided by the mentor (if present), commits / pull requests on GitHub or other concise examples of your code. Links to GitHub profiles and repositories won't be considered.”

Alternatively, the above could be combined into a single label/description of the field, i.e. “Please provide examples of your code: links to completed coding challenge provided by the mentor (if present), commits / pull requests on GitHub or other concise examples of your code. Links to GitHub profiles and repositories won't be considered.”

carpodaster:
Do we expect them to copy/paste entire code blobs into that textarea field? Why do we not consider external links? If the code exercise is mainly for the mentors, they will hardly be able to blindly select people anyway, right?

alicetragedy:
@carpodaster no, it should/could still be links, sorry if the wording made it sound like we wouldn't take links. what we don't want are links to *profiles* and *complete repos*, as often they're just e.g. super basic, generated code from scaffolds in the case of Rails apps.
But I think we could change the wording to make what we want more explicit. Maybe, “**Links to** completed coding challenge provided by the mentor (if present), commits / pull requests on GitHub or other concise examples of your code.”

mkalininait:
Good point @alicetragedy @carpodaster

alicetragedy:
I've updated the description accordingly :)

mkalininait:
Tested locally, looks good.

PR #607, opened 2017-01-11
improve wording of code samples label field

alicetragedy:
closes https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/582
I've also updated the wording for the rating app tooltips, lest we forget...

here's a screenshot if what it looks like, just so you get an idea of what information can be found where. We can definitely a) remove the hint and add all the text to the label or b) add the rest of important info (“we don't accept links to github profiles”) to the label, and leave the examples of what we want as a hint.
screen shot 2017-01-11 at 16 54 33


alicetragedy:
@carpodaster done! 🎉 
some of the values didn't seem super up-to-date in the locale file, so I wasn't sure if it needed updating. We might want to review it separately to make sure it's not missing anything :)

Issue #593, opened 2017-01-07
Application process: Force users to fill out their profiles


mkalininait:
In 2016 we had 2 problems:
 - Students could submit an application, while for some team members important information was missing (email, location etc)
 - We didn't have enough information about coaches, while they play a crucial role in the program.

**Possible solution:**  
1. After Users sign in with their Github profiles for the first time, they should be redirected to the Edit User Profile form pre-filled with the information from Github. This sign up process should be considered completed only when Users submit this form (with all required fields filled out).  
2. When a user has been added to a team, send an email asking to review/verify their profiles and confirm the membership.  
3. On application submission, validate that all members of the team have filled out all the required fields in their profiles and that they are confirmed team members of their team.  
4. User Profile:
 - Make the following fields required in the User Profile: Name, Location (Country, City)
 - Consider changing the Location fields so that they are consistent with [this requirement](https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/366) 
 - Make "Github handle" not editable

To be discussed:

 - Add the following fields (all required?): 
  -- "Technologies of expertise" (or "Technologies of interest"? or both?) - tags separated by comma (similar to the "Tags" field in Project)
  -- "Coaching experience" (only for coaches) - multiline textarea field
  -- "Availability (hours per week)" (only for coaches) - numeric field

Please share your opinions on this.

carpodaster:
(The Star Wars kid in me got really excited when I read _Force Users_ 😉 )

![](https://images.duckduckgo.com/iu/?u=https%3A%2F%2Fi.ytimg.com%2Fvi%2F000U_VNRW8s%2Fhqdefault.jpg&f=1)

bitboxer:
For 1: What does "only considered complete" mean? Is there a penalty if a users doesn't fill out the form and closes that window?
For 2: What should the wording of the email be?
For 3: What should the message be if a team member hasn't filled out all required fields?
For 4: When I am seeing this correctly, the github profile name is not editable already, or am I missing something?

alicetragedy:
[...]

PR #610, opened 2017-01-13
[#593] Application process: Force users to fill out their profiles

bitboxer:
The ticket was rather big, I hope I haven't forgot some part of it while implementing it.


spec/factories/application_drafts.rb
```diff
@@ -14,6 +14,7 @@
 
       after(:create) do |draft|
         draft.students.each do |student|
+          require 'pry'
```
carpodaster: I think this causes CI to break. Is it a local debug artifact?

bitboxer: Oh, sorry. I thought I already pushed a fix to this. Fixed now.

carpodaster:
@bitboxer I'm not familiar with advanced Devise Fu but will look into it later tonight.

bitboxer:
Don't worry. I totally understand. This is a rather big PR with changes all over the place.

db/migrate/20170112142045_add_confirmable_to_devise.rb
```diff
@@ -0,0 +1,15 @@
+class AddConfirmableToDevise < ActiveRecord::Migration[5.0]
+  def up
+    add_column :users, :confirmation_token, :string
+    add_column :users, :confirmed_at, :datetime
+    add_column :users, :confirmation_sent_at, :datetime
+    add_index  :users, :confirmation_token, unique: true
+    add_column :users, :unconfirmed_email, :string
+    execute("UPDATE users SET confirmed_at = NOW()")
```
carpodaster: Shall we be a bit more picky and only confirm those who were part of a selected team in any of the past seasons (as a student, coach, supervisor or mentor)?

bitboxer: I think it would confuse users who come back and don't expect this. There is no easy way for those users to understand what is wrong and why they now need to confirm their accounts when those accounts are years old.

carpodaster: Makes sense, I buy that!

bitboxer:
Sadly I can't. Have to wake up in 6 hours to fly to copenhagen.

carpodaster:
Alright + no worries. I'll do it. Good night and safe travels :sleeping: :airplane:

Issue #592, opened 2017-01-07
Application Form: "Diversity" field


mkalininait:
Our program aims at supporting diversity, and at the same time we don't have a special field for this on the application form. It would be great it we could support not only female/non-binary/trans people, but also take other special life circumstances into account.

**Possible solution:**  
Add an extra text field next to the gender field, something like: "Please tell us about any special circumstances which you want us to consider while reviewing your application (e.g. LGBTQI / people of colour / people with disabilities / single parents / people taking care of family members etc.)”. 

Of course, people should share only what they feel comfortable sharing.

anikalindtner:
plus one for the diversity field

alicetragedy:
+1 from my side as well

mkalininait:
Tested locally, looks good.

Issue #587, opened 2017-01-05
Application Form: Blind selection warning


mkalininait:
In 2016 the blind phase of the selection process wasn't really blind. Although we asked the students not to mention some of the information in certain fields (e.g. location), we still could sometimes tell from the writing where the student comes from (mentions of workshop names, university names etc.)

**Possible solution**  
After the block "Basics" and before the block "Background Information" put a warning, something like: "In your answers below, please don't mention names (geographical features, universities, companies etc.) which can point to your location and introduce bias during the selection process!".

F3PiX:
I added the warning to the Application Form. 

![schermafbeelding 2017-01-09 om 18 30 49](https://cloud.githubusercontent.com/assets/6314687/21776257/faa3863c-d699-11e6-827d-2b0e5de1fa83.png)


Am I right that it is meant for both the "Tell us about you" and the "Programming Background"? Not for everything after the Basics? 

(Note: I didn't add the warning to the User Profile fields, because I suppose we are going to do a shared partial, right, @carpodaster?)

mkalininait:
@F3PiX Actually this warning applies to *all* fields after the "Basics" section. I should have specified this more explicitly :/

I was thinking that we could make a warning like this, so that it's super visible:
![warning-app-form-2017](https://cloud.githubusercontent.com/assets/1745735/21803580/ee822c82-d72a-11e6-990d-544d13415fd6.png)

We don't need such a warning in the User Profile, because we don't look into user profiles during the blind phase of selection.

carpodaster:
> We don't need such a warning in the User Profile, because we don't look into user profiles during the blind phase of selection.

@mkalininait this only means that there are two ways of _filling out_ the info. It's not visible on a user's profile anyway. It's subject to removal (see #184).

The color of the info box should probably be yellow/orange instead of blue.

mkalininait:
@carpodaster @F3PiX I had something like this in mind
![warning-2-app-form-2017](https://cloud.githubusercontent.com/assets/1745735/21804008/1c4cc7ba-d72d-11e6-982a-d61775c8d4ce.png)

mkalininait:
We should also remove the same warning from the name of this field: "Why are you applying to this program?"  since we have one global warning now.

mkalininait:
@F3PiX Thanks!

mkalininait:
Tested locally, looks good.

PR #613, opened 2017-01-15
[#592] Application Form: "Diversity" field

mkalininait:
Closes https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/592

One thing to think about: this field is not hidden during the blind phase at the moment. Together with other information people may mention their age and location there, which we kinda don't want to know before the non-blind phase. Should we ask students not to mention this info?

There was the issue "Blind selection warning" (https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/587), but it only applies to the fields which go below the 'Basics' section.


mkalininait:
Screenshot:
screen shot 2017-01-16 at 00 48 07

app/views/application_drafts/_student_fields.html.slim
```diff
@@ -5,7 +5,7 @@
...
   = s.input :application_gender_identification, label: 'Your gender self-identification', input_html: { disabled: !may_edit?(s.object) }, wrapper_html: { class: student_field_error_class(index, application_draft, :application_gender_identification) }, hint: (student_field_error_hint(index, application_draft, :application_gender_identification) ||  'No-one is excluded from applying on the basis of gender, but people who self-identify as female, or have experiences being socialized as female, are given preference during selection.')
-
+  = s.input :application_diversity, label: 'Tell us about any special circumstances which you want us to consider while reviewing your application', input_html: { disabled: !may_edit?(s.object) }, wrapper_html: { class: student_field_error_class(index, application_draft, :application_diversity) }, hint: (student_field_error_hint(index, application_draft, :application_diversity) || 'e.g. LGBTQI, people of color, people with disabilities, single parents, people taking care of family members etc.')
```
carpodaster: Shall we make this field character-limited as well? In that case, we need a `as: :character_limited` in the form + it has be added to `Student::CHARACTER_LIMIT_FIELDS`

mkalininait: @carpodaster I don't have a strong opinion on this. It seems to me that it's pretty hard to write more than 2000 chars about your special circumstances, but what do I know. The limit wouldn't harm anyone, I guess.

mkalininait:
@carpodaster It would look a bit confusing, because there are the "Name", "Age" and "Location" fields in the first section, don't you think so? Maybe we make a mini-warning in the field hint?

carpodaster:
Yeah, I guess so. Putting it as a hint seems good! :+1:

mkalininait:
@carpodaster Done! Can I merge it now?

carpodaster:
@mkalininait I'll merge. I will add the character limit after that.

PR #616, opened 2017-01-17
Add whitelist check to Orga::ExportsController

### LINKED ISSUE #615: Possible security problem in Orga::ExportsController
In the ExportsController we have this starting at line 10:

```
exporter, meth = params[:export].to_s.split('#', 2)
csv = exporter.constantize.public_send(meth)
```

Instead of constantizing a random user input and calling a method on it, we should check against the whitelist that is generated in the `available_exports` method in that class to prevent a security problem.

bitboxer:
closes #615

I also checked what brakeman found and fixed a few minor things

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,6 +1,5 @@
 class ApplicationController < ActionController::Base
-  protect_from_forgery
-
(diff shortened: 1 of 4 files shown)


carpodaster:
🔝

PR #617, opened 2017-01-17
Removed return_to param and just redirect back

bitboxer:


bitboxer:
I fixed another security problem here that I overlooked: you could change the attendance status of every user in the database.

carpodaster:
Awesome – thanks, @bitboxer !

Issue #580, opened 2016-12-23
Application Form: Set pre-defined values for “how did you find out about RGSoC”


alicetragedy:
Right now, the `:heard_about_it` field is just a text field; having a list of pre-defined answers rather than a free-form field would allow us to have a better overview of how applicants find out about us, which is valuable information.

**Implementation**
- this list could lead either to something like a dropdown (i.e. only one possible answer) or a list of checkboxes (leading to one or more answers). Maybe the latter is preferable.

**Available Options**

Friends
RGSoC alumna
Social media (Twitter, Facebook)
Conference
Study group / Workshop
Newsletter
Other (please specify)
[additional text field for other]

mkalininait:
I would propose to extend the list in a way, which helps us differentiate our promotion channels better. Something like this:

**RGSoC direct outreach:**
 * RGSoC Blog
 * RGSoC Twitter
 * RGSoC Facebook
 * RGSoC Newsletter
 * RGSoC Organisers

**Past participants or 'Partners':**
 * Past RGSoC participants
 * Another diversity initiative outreach
 * Study group / Workshop
 * Conference

**Other channels:**
 * Friends
 * Mass media
 * Other (please specify)
 * [additional text field for other]

alicetragedy:
works for me!

ramonh:
Hello! I'll go ahead and give this a go if nobody has yet

alicetragedy:
@ramonh awesome! you can use the final list that Maria made. It doesn't need to be split into categories necessarily, those are just for us (I think).

ramonh:
Thanks for clarifying, @alicetragedy! Just to make sure, you mean that on the actual form the student sees the categories won't be displayed?

alicetragedy:
@ramonh yes, that's correct — I think it would be ok if it was just a list of all the options as checkboxes, without the user/student seeing what category each checkbox belongs to :)

ramonh:
Great, thank you, @alicetragedy!

carpodaster:
@ramonh @mkalininait @alicetragedy can we close this?

alicetragedy:
yes :D

PR #623, opened 2017-01-19
Use predefined values for "How have you heard about the program?" field

ramonh:
This addresses #580 by adapting the model of `application_drafts` to allow to have several `heard_about_it` options.

We've got predefined options this as outlined in the issue, categorised accordingly.

The applicant can also write down a value for `other`. Although the issue mentioned a checkbox to "activate" the "Other" field, I felt we could forgo this, as it might create an extra click in the workflow, and applications can disregard the "Other" field if they don't wish to fill it in.

This also removes the presence validation for `heard_about_it`.

### DIFF
--- a/app/controllers/application_drafts_controller.rb
+++ b/app/controllers/application_drafts_controller.rb
@@ -82,7 +82,7 @@ class ApplicationDraftsController < ApplicationController
 
   def application_draft_params
     params.require(:application_draft).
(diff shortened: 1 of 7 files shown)


ramonh:
Ahem, I forgot to attach a screenshot 😅 

screen shot 2017-01-19 at 17 12 22

ramonh:
Hey @alicetragedy, many thanks for the feedback! This makes a lot of sense. 

What I've done is changed it so that at least one option needs to be selected. What do you think?

alicetragedy:
@ramonh that sounds perfect. thank you! 🙌

PR #624, opened 2017-01-20
[#420] New Team / Edit Team issues

### LINKED ISSUE #420: New Team / Edit Team issues
1. When an error happens, all my deleted fields come back.
   For example, I have a team with Student X and Student Y. If I delete Student X and add Student Z, who is already a member of another team, and submit the form, I will get an error, and Student X will appear on the form again.
2. I think we should not allow creating empty teams. Now it's possible.

bitboxer:
closes #420

### DIFF
--- a/app/views/teams/_form.html.slim
+++ b/app/views/teams/_form.html.slim
@@ -21,19 +21,20 @@
     h2 Members
   fieldset
     = f.simple_fields_for :roles do |r|
-      = r.input :github_handle, required: false
-      / = r.input :name, as: :radio_buttons, collection: accessible_roles.map { |r| [r.capitalize, r] }, required: false, label: false
-      .form-group
-        label.control-label.col-sm-3 Team Member Role
-        .col-sm-9
-          - accessible_roles.each do |role|
-            .radio
-              label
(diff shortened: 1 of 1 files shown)


app/views/teams/_form.html.slim
```diff
@@ -21,19 +21,20 @@
...
-        .col-sm-9
-          - accessible_roles.each do |role|
-            .radio
-              label
-                = r.radio_button :name, role, required: true, disabled: (!current_user.admin? && r.object.persisted?)
-                = role.capitalize
-          .help-block Note: Coaches will be informed via email and have to confirm their role.
-          .form-btn-group
-            = r.link_to_remove 'Remove team member', class: 'btn btn-default'
+      - unless r.object._destroy
```
carpodaster: Since I haven't seen this before: is this something `accepts_nested_attributes_for` adds? I assume it holds `true` when it was marked to be destroyed in a previous form submit?

bitboxer: Yes, this is added by the nested attributes for. This is how the delete is handled. If you click on the delete button, it basically sets this hidden attribute. Sadly there is no other elegant way to make sure the deleted items don't reappear after a form validation fails and it re-renders the form.

Issue #621, opened 2017-01-18
Bug: "My Team" button opens wrong page


mkalininait:
I have two teams in different seasons: 2016 and 2017. The "My Team" button opens the team of 2016 for some reason.

![myteam](https://cloud.githubusercontent.com/assets/1745735/22079333/be95bd9e-ddbb-11e6-995e-6a2b35831d6c.gif)

Nada1996:
@mkalininait I think the problem is in the definition of current team
`  def current_team
    @current_team ||= current_user.teams.first
  end`
and on your profile ,it seems like the teams are sorted by chronological order, so if I wasn't mistaken, this makes "Team Rackless pinguins [2016] " the first team , therefore = @current_team
`link_to 'My Team', current_student.current_team, class: 'btn btn-default btn-sm' if current_student.current_team`

Hope this helps 😃

mkalininait:
@Nada1996 Cool, I think that's the problem! Thank you for the investigation 👍

Nada1996:
I hope that this solves the issue.
I'm glad to help 😃

mkalininait:
@Nada1996 Would you like to make a pull request? If you have time of course. Investigation is often the hardest part, and you already did it :)
Otherwise we will make it, not a problem either.

Nada1996:
@mkalininait yeah sure, I'll start working on it 👍

mkalininait:
@Nada1996 I've just noticed the following thing: in my application draft, on the "About your pair" tab, I see my partner from 2016, not the current one, too. Will the currently submitted PR resolve this issue, too?

carpodaster:
> UPD: It's actually not just the partner. The whole application draft is of 2016, and not the current one.

@mkalininait that sounds like a different problem. This sounds like one of the problems that Maud described in #611

carpodaster:
As for the original bug report: what is our expected behaviour for the "My Team" button, considering one can be part of more than one team (past and present)?

Can we just remove it or a "My Team(s)" item to the nav bar that leads to a filtered `teams#index` with only those teams I am part of?

mkalininait:
@carpodaster Sorry for not replying earlier, I've missed this comment somehow.

The initial idea was that the "My team" button opens my team in the current season. You are right that some people can be part of multiple teams, for example coaches. Although it's not a very popular case, we should take it into account, too.

Your solution sounds good to me. Maybe we can even add this filter to the teams page explicitly:
[...]

PR #625, opened 2017-01-20
Solving bug from issue #621

Nada1996:
This PR solves the issue of issue #621 , where the "My team" button goes to the oldest team instead of the current one.
So Teams.last correspond to the most recently added team therefore the current team

### DIFF
--- a/app/views/application_drafts/_student_fields.html.slim
+++ b/app/views/application_drafts/_student_fields.html.slim
@@ -43,6 +43,7 @@
   .form-inline
     = s.input :application_language_learning_period, :collection => User::MONTHS_LEARNING, include_blank: true, label: 'For how many months have you been learning the language of your primary project?', disabled: !may_edit?(s.object), wrapper_html: { class: student_field_error_class(index, application_draft, :application_language_learning_period)}, hint: student_field_error_hint(index, application_draft, :application_language_learning_period)
 
+  = s.input :application_learning_history, as: :character_limited, label: 'Please summarize what you have been doing to learn programming', input_html: { disabled: !may_edit?(s.object) }, wrapper_html: { class: student_field_error_class(index, application_draft, :application_learning_history) }, hint: (student_field_error_hint(index, application_draft, :application_learning_history) || 'e.g. entered a study group, learning Rails by doing exercises etc.') 
 
   = s.input :application_code_samples, as: :text, label: 'Please provide examples of your code.', input_html: { rows: 4, disabled: !may_edit?(s.object) }, wrapper_html: { class: student_field_error_class(index, application_draft, :application_code_samples) }, hint: (student_field_error_hint(index, application_draft, :application_code_samples) || 'e.g. Links to completed coding challenge provided by the mentor (if present), commits or pull requests on Github or other concise examples of your code. Links to GitHub profiles and repositories won\'t be considered.')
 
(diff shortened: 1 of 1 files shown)


app/controllers/students/base_controller.rb
```diff
@@ -8,7 +8,7 @@ def must_be_student
   end
 
   def current_team
-    @current_team ||= current_user.teams.first
+    @current_team ||= current_user.teams.last
```
bitboxer: This could also be problematic if you allow to create a team in the next season or so. Maybe there is a way to actually get the team from the current season? Would something like this `current_user.teams.where(season: Season.current).first` work?

Also it would be great if you could add a test for this.

F3PiX: We even [have](https://github.com/rails-girls-summer-of-code/rgsoc-teams/blob/master/app/models/team.rb#L59) `in_current_season` ;-)

bitboxer:
If you need any help writing the test, I can do a hangout with you to pair on this :+1: .

app/controllers/students/base_controller.rb
Nada1996: This sounds like a much better approach definitely prevent future issue 
So we can change it to this 
`current_user.teams.in_current_season`
?

Nada1996:
> If you need any help writing the test, I can do a hangout with you to pair on this

Sure this will be so helpful 👍

app/controllers/students/base_controller.rb
bitboxer: Yes, this could work. Haven't tried it myself yet. Would be awesome if you could check it and reply back here with your findings.

bitboxer:
In what timezone are you and when would you be available? Lets find something that works for you. Small disclaimer: will be on a short city trip starting on tuesday and will have to check my hotel wifi if it is good enough for hangouts.

app/controllers/students/base_controller.rb
carpodaster: One other thing: do we need to check for the current team where the user assumes a `student` role?

bitboxer: I don't understand that question. Can you explain what you mean by that?

carpodaster: While you can only access the controller as a student, it might be possible that you have once been a student but your most recent team membership is as a supervisor or coach. That way around `current_user.(any_team).in_current_season` could yield an irrelevant team, ie. one where you are not a student on that team.

bitboxer: Ah, I see. I have no Idea what is actually preferred here.

mkalininait:
@Nada1996 Do you think you could write the test and finish the issue? Or should we take it over? We would like to merge it before applications open tomorrow.
Thanks!

Nada1996:
@mkalininait I've working on the test but I'm having some issue with my Virtual machine, I'll keep working on it even 
I understand that you have that you have time constrains ,I think it's better if you take over , because I don't know when the error in my VM Ubuntu will be solved but I'll push the changes the minute I finish it anyways

alicetragedy:
@Nada1996 did you have any luck fixing the issue with your VM and finishing writing the tests for this? :)

mkalininait:
@alicetragedy @Nada1996 It actually appeared to be more complicated [as always :D], read [here](https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/621#issuecomment-276913772). So we probably want to reconsider the solution.

F3PiX:
@Nada1996 I am going to close this for now. Feel very welcome to open it again when you are ready to finish it!

PR #626, opened 2017-01-20
Fixed export functionality

bitboxer:
The dynamically created methods overwrote the param injected
into the method. This lead to the problem that the instance
could only export one thing. Every export after it was damaged
because of a wrong value in the season variable.

The navigation was kinda broken on all browsers because of
a wrong nesting of rows and widths of columns inside of the
row. For example in Safari you could not click on anything that
was close to the top navigation because of an div overlap.

I replaced it completely with a new, cleaned up navigation and
made it look like the old one.

bildschirmfoto 2017-01-20 um 23 08 58

bildschirmfoto 2017-01-20 um 23 08 52

### DIFF
--- a/app/assets/stylesheets/global.sass
+++ b/app/assets/stylesheets/global.sass
@@ -77,6 +77,7 @@ html, body
     margin-top: 12px
     outline: none
     text-align: center
+    border: none
     i
       position: relative
 
@@ -151,6 +152,7 @@ html, body
       background-color: transparent !important
       a
         background-color: transparent !important
+        text-shadow: none
 
     .navbar-collapse.collapsing,
     .navbar-collapse.collapse.in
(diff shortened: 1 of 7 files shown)


app/assets/stylesheets/global.sass
```diff
@@ -123,50 +135,40 @@ html, body
...
-      content: none !important
-
-
-  #menu-navbar
-    @media screen and (max-width: $screen-md)
-      z-index: 100
+          a
+            color: $railsred
 
   .navbar-toggle
```
alicetragedy: For some reason, there's now a side effect: the burger icon is getting a black border. Not exactly sure where it comes from 🤔 but a `border: none` should fix it..

bitboxer:
@alicetragedy Fixed both issues. And yes, side effects are always an issue. I tried to click through the pages and saw no other obvious css problem. Hope I haven't missed one :wink:

alicetragedy:
thank you @bitboxer for making the changes so quickly. 🎉

PR #628, opened 2017-01-23
Fix broken `heard_about_it` migration

ramonh:
This should fix #623, which was reverted in #627, and addresses #580. 😅 

The fix is brought about by replacing a broken migration with one that uses Postgres' `string_to_array` function to convert the field.


app/models/application_draft.rb
```diff
@@ -20,20 +44,21 @@ class ApplicationDraft < ActiveRecord::Base
...
   validate :accepted_projects_required, on: :apply
+  validate :at_least_one_heard_about_it, on: :apply
 
   validates *STUDENT0_REQUIRED_FIELDS, presence: true, on: :apply
   validates *STUDENT1_REQUIRED_FIELDS, presence: true, on: :apply
   validates *STUDENT0_CHAR_LIMITED_FIELDS, length: { maximum: Student::CHARACTER_LIMIT }, on: :apply
   validates *STUDENT1_CHAR_LIMITED_FIELDS, length: { maximum: Student::CHARACTER_LIMIT }, on: :apply
 
   before_validation :set_current_season
+  before_save :clean_up_heard_about_it
```
klappradla: You should be able to get the same behavior with
```ruby
validate : heard_about_it, presence: true
```
Since you already set the array type on db level. So you could spare both, the before-save callback and the custom validation.

I'd just add a test for it and give it a try.

app/views/application_drafts/_form.html.slim
```diff
@@ -72,7 +72,15 @@ p.attention
       = f.input :voluntary_hours_per_week, as: :integer, label: 'As a voluntary team, how many hours per week would you be able to work on the project?', hint: 'Please only enter a full number, e.g. 15', input_html: { disabled: !application_draft.as_student? || application_draft.applied? }
 
       h2 Miscellaneous
-      = f.input :heard_about_it, as: :text, label: 'How have you heard about the program?', input_html: { rows: 3, disabled: !application_draft.as_student? || application_draft.applied? }
+
+      h3 How have you heard about the program?
+      = f.input :heard_about_it, collection: ApplicationDraft::DIRECT_OUTREACH_CHOICES.map {|choice| [choice, choice]}, as: :check_boxes, label: "RGSoC direct outreach", :input_html => {multiple: true }, error: false
```
klappradla: I'm not sure if I understand this correctly. So this will submit 4 values for the same key and therefore and up as an array? How do you make sure it preserves the order?

db/migrate/20170123173604_adapt_heard_about_it_to_array.rb
```diff
@@ -0,0 +1,5 @@
+class AdaptHeardAboutItToArray < ActiveRecord::Migration[5.0]
+  def change
+    change_column :application_drafts, :heard_about_it, :string, array: true, default: [], using: "(string_to_array(heard_about_it, ','))"
```
klappradla: Is this really reversible with `rails db:rollback`? Probably...

db/schema.rb
```diff
@@ -10,32 +10,32 @@
...
   enable_extension "plpgsql"
   enable_extension "hstore"
 
   create_table "activities", force: :cascade do |t|
     t.integer  "team_id"
-    t.string   "kind",         limit: 255
-    t.string   "guid",         limit: 255
-    t.string   "author",       limit: 255
-    t.string   "title",        limit: 255
+    t.string   "kind"
```
klappradla: Why did you remove **all** char-limits without adding a migration? Is this done automatically via an update to Rails? I don't really get this 🙈

ramonh: Huh?! This was not intended. Will need to fix this.

ramonh: For what it's worth, I think it was done automatically, but don't know why I wasn't informed.

carpodaster: I've yet to get behind how Rails generates the schema.rb. It would seem the schema dump depends heavily on the actual pgsql patchlevel used when the migrations were run. :woman_shrugging:

db/migrate/20170123173604_adapt_heard_about_it_to_array.rb
ramonh: Oh! No it's not. Will need to add a `up` and `down`

app/views/application_drafts/_form.html.slim
ramonh: Having the `multiple: true` allows me to have any number of inputs with the `name=application_draft[heard_about_it][]` attribute. 

From what I understand, the order doesn't make a difference

db/schema.rb
F3PiX: Is this about this comment by Raphael Franca  https://github.com/rails/rails/issues/19001#issuecomment-75047033.

app/models/application_draft.rb
ramonh: You're right! Arg, thank you for catching this.

db/schema.rb
ramonh: Good catch @F3PiX. Does this mean, if I understand correctly, that it's not relevant anymore?

F3PiX: I think so, yes.

ramonh:
Thank you very much for your review @klappradla! Helped me a lot. I've made some cleanups. Turns out I needed the cleanup `before_save` callback to remove empty strings, but the rest worked out, hopefully

db/schema.rb
bitboxer: Yes, it relies on the used psql version. Because of that I am always in favour of using only one postgres version everywhere. From dev over staging to production. Even small version differences can have huge changes in behavior. I have a tool called [pgvm](https://github.com/guedes/pgvm) on my machine when switching between projects. Other people I know use separate docker containers for each project and only work inside of the container.

alicetragedy:
I'm not sure what the current status is here. @ramonh, do you still need someone to review this? I feel a bit unsure about taking a look at it, but it would probably be good to merge it soon so we can also test it before opening applications. :)

ramonh:
Hey, thanks @alicetragedy. I think it would be good to have another look at it. I re-reviewed the migration command string `string_to_array` and it seems safe, but one can never be too sure.

ramonh:
Oh but there's conflicts here, do I need to rebase?

alicetragedy:
@ramonh yes, I think rebasing would be the way to go (I have tried resolving conflicts from the command line using the GH instructions in the past but iirc it just created an ugly merge commit)

ramonh:
Alright! Rebased and cleaned up. Thanks for your help @alicetragedy

app/views/application_drafts/_form.html.slim
carpodaster: Interesting, I didn't know that you could have more than one input field adding up to the same array param value. Does the form keep the selected values over its different input elements? As in: if you select values, save and (hard)reload the form: are the previously selected values still there.

klappradla:
Once this branch is put back up to date with current `master` I think this should be fine. After all, I also don't see a problem in having the `limit: 255`s removed on not as long as the other relevant information from the schema is there.

alicetragedy:
@ramonh sorry, could you rebase this again? we might be able to merge then... Right on time for the applications opening at 18:00 UTC 😱

ramonh:
Alright, we should be in business! Thanks for the feedback again

alicetragedy:
@klappradla you want to merge this one? ;)

alicetragedy:
ok, I'm doing this. 😁

Issue #591, opened 2017-01-07
Application Form: "Gender" autocomplete


mkalininait:
To make it a bit easier for us to process applications we decided to make an improvement over the field "Gender". We've been thinking a lot about how to implement a new/better “Gender Self-Identification” field so that on the one hand it had some pre-defined options and on the other hand wasn't exclusive for anyone. The majority voted for the following solution.

The Gender field should look like a simple text field with autocomplete (similar to how we have the license autocomplete for project submissions). It can look like this:
![gender](https://cloud.githubusercontent.com/assets/1745735/21742349/1962c33e-d4ed-11e6-8e2f-08261fbb3dfe.jpg)

Here is the list of gender options on Facebook, we could use it for starters or find a different one: http://alphabetizer.flap.tv/lists/list-of-facebook-gender-options.php

alicetragedy:
I'd love to take a shot at this. :)

alicetragedy:
My proposed list to choose from for now: "Agender", "Bigender", "Cisgender Woman", "Gender Fluid", "Gender Nonconforming", "Gender Questioning", "Genderqueer", "Non-binary", "Transgender Woman". (we can always expand on it if needed, this list specifically leaves out “Cisgender Man” as requested by Anika). 

The field should give the possibility to choose from the above mentioned list, or to add your own (like in a “normal” input field) if your preferred option is not available.

carpodaster:
Fixed in #631

mkalininait:
Tested locally. Looks good!

PR #631, opened 2017-01-26
Add autocomplete to gender identity field and turn the field into single liner

ramonh:
This addresses #591 by both adding autocomplete to the gender identity field, as well as turning the field itself into a single line textfield from a 2-line text area.

The values for identities come from the proposal [here](https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/591#issuecomment-273421058).


carpodaster:
I checked it out locally - looks good! One thing I did notice though: I can imagine many applicants want to just enter "female", possibly meaning "cisgender female" because not everyone is familiar with the more specific gender terms. "female" doesn't trigger the autocomplete list, suggesting that it's a "wrong answer"?

This is nothing about this PR but more about either the list of autocomplete options or the way jQuery Autocomplete works.

alicetragedy:
@carpodaster yeah that's a good point.. but typing “Woman” works (as we have “Cisgender Woman“ and “Transgender Woman” as options), so it's not really an issue of how the autocomplete works, more a problem of the original list I made. We could definitely add the option “Female” maybe? wdyt @ramonh?

ramonh:
Could definitely do, @alicetragedy. Thanks for the feedback, @carpodaster! 

In your opinion, should this be a separate commit for documentations' sake, or just squashed? Or am I perhaps overthinking this 😨

alicetragedy:
@ramonh I have no real opinion there, but I think squashed sounds good 👍

ramonh:
Alrighty, I've added the field! Thanks again for the feedback.

alicetragedy:
I'm going to go ahead and merge this.... #yolomerge

Issue #649, opened 2017-02-08
Application form: Mark fields containing info which will be provided to mentors


alicetragedy:
For some reason, I thought this was already done, but it's not. 
The application form should show which fields will be visible to mentors: 

- Why did you select this project?
- Which features are you planning to work on?
- What is your programming level?
- Provide examples of your code
- What you have been doing to learn programming
- How did you get into programming?
- Short summary of your programming skills

Not sure if we should place a notice with a list of fields, have single notices for each field, or add a sentence about it as a `hint` for each field. Thoughts? 🤔

mkalininait:
@alicetragedy Do we really show "How did you get into programming?" (it was not on the list [here](https://github.com/rails-girls-summer-of-code/organization/wiki/Processes:-Application-Process-2017-(design)#other-changes), that's why I'm clarifying).
Do we not show "For how many months have you been learning the language of your primary project?"?

mkalininait:
Maybe we can add notices for the blocks of fields?
 * "Your programming skills"
 * "Your 1st/2nd project choice"

And I'd probably **not** put them in the yellow/blue box (we already have plenty of notes and warnings!), but rather write as a normal text under the title.

alicetragedy:
I've taken the list from the current existing mentor view, we can also make changes to this if needed (but then it has to happen now). I'm also not sure about the “for how many months”, I also feel like we wanted to actually remove that field entirely from the application...? Anyway it's in there now, so we should probably keep it that way and fix the mentor view accordingly. 

I think if we manage to make it so that we share entire “sections” of the form, it would make sense to use a real `notice` element (the yellow or blue boxes) to make it visible — that's what the element is there for. 
if it's single fields we are picking out, it makes sense to have single lines of “normal” text under each label (or each field, depending on how we want to do it) to not overwhelm the user with a bunch of notices scattered everywhere.

alicetragedy:
@mkalininait I had created an issue for the mentor view originally, and I took the list of fields from the doc you linked to — but it could very well be that I missed some or added one accidentally that we didn't want to have in there. Anyway, I'll look at this later today and can make whatever changes are required :)

mkalininait:
@alicetragedy 
**"I'm also not sure about the “for how many months”..."** – there were 2 fields: "For how many months have you been learning since your first workshop" and  "For how many months have you been learning the language of your primary project". We removed the first and left the second.
[...]

PR #651, opened 2017-02-09
Update application form and application overview in mentor dashboard

alicetragedy:
this should close https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/649

I've added info alerts to the application form to show students which fields will be visible to mentors, and removed the `code_background` field from the mentor view, which was replaced by the `language_learning_period`. 
Tests have been updated accordingly.
For some reason though, I get a weird `JSON::ParserError` when I try to view the application now. Is there anything I've missed? Would love a push in the right direction 😬

### DIFF
--- a/app/models/mentor/student.rb
+++ b/app/models/mentor/student.rb
@@ -6,12 +6,15 @@ module Mentor
       '_application_coding_level',
       '_application_code_samples',
       '_application_learning_history',
-      '_application_skills',
-      '_application_code_background'
+      '_application_language_learning_period',
+      '_application_skills'
+      
     ]
 
     attr_accessor :coding_level, :code_samples
-    attr_accessor :learning_history, :code_background
+    attr_accessor :learning_history
+    attr_accessor :language_learning_period
     attr_accessor :skills
+    
   end
 end
--- a/app/views/mentor/students/_student.html.slim
+++ b/app/views/mentor/students/_student.html.slim
@@ -11,8 +11,8 @@ div.panel.panel-default
   div.panel-body #{auto_link simple_format(student.learning_history)}
 
 div.panel.panel-default
-  div.panel-heading #{ApplicationDraft.human_attribute_name :student_application_code_background }
-  div.panel-body #{auto_link simple_format(student.code_background)}
+  div.panel-heading #{ApplicationDraft.human_attribute_name :student_application_language_learning_period }
+  div.panel-body #{auto_link simple_format(student.language_learning_period)}
 
 div.panel.panel-default
   div.panel-heading #{ApplicationDraft.human_attribute_name :student_application_skills }
--- a/spec/factories/application.rb
+++ b/spec/factories/application.rb
@@ -31,13 +31,13 @@ FactoryGirl.define do
         "student0_application_code_samples": FFaker::Lorem.paragraph,
         "student0_application_learning_history": FFaker::Lorem.paragraph,
         "student0_application_skills": FFaker::Lorem.paragraph,
-        "student0_application_code_background": FFaker::Lorem.paragraph,
+        "student0_application_language_learning_period": FFaker::Lorem.paragraph,
 
         "student1_application_coding_level": "5",
         "student1_application_code_samples": FFaker::Lorem.paragraph,
         "student1_application_learning_history": FFaker::Lorem.paragraph,
         "student1_application_skills": FFaker::Lorem.paragraph,
-        "student1_application_code_background": FFaker::Lorem.paragraph
+        "student1_application_language_learning_period": FFaker::Lorem.paragraph
       }}
     end
   end
--- a/spec/factories/application.rb
+++ b/spec/factories/application.rb
@@ -26,18 +26,18 @@ FactoryGirl.define do
         'plan_project2': (project2.nil? ? FFaker::Lorem.paragraph : ''),
         'why_selected_project1': FFaker::Lorem.paragraph,
         'why_selected_project2': (project2.nil? ? FFaker::Lorem.paragraph : ''),
-        
+
         "student0_application_coding_level": "1",
         "student0_application_code_samples": FFaker::Lorem.paragraph,
         "student0_application_learning_history": FFaker::Lorem.paragraph,
         "student0_application_skills": FFaker::Lorem.paragraph,
-        "student0_application_language_learning_period": FFaker::Lorem.paragraph,
+        "student0_application_language_learning_period": User::MONTHS_LEARNING.sample,
 
         "student1_application_coding_level": "5",
         "student1_application_code_samples": FFaker::Lorem.paragraph,
         "student1_application_learning_history": FFaker::Lorem.paragraph,
         "student1_application_skills": FFaker::Lorem.paragraph,
-        "student1_application_language_learning_period": FFaker::Lorem.paragraph
+        "student1_application_language_learning_period": User::MONTHS_LEARNING.sample
       }}
     end
   end
--- a/spec/models/mentor/student_spec.rb
+++ b/spec/models/mentor/student_spec.rb
@@ -7,7 +7,7 @@ describe Mentor::Student do
     it { is_expected.to respond_to :coding_level     }
     it { is_expected.to respond_to :code_samples     }
     it { is_expected.to respond_to :learning_history }
-    it { is_expected.to respond_to :code_background  }
+    it { is_expected.to respond_to :language_learning_period }
     it { is_expected.to respond_to :skills           }
   end
 end
--- a/spec/models/mentor/student_spec.rb
+++ b/spec/models/mentor/student_spec.rb
@@ -4,10 +4,10 @@ describe Mentor::Student do
   describe 'attributes' do
     subject { described_class.new }
 
-    it { is_expected.to respond_to :coding_level     }
-    it { is_expected.to respond_to :code_samples     }
-    it { is_expected.to respond_to :learning_history }
+    it { is_expected.to respond_to :coding_level             }
+    it { is_expected.to respond_to :code_samples             }
+    it { is_expected.to respond_to :learning_history         }
     it { is_expected.to respond_to :language_learning_period }
-    it { is_expected.to respond_to :skills           }
+    it { is_expected.to respond_to :skills                   }
   end
 end
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: ruby
 cache: bundler
 sudo: false
 rvm:
-- 2.3.1
+- 2.3.3
 bundler_args: "-j4 --without development production --quiet"
 notifications:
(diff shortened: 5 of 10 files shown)


carpodaster:
@alicetragedy I've checked out the branch locally to reproduce the JSON error but got an `undefined method 'join' for #<String:0x007ff9900b86f0> Did you mean? JSON` instead. That was because I `dup`ed application data from last year.

I could get around that fix with this patch:

```diff
--- a/app/views/mentor/students/_student.html.slim
+++ b/app/views/mentor/students/_student.html.slim
@@ -1,6 +1,6 @@
 div.panel.panel-default
   div.panel-heading #{ApplicationDraft.human_attribute_name :student_application_coding_level }
-  div.panel-body #{SkillLevel::MATRIX[student.coding_level].join('<br>').html_safe}
+  div.panel-body #{Array(SkillLevel::MATRIX[student.coding_level]).join('<br>').html_safe}

 div.panel.panel-default
   div.panel-heading #{ApplicationDraft.human_attribute_name :student_application_code_samples }
```

Do you have a stacktrace for your error?

alicetragedy:
@carpodaster 

```
JSON::ParserError - 784: unexpected token at '{"student0_application_skills": "*this*\r\nand **this**", "student0_application_code_samples": "yolo", "student0_application_coding_level": 2, "student0_application_learning_history": "moar", "student0_application_language_learning_period": 10-12}'
```

carpodaster:
Could be that the value `10-12` is not valid json. No idea how it got there w/o quotation marks, though. Gonna check tomorrow

klappradla:
You're right @carpodaster: `10-12` is invalid JSON (and invalid Ruby) and not parsable by the default JSON parser from stdlib. It should be `"10-12"` as a string @alicetragedy. Do you have an idea how this value ended up there?  I don't even know how it could have been serialized to *non-string* 🙈

alicetragedy:
@klappradla @carpodaster I honestly have no idea how that ended up there. I should probably get a clean production db dump to avoid this stuff in the future. 🙈 will try to get that cleaned up asap. Sorry for wasting your time there 🙈

spec/models/mentor/student_spec.rb
```diff
@@ -7,7 +7,7 @@
     it { is_expected.to respond_to :coding_level     }
     it { is_expected.to respond_to :code_samples     }
     it { is_expected.to respond_to :learning_history }
-    it { is_expected.to respond_to :code_background  }
+    it { is_expected.to respond_to :language_learning_period }
```
klappradla: It's not a must, but it kinda hurts my eye to not see the `}`'s aligned 😉 
There's "aligning" or "not aligning", but the compromise, I don't know...

app/models/mentor/student.rb
```diff
@@ -6,12 +6,15 @@ class Student
...
       '_application_learning_history',
-      '_application_skills',
-      '_application_code_background'
+      '_application_language_learning_period',
+      '_application_skills'
+      
     ]
 
     attr_accessor :coding_level, :code_samples
-    attr_accessor :learning_history, :code_background
```
klappradla: Seeing this, I would probably have kept both. Is `code_background` *too much* or *too little* information for the mentors? Or is it just to personal?

spec/factories/application.rb
```diff
@@ -31,13 +31,13 @@
         "student0_application_code_samples": FFaker::Lorem.paragraph,
         "student0_application_learning_history": FFaker::Lorem.paragraph,
         "student0_application_skills": FFaker::Lorem.paragraph,
-        "student0_application_code_background": FFaker::Lorem.paragraph,
+        "student0_application_language_learning_period": FFaker::Lorem.paragraph,
```
klappradla: From what I remember, `learning-period` is a string of numbers, e.g. `"1-3"`, etc. I'd suggest to also keep it like this in the Factories, e.g. using something like
```ruby
User::MONTH_LEARNING.sample
```
It may be personal preference, but I always tend to think, having ones test data as close to the *"real world"* as feasable without too much effort, helps keeping the test suite sane and readable *(and hopefully powerful enough to find out weirdnesses)*

app/views/mentor/students/_student.html.slim
```diff
@@ -11,8 +11,8 @@ div.panel.panel-default
   div.panel-body #{auto_link simple_format(student.learning_history)}
 
 div.panel.panel-default
-  div.panel-heading #{ApplicationDraft.human_attribute_name :student_application_code_background }
-  div.panel-body #{auto_link simple_format(student.code_background)}
+  div.panel-heading #{ApplicationDraft.human_attribute_name :student_application_language_learning_period }
+  div.panel-body #{auto_link simple_format(student.language_learning_period)}
```
klappradla: I'd spare the `auto_link` here. The field is not intended to contain html-links and I'd suggest to just keep things as simple as possible here.

spec/models/mentor/student_spec.rb
alicetragedy: haha, yeah, I totally get you. 👍

app/models/mentor/student.rb
alicetragedy: `code_background` tends to get a little on the personal side, and is often just a lot of information that is not super necessary to mentors. I think we were mainly worried here about how much time mentors would spend reading this info if it's not particularly relevant to the skills they would bring to the table.

spec/factories/application.rb
alicetragedy: 👍

alicetragedy: For some reason adding `User::MONTHS_LEARNING.sample` gives me a ton of failures on the test suite (`mentor/applications_controller` and `mentor/application` specs), all related to the `JSON::ParserError` I was getting before with my local application. Because of this I'm wondering whether that parsing error is an error with my local data after all – the test data shown when running the tests gives me stuff like `"student1_application_language_learning_period": 10-12` which as you and Carsten already pointed out, is invalid. 
I'm confused and not sure how to best look into this 😭

klappradla: Sucks, but good to know. Seems we're doing something wrong with either serializing the data from the `ApplicationDraft` or retrieving in from the DB later... From what I remember the `User::MONTHS_LEARNING` array is Strings only.

klappradla:
Ok, @alicetragedy (cc @carpodaster ) I made some experiments 👨‍🔬:

I think you found a _"bug"_ in `JSON` from stdlib Ruby < 2.4.

This is Ruby `2.3.1`:
```
❯ pry
[1] pry(main)> require 'json'
=> true
[2] pry(main)> JSON.parse("1-3".to_json)
JSON::ParserError: 784: unexpected token at '"1-3"'
from /Users/mamu/.rbenv/versions/2.3.1/lib/ruby/2.3.0/json/common.rb:156:in `parse'
[3] pry(main)>
```

This is Ruby `2.4.0`:
```
❯ pry
[1] pry(main)> require 'json'
=> true
[2] pry(main)> JSON.parse("1-3".to_json)
=> "1-3"
[3] pry(main)>
```

Suggestion: *Let's just switch to 2.4.0* - shouldn't impact anything else.

klappradla:
Just if you wonder: [yes 2.4 is available on Heroku](https://blog.heroku.com/ruby-2-4-features-hashes-integers-rounding)

carpodaster:
Not in front of a computer until Monday. Do the tests pass on 2.4? I wanted to upgrade some apps st work and noticed one or two gems not coping too well with the Integer unification

klappradla:
Yeah @carpodaster depending on how up to date the dependencies are, an update to 2.4 would also mean updating **them all** 😉  - everything else would end in warnings-hell.

I gave `bundle update json` and using `2.3.3` a quick try and this looks good.

klappradla:
Hey @alicetragedy, you wanna give this another try? I added some commits updating to Ruby 2.3.3, this should fix the JSON problems *(I hope)*.

alicetragedy:
@klappradla sure, I'll give it a try tomorrow or Sunday, whenever I find stable internet ✌️ thanks so much for looking into it, I'm kind of happy to see the issue wasn't really on my end :D

alicetragedy:
@klappradla I seem to be getting the issue still, also when I run the tests. Do they all pass for you? 🤔

klappradla:
Yes @alicetragedy they should be passing - see [Travis](https://travis-ci.org/rails-girls-summer-of-code/rgsoc-teams/builds/202757178)
Are you sure you're on Ruby `2.3.3` and the tests are also ran with `2.3.3`? I don't know which Ruby version manager you're using, but they pretty much all work with `.ruby-version` files. So I'd suggest you just put one in the root directory of the project, containing just this string: `2.3.3` - then you're Ruby version manager should pick up the right version.
I'm for instance using Rbenv - which does not work with specifying a Ruby version in the Gemfile.

alicetragedy:
@klappradla is there anything else that could cause “my” tests to fail? I use rvm and had already set it to `2.3.3`,  `rvm current` also returns `ruby-2.3.3`, I've added a `.ruby-version` file just in case too, but it seems like something's still not quite right. 🤔 
I still get `1287 examples, 10 failures, 3 pending` every time I run all tests. This must be clearly some issue on my system then, because otherwise it makes no sense :(

klappradla:
And the failures are still due to JSON-parsing issues I suppose...?

Only thing I can think of right now would be around *"there has to still be something wrong with your dependencies"*. Your Postgres version should probably be around the one used on Travis *(>= 9.3)*, this shouldn't impact anything.

Maybe still gem dependencies. Are you using rvm's "gemsets"? If so, you could create a new gemset for Ruby 2.3.3, e.g.
```sh
# create it
rvm gemset create teamsapp_json

# optionally use a .ruby-gemset file as well
echo teamsapp_json > .ruby-gemset

# use it
rvm use 2.3.3@teamsapp_json

# bundle from scratch
gem install bundler
bundle install
```

If you don't care about gemsets and do not need your installed Ruby 2.3.3 to have things setup a certain way, you could also just empty your current gemset and pull in stuff from scratch there. Would be like:
```sh
# find out the currently used gemset
rvm gemset list # (I'm not 100% sure about the command, should be something like this)
rvm --force gemset empty <name-of-your-current-gemset>

# bundle from scatch
gem install bundler
bundle install
```

It's been years that I used rvm, but I think the workflow should be still approx. like this.

klappradla:
Just randomly saw your commit-message around Ruby 2.3.3 @michaelem - maybe you can could also take a look here and see if you run into any issues with the state of this PR...? Especially after updating to the here locked gem versions (JSON for instance).

alicetragedy:
yes, the failures are still due to the JSON parsing issues. I will see if that problem comes from my dependencies as you suggested, maybe I can find something there. Will create a clean gemset and see if it helps 😅 thanks so much for painfully going through all this with me 🙈

klappradla:
✔️

PR #652, opened 2017-02-09
Update applications.sass

gargi-gupta:


carpodaster:
Hello @gargi-gupta and thank you for taking the time to contribute to RGSoC!

It is not really clear why this change is necessary. This change also only affects a feature of the Teams App that's not visible to the public.

Issue #644, opened 2017-02-06
GitHub ribbon causes vertical overflow of body


untitaker:
On Firefox 51, the GitHub ribbon causes a vertical scrollbar for the entire page. After scrolling sideways to the right:

![2017-02-06-181144_280x298_scrot](https://cloud.githubusercontent.com/assets/837573/22657737/ce4fa4da-ec97-11e6-8750-e81d079a317d.png)

carpodaster:
Thanks, @untitaker, for reporting this.

@klappradla @alicetragedy could this be a side-effect of the new navigation bar?

untitaker:
If I delete the `.ribbon` element this issue is gone, IIRC the problem has been there before too.

klappradla:
Oh thanx @untitaker 💚  - that has to to with the new navigation, you're right @carpodaster. I had to adapt some of the styling, since the old ribbon was not *"bootstrap-confirm*". I'll get Firefox and take care of this *(unless it's super urgent)*.

klappradla:
Hm, to be honest, I'd prefer to fix this issue by just replacing the current approach with an adapted version of [this one](https://github.com/simonwhitaker/github-fork-ribbon-css) - it's simply the best 💚 
The one we currently have uses an approach [like this](https://codepen.io/hmps/pen/kiFum), which results in overflowing edges and is way harder to fix than just using a single `a` tag.

I'm way too tired now, but I'll find time during this week I think.

alicetragedy:
Sounds good @klappradla. Thanks for looking into it!

PR #654, opened 2017-02-09
Fix behavior of default avatar in navbar

klappradla:
As I briefly noted in https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/644#issuecomment-277783073, the "default" avatar image returned if there is no Github-provided avatar for a user kinda bloats the navigation bar.

This PR aligns the behavior of both types of images, default and "normal" avatar to just work fine with the navigation layout.


mkalininait:
Awesome!

Issue #648, opened 2017-02-08
Update mentor dashboard to be visible to all submitters of a project


alicetragedy:
Right now the mentor dashboard is only visible to mentors who **have been added to a team as mentors**. 
Since we are not particularly explicit to applicants or coaches about adding mentors to their teams anywhere, it might be that some teams do so and some don't (or alternatively, that some teams add one mentor only when they've applied for two projects... basically, a lot could go wrong). 
To avoid penalising some teams over others, it would be good to allow the mentor dashboard to be visible to all submitters of a (non-rejected?) project. 

Note: I'm not sure if we want to make this open to all Seasons or not. Ideally it would have to be only for the **current season**, I don't think it's relevant for past mentors to have access to their dashboard in any season if they don't have an accepted project in the current one.

Note from Max: Would be nice to have a shortcut for grouping them all — something like `role.submitter?`. Otherwise I’d probably have to do some heavier queries around “all projects of the season, give me your submitters” just to find out which navigation elements to render.

PR #656, opened 2017-02-09
Show mentor dashboard for project submitters

klappradla:
Hey @alicetragedy, this adds the changes specified in #648: submitters of accepted projects for the current season have access to the mentor dashboard during application phase.

The applications they see are of course also from the current season.

### DIFF
--- a/app/controllers/mentor/applications_controller.rb
+++ b/app/controllers/mentor/applications_controller.rb
@@ -11,7 +11,10 @@ class Mentor::ApplicationsController < Mentor::BaseController
   private
 
   def projects
-    Project.where(submitter: current_user)
+    Project.
+      in_current_season.
+      accepted.
+      where(submitter: current_user)
   end
 
   def application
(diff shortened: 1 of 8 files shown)


carpodaster:
> I'm not sure it makes sense to allow both submitters AND users with role “Mentor” to view the dashboard as it wouldn't really solve our problem. wdyt?

I believe we only assign mentor roles when the team has been officially picked.

alicetragedy:
@carpodaster Ah sorry, I think I am confused about the roles again... What's a person that is assigned to a team as a Mentor called? Currently there is the possibility for people that have been added to a team as “Mentors” to view the mentor dashboard — this PR would remove that possibility. so I was wondering whether we would still want that or not.

mkalininait:
@alicetragedy maybe we just ignore those 3 projects where "submitters are not “necessarily” the final mentors" for the moment and implement a good, consistent solution for the next year? I'm a bit afraid of making the app logic even more complicated while fixing things in a rush. Maybe we could find custom workarounds for those 3 projects...

alicetragedy:
anyone want to do the honours of pressing the Merge button? ;) :man_dancing:

PR #664, opened 2017-02-13
Updated the help pages (screenshots, new info, cross-links etc)

mkalininait:


app/views/pages/create-an-account.html.slim
```diff
@@ -24,12 +25,18 @@ p
...
 
 img
-  = image_tag(("pages/help/sign-in-link.png"), class: "help-image")
+  = image_tag(("pages/help/nav-bar-sign-in.png"), class: "help-image")
 
 h2 2. Fill out your profile
 
 p
-  ' Once signed in, you should update your profile information. There are a few mandatory fields:
+  ' Once signed in, you should update your profile information.
```
alicetragedy: can you change this part to the following and remove the image?

“once signed in for the first time, you will be redirected to the ‘Edit your profile’ section. In order to keep using the app, you MUST fill out your profile with the mandatory fields (see list below). Don't forget to click on the ‘Save’ button to save your profile — otherwise you will not be able to proceed.”

The problem seems to be that users log in and disregard the “edit profile” screen they get. Then they just click elsewhere in the app, and are automatically logged out/unable to do anything.

app/views/pages/create-an-account.html.slim
```diff
@@ -64,12 +71,56 @@ p
...
+.basic-list
+  ul
+    li Successfully authenticated from Github account.
+    li You have to confirm your account before continuing.
+
+img
+  = image_tag(("pages/help/notice-cannot-sign-in.png"), class: "help-image")
+
+p
+  ' This is a known bug which happen rarely. Please contact us at <a href="mailto:contact@rgsoc.org">contact@rgsoc.org</a>.
```
alicetragedy: which happens

mkalininait:
@alicetragedy I've committed the changes. Although the bug with signing in happens because users don't fill out their profiles, we can't really control this. We can shout about filling out the profiles on every corner (an this is probably the right thing to do), but we can't force them to click the "Save" button in the end. 
All I want to say is that updating Help is good, but we need to fix that bug with signing in anyway.

alicetragedy:
@mkalininait yep, I agree with that, there's nothing we can really do about it too much (except fix the bug :p )

PR #665, opened 2017-02-22
fix an error thrown when an unconfirmed user logs in

michaelem:
This should address the error thrown when some unconfirmed users log in.


carpodaster:
Thanks @michaelem! I'm not too familiar with Devise (in fact, I don't like it at all because of all the hidden magic). Where is `new_user_session_path` defined? It's not a named route helper known to `rake routes`.

I don't think this error is triggered for unconfirmed users only. It seems to appear when the GH authentication fails for whatever reason. Sadly, that is hard to reproduce locally 😞

bitboxer:
I think this is good and should be merged. What is the policy here to do so?

carpodaster:
> I think this is good and should be merged

I don't think so, no. See comment above: not familiar with Devise, but it looks like we'd exchange the error message in a name error:  `new_user_session_path` is not defined.

bitboxer:
Ah, yes, that is right. If you open [the documentation](https://github.com/plataformatec/devise/wiki/OmniAuth%3A-Overview#using-omniauth-without-other-authentications) it says that
you need to add this, too:

´´´ruby
  get 'sign_in', :to => 'devise/sessions#new', :as => :new_user_session
  get 'sign_out', :to => 'devise/sessions#destroy', :as => :destroy_user_session)
´´´

klappradla:
Is this still relevant / decided yet @michaelem & @bitboxer ?

michaelem:
So, apparently I had that path defined locally when I tried to fix this back in February, sorry for the confusion. Added an additional commit to use the github auth path instead, which is actually defined.
What do you think @carpodaster ?

(If this gets merged, please squash 😉)

carpodaster:
Thank you, @michaelem ! :green_heart:

Issue #657, opened 2017-02-10
Orga dashboard: link to rating dashboard is missing


mkalininait:
It's a bit tricky though.
The orga role has access to the orga dashboard.
The reviewer role has access to the rating dashboard (/rating).
A person who is a reviewer but not an orga won't see the orga dashboard and thus the link to the rating dashboard.
We could create a separate reviewer dashboard and show it on the nav bar (is it a bit excessive?). Or we could give reviewers access to the orga dashboard, but show only one link to the rating dashboard there.

PR #672, opened 2017-02-25
Add link to rating dashboard to user navigation

klappradla:
Hey @mkalininait,

this adds a link to the rating dashboard to the user navigation, as described in #657. The link is displayed throughout all phases, I could not find a specification for something like a *"rating phase"* (cc @alicetragedy).

Be aware of the fact that the current implementation of the applications-table will probably not work with the current DB schema - and therefore break 💥 if you try to access it now.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -79,6 +79,10 @@ class User < ActiveRecord::Base
       where(name: 'organizer')
     end
 
+    def reviewer
+      where(name: 'reviewer')
+    end
+
     def supervisor
       where(name: 'supervisor')
     end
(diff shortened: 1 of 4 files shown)


mkalininait:
Thank you Max @klappradla !

klappradla:
Correct me if I'm wrong @carpodaster, but from what I remember, there were changes to the applications' data model that have not yet been reflected in the rating parts of the app. Therefore I think the big application table is not yet renderable...

klappradla:
If you want me to adjust (in some way) the application table for this, I'd suggest to simply throw in a new issue and assign it to me. For its core functionality, I think this PR could be savely merged before that. No reason to access the applications' table now anyways.

mkalininait:
@klappradla @carpodaster We want to come up with a list of things which need to be adjusted in the rating parts of the app this week. But it's not done yet.

alicetragedy:
+1 on merging this. By beginning of next week we should have a full overview of the changes that need to be made to the “ratings app”, and we will not need to start rating until later in the week. :partyparrot:

PR #683, opened 2017-03-01
[682] Case-insensitive lookup of GitHub handle for role assignment

### LINKED ISSUE #682: Bug: github handle is case sensitive
To reproduce: add a coach to a team. their github username must be capitalised to match, or else the app saves it as blank.

carpodaster:
Fixes #682

### DIFF
--- a/spec/controllers/teams_controller_spec.rb
+++ b/spec/controllers/teams_controller_spec.rb
@@ -170,7 +170,7 @@ RSpec.describe TeamsController do
             expect(User.last.github_handle).to eql github_handle
           end
 
-          it 'finds an existing user by case-insenstive github_handle' do
+          it 'finds an existing user by case-insensitive github_handle' do
             existing_user = create :user, github_handle: github_handle.upcase
             expect {
               patch :update, params: { id: team.to_param, team: valid_attributes }
--- a/spec/controllers/teams_controller_spec.rb
+++ b/spec/controllers/teams_controller_spec.rb
@@ -137,7 +137,7 @@ RSpec.describe TeamsController do
     end
   end
 
-  describe "PUT update" do
+  describe "PATCH update" do
     before { sign_in user }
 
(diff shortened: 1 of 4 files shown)


spec/controllers/teams_controller_spec.rb
```diff
@@ -146,30 +146,52 @@
...
+            expect {
+              patch :update, params: { id: team.to_param, team: valid_attributes }
+            }.to change { User.count }.by 1
+
+            expect(User.last.github_handle).to eql github_handle
+          end
+
+          it 'finds an existing user by case-insenstive github_handle' do
+            existing_user = create :user, github_handle: github_handle.upcase
+            expect {
```
F3PiX: As I understand issue #682, the problem is with lowercase input. Does it make sense to add a spec with `github_handle.lowercase`?

carpodaster: I just wanted to have a different case. I thought of adding a lambda `randomize_case` but decided against it. Do you think it would make the intention of the example more clear? If so, I'll add that.

spec/controllers/teams_controller_spec.rb
```diff
@@ -146,30 +146,52 @@
...
+
+          it 'creates a new user from github_handle' do
+            expect {
+              patch :update, params: { id: team.to_param, team: valid_attributes }
+            }.to change { User.count }.by 1
+
+            expect(User.last.github_handle).to eql github_handle
+          end
+
+          it 'finds an existing user by case-insenstive github_handle' do
```
alicetragedy: s/insenstive/insensitive

spec/controllers/teams_controller_spec.rb
alicetragedy: yeah, I think either randomizing or explicitly setting it to lowercase would make sense. +1 on Maud's suggestion.

Issue #647, opened 2017-02-08
Add link to user (mentor) in single project overview


alicetragedy:
As we discussed some time ago with @carpodaster, it would be useful to have a link to the teams app user who submitted the project (i.e. mentor) on the overview of each project (right now we just show a link to the github profile of the mentor) so that interested students/teams are able to contact them. 

Note: mentors have been informed that they can provide additional / preferred ways of contact, for example a mailing list or link to IRC/slack/gitter, if they want to, which will be added to the project description as a “Note to Students”.

Maybe next year we could think about adding an additional field: “How to contact the mentor” or something like that, so mentors can input their preferred way of being contacted.

mkalininait:
While working on this task I discovered that `@project.mentor` is always null, as long as it is not equal to `@project.submitter`. As we discussed in the dev channel in Slack, we could implement Issue #342 or choose an easier solution and state that project mentors must submit their projects themselves (meaning that we would always count project submitters for primary mentors).

**FYI:** This year (2017) we had 58 project submissions. Three of them have `submitter.github_handle != mentor_github_handle`:
 * [The Processing Foundation](https://teams.railsgirlssummerofcode.org/projects/155-the-processing-foundation) - has 2 mentors
 * [Wikimedia Foundation](https://teams.railsgirlssummerofcode.org/projects/157-wikimedia-foundation)
 * [OpenDF](https://teams.railsgirlssummerofcode.org/projects/165-opendf)

(all have been accepted)

If we choose the easier solution, then the question is what we do with the mentor fields on the project submission form:
* Mentor name
* Mentor github handle
* Mentor email

We could **a) just hide them** or **b) rename them to "Additional mentor name/github_handle/email"** (and then we allow projects to have multiple mentors and students to see them all). On the project page, we could then display:
 * "Submitter / Primary mentor" (we don't display with field at the moment)
 * Additional mentors (what's currently called "Mentor")

That do you think?

mkalininait:
Or, we can ignore my previous comment for the moment and just implement a super quick solution for this task:
 * show links to the Teams app for those projects where `submitter.github_handle = mentor_github_handle` (95% of all projects)
 * leave everything as is (no links) for those projects where `submitter.github_handle != mentor_github_handle` (5% of all projects)

![joxi_screenshot_1486576221858](https://cloud.githubusercontent.com/assets/1745735/22750840/8e4e4962-ee32-11e6-81c8-20b4b160c468.png)
[...]

PR #684, opened 2017-03-02
sets permissions to see the email address in the user profile

mkalininait:
This PR should fix this bug: https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/647#issuecomment-278918784

I didn't write the tests yet, because I don't know how to do it yet, and then I got sick. I decided to make this PR in case someone wants to add the tests and release the bugfix. Otherwise I'll add them to this PR after I recover.

Here is the logic of permissions for the tests:

**email address is hidden:**
 * allow: admin, user's supervisor in current season (must be also a confirmed user)
 * disallow: not admin, not user's supervisor in current season

**email address is not hidden:**
 * allow: admin, confirmed user, user him/herself*
 * disallow: not admin, not confirmed user, not user him/herself

_\* A user sees his/her email address only if it's not hidden, because it looks like a good UX for me: a user can change the "Hide email" option and see the changes on his/her profile._

@carpodaster I was not sure if I should add something to the users controller `/app/controllers/users_controller.rb` It works without adding anything there, too :)


mkalininait:
Hi @ramonh Could you please take a look at the tests I wrote? I'd be very thankful!

Note: Some a/the articles might be missing; English is hard! :D

ramonh:
Sure thing, @mkalininait! Will do this today.

carpodaster:
@mkalininait I have a longish train train coming up today. I'll look into it later!

spec/models/ability_spec.rb
```diff
@@ -25,6 +25,107 @@
...
+              allow(user).to receive(:admin?).and_return(true)
+              allow(ability).to receive(:supervises?).with(other_user, user).and_return(false)
+            end
+            it 'allows to see hidden email address' do
+              other_user.hide_email = true
+              expect(ability).to be_able_to(:read_email, other_user)
+            end
+          end
+
+          context 'when user\'s supervisor in current season (confirmed)' do
```
ramonh: Looks great! Minor stylistic note here: You can use double quote markss around a single quote to avoid having to escape the single quote mark:

`"when user's supervisor in current season (confirmed)"`

spec/models/ability_spec.rb
```diff
@@ -25,6 +25,107 @@
...
 
+      describe 'who is allowed to see email address in user profile' do
+
+        # email address is hidden: admin, user's supervisor in current season (confirmed)
+        # email address is not hidden: admin, confirmed user, user him/herself
+
+        let(:other_user) { FactoryGirl.create(:user) }
+
+        context 'when email address is hidden' do
+          context 'when an admin' do
```
ramonh: Maybe here it would be good to clarify that the user does not supervise the user? What do you think?

ramonh:
Hey @mkalininait, I just added a few minor stylistic notes, but it looks great otherwise!

spec/models/ability_spec.rb
mkalininait: In all other specs we use single quotes for `describe` and `context`. If I change my single quotes to double quotes, it will look inconsistent. Is it ok? Or should I fix the quotes everywhere? (seems to be a nightmare 😃 ) 
Idk, what do you think?

spec/models/ability_spec.rb
mkalininait: Then I kinda need to add it to every context "when a [roleX], but not a [roleY], [conditionZ]?
e.g. "when a confirmed user, but not an admin", "when user's supervisor in current season (confirmed), but not an admin", "when an admin, but may be unconfirmed" etc.

mkalininait:
@carpodaster Carsten, are you planning to look at this PR or should we just merge it? No pressure!

app/models/ability.rb
```diff
@@ -76,4 +84,12 @@ def on_team_for_season?(user, season)
     season && user.roles.student.joins(:team).pluck(:season_id).include?(season.id)
   end
 
+  def supervises?(user, supervisor)
+    is_supervisor = Array.new
+    user.teams.in_current_season.each do |team|
```
carpodaster: Since we are only interested if they are a supervisor in at least one team, the entire method body can be rewritten as:

```ruby
user.teams.in_current_season.any? { |team| team.supervisors.include?(supervisor) }
```

(Disclaimer: written from memory, not tested)

spec/models/ability_spec.rb
carpodaster: (We don't use single or double quotes consistently, so don't bother changing it everywhere. I, too, prefer single quotes but I'll also switch to double quotes when I have to resort to escape sequences)

spec/models/ability_spec.rb
```diff
@@ -25,6 +25,107 @@
...
+            it 'disallows to see not hidden email address' do
+              other_user.hide_email = false
+              expect(ability).not_to be_able_to(:read_email, other_user)
+            end
+          end
+        end
+
+      end
+
+
```
carpodaster: (additional new line)

spec/models/ability_spec.rb
```diff
@@ -81,6 +182,7 @@
 
       end
 
+
```
carpodaster: (additional new line)

app/models/ability.rb
```diff
@@ -12,6 +12,14 @@ def initialize(user)
     can :crud, User, id: user.id
     can :crud, User if user.admin?
 
+    # visibility of email address in user profile
+    can :read_email, User, id: user.id if !user.hide_email?
+    can :read_email, User if user.admin?
+    can :read_email, User do |other_user|
+      (user.confirmed? && supervises?(other_user, user)) ||
```
carpodaster: Since we always require the user to be confirmed, how about:

```ruby
can :read_email, User do |other_user|
  user.confirmed? && (supervises?(other_user, user)) || !other_user.hide_email?)
end
```

mkalininait: Makes sense. Nor sure why I didn't notice this :)

app/models/ability.rb
mkalininait: This looks much cooler 👍

spec/models/ability_spec.rb
ramonh: Hm, good point. In those contexts, it seems like overkill

spec/models/ability_spec.rb
ramonh: Good point! I didn't think of consistency, but rather that, as Carsten said, to avoid escaping quotes

mkalininait:
@carpodaster @ramonh Thanks for the review. I made the changes and tested them. Works fine. 
I just didn't understand where Carsten suggested the additional lines. It looks like they are in place already.

carpodaster:
@mkalininait I meant there were extra lined that should be removed. But it doesn't matter too much, it's just 💅  LGTM!

mkalininait:
@carpodaster Ah, I see :) I thought those extra lines would improve code readability. Will keep this in mind for future PRs!

PR #688, opened 2017-03-08
[364] Add mentor sign-off and fav

### LINKED ISSUE #364: Change mentor sign-off to »mentor fav« (mentor pick)
After the application phase is closed, present the mentors a list of applications referencing their project. They get to pick one team that they favor, maybe because the students have been in contact with them beforehand.

That mentor fav can serve as a soft boon during the review of the applications

carpodaster:
Fixes #364 – see below for screenies.

Feedback about the confirmation dialogues' text blurb most welcome.

Right now the sign-off and the fav is a one-way ticket. One cannot undo this. If an undo-function is desired, I suggest we still merge this first so that mentors can do their thing already.

The feature is currently deployed on staging.

**General overview**

----

**Confirm blurb sign-off**
confirm_signoff
----

**Confirm blurb fav**
confirm_fav
----

### DIFF
--- a/app/controllers/mentor/applications_controller.rb
+++ b/app/controllers/mentor/applications_controller.rb
@@ -15,6 +15,13 @@ class Mentor::ApplicationsController < Mentor::BaseController
     redirect_to action: :index
   end
 
+  def fav
+    @application = Application.find(application.id)
+    @application.update_attribute(:mentor_fav, true)
+    flash[:notice] = "Successfully fav'ed #{@application.team.name}'s application."
+    redirect_to action: :index
+  end
+
   private
 
   def projects
--- a/app/controllers/mentor/applications_controller.rb
+++ b/app/controllers/mentor/applications_controller.rb
@@ -8,6 +8,11 @@ class Mentor::ApplicationsController < Mentor::BaseController
     @comment     = application.find_or_initialize_comment_by(current_user)
   end
 
+  def signoff
+    Application.find(application.id).sign_off! as: current_user
+    redirect_to action: :index
+  end
+
   private
 
   def projects
--- a/app/controllers/mentor/applications_controller.rb
+++ b/app/controllers/mentor/applications_controller.rb
@@ -1,22 +1,21 @@
 class Mentor::ApplicationsController < Mentor::BaseController
+  before_action :application, only: [:show, :signoff, :fav]
+
   def index
     @applications = applications
   end
 
   def show
-    @application = application
-    @comment     = application.find_or_initialize_comment_by(current_user)
+    @comment = application.find_or_initialize_comment_by(current_user)
   end
 
   def signoff
-    @application = application
     Application.find(@application.id).sign_off! as: current_user
     flash[:notice] = "Successfully signed-off #{@application.team_name}'s application."
     redirect_to action: :index
   end
 
   def fav
-    @application = application
     Application.find(@application.id).update_attribute(:mentor_fav, true)
     flash[:notice] = "Successfully fav'ed #{@application.team_name}'s application."
     redirect_to action: :index
@@ -32,7 +31,7 @@ class Mentor::ApplicationsController < Mentor::BaseController
   end
 
   def application
-    Mentor::Application.find(id: params[:id], projects: projects)
+    @application ||= Mentor::Application.find(id: params[:id], projects: projects)
   end
 
   def applications
--- a/app/controllers/mentor/applications_controller.rb
+++ b/app/controllers/mentor/applications_controller.rb
@@ -9,16 +9,16 @@ class Mentor::ApplicationsController < Mentor::BaseController
   end
 
   def signoff
-    @application = Application.find(application.id)
-    @application.sign_off! as: current_user
-    flash[:notice] = "Successfully signed-off #{@application.team.name}'s application."
+    @application = application
+    Application.find(@application.id).sign_off! as: current_user
+    flash[:notice] = "Successfully signed-off #{@application.team_name}'s application."
     redirect_to action: :index
   end
 
   def fav
-    @application = Application.find(application.id)
-    @application.update_attribute(:mentor_fav, true)
-    flash[:notice] = "Successfully fav'ed #{@application.team.name}'s application."
+    @application = application
+    Application.find(@application.id).update_attribute(:mentor_fav, true)
+    flash[:notice] = "Successfully fav'ed #{@application.team_name}'s application."
     redirect_to action: :index
   end
 
--- a/app/controllers/mentor/applications_controller.rb
+++ b/app/controllers/mentor/applications_controller.rb
@@ -9,7 +9,9 @@ class Mentor::ApplicationsController < Mentor::BaseController
   end
 
   def signoff
-    Application.find(application.id).sign_off! as: current_user
+    @application = Application.find(application.id)
+    @application.sign_off! as: current_user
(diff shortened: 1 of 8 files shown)


app/views/mentor/applications/index.html.slim
```diff
@@ -9,15 +9,26 @@ table.applications.table.table-striped.table-bordered.table-condensed.table-hove
...
-        td = application.project_plan
+        td
+          = link_to application.project_name, project_path(application.project_id)
+          - if application.first_choice
+            |&nbsp;#{icon('star')}
+        td
+          - if application.signed_off?
+            = content_tag :em, icon('ok'), class: 'btn btn-disabled', disabled: true, title: "Signed-off at #{l application.signed_off_at}"
+          - else
+            = link_to icon('ok'), signoff_mentor_application_path(application), method: :put, class: 'btn btn-primary', data: { confirm: 'By signing-off this application, you deem this team fit to work on your project and that they meet the requirements. Proceed?' }
```
alicetragedy: maybe we could add something that lets people know that it can't be undone, like “This cannot be undone. Proceed?”

app/views/mentor/applications/index.html.slim
```diff
@@ -9,15 +9,26 @@ table.applications.table.table-striped.table-bordered.table-condensed.table-hove
...
+        td
+          - if application.mentor_fav?
+            = content_tag :em, icon('heart'), class: 'btn btn-disabled', disabled: true, title: "Marked as one of your favorites"
+          - else
+            = link_to icon('heart'), fav_mentor_application_path(application), method: :put, class: 'btn btn-success', data: { confirm: 'By fav\'ing this application, you declare that this would be one of your preferred teams. You can fav more than one application. Proceed?' }
```
alicetragedy: same as with the mentor sign-off, adding a line about it not being undo-able.

alicetragedy:
Somewhat unrelated, but something that came to mind:
if we have applications without a team (i.e. where the team has been deleted because, for example, they wanted to submit a new application), do we show this “old” application to mentors? this could potentially cause an error or be an issue (also, no need for them to review something if it's “void” by default).

app/views/mentor/applications/index.html.slim
carpodaster: Good point! Done.

app/views/mentor/applications/index.html.slim
carpodaster: Good point! Done.

alicetragedy:
Also another thing that came to mind:
Would it make sense to show the mentor their comments in the table as well OR to offer sign-off or fav directly in the show view? Here's my thinking behind this:

I am a mentor and have a shit-ton of applications because my project is the best. I look at the applications and make notes to myself in the appropriate field. However, by the time I'm at the bottom of the list, I honestly can't remember which teams were my favourites, so I have to click through every single application again, look at my comment, then switch back to the index view, and click the fav button.


Anyway, these are more ideas thrown out there, and I think it works well without, it's more something I realised after looking at the (long) list of applications in the screenshots. Judging by our current numbers, it's not quite as bad..

carpodaster:
@alicetragedy it could be that they're already filtered out, if I read [this line of code](https://github.com/rails-girls-summer-of-code/rgsoc-teams/blob/master/app/models/mentor/application.rb#L136) correctly. Maybe @klappradla can verify?

app/controllers/mentor/applications_controller.rb
```diff
@@ -8,6 +8,20 @@ def show
     @comment     = application.find_or_initialize_comment_by(current_user)
   end
```
F3PiX: We have two kind of 'comments' in the project: the polymorphic comments, like for projects etc, and the 'notes', like the notes in the supervisor dashboard. I think this is the note-kind, right? In that case, I'd prefer to reserve the term comment for the 'polymorphic' comments.

carpodaster: Yes, it's just a personal note for the mentor. It's already in production and this PR doesn't target the commenting function. I'm not too invested in whether it's gonna be note or comment. Comment has the bonus of riding on already existing logic.

F3PiX:
Very nice feature. 🍨 🍪 🎉 . 
One question: it looks like a mentor can fav an application without signing it off. Is it worthwhile to require the signing off before enabling fav-ing?

ramonh:
Looks fantastic to me! 👍

carpodaster:
>  it looks like a mentor can fav an application without signing it off.

Hah, yes. I've noticed that after I've uploaded the screenies. I wouldn't enforce it and rely on some common sense (famous last words… 🙃). Not having signed-off an application will rule it out anyway. Otoh, it's not too difficult to implement, but I won't have time for that until tomorrow evening. If we want that extra check, we should probably still merge first and then add it later. Would you like to send a PR for it? 😉 

Related: you could also fav *all* your teams which doesn't make any sense at all.

klappradla:
For what you wrote concerning **deleted** / **non-existing** teams @alicetragedy: @carpodaster is right -> putting together the `Mentor::Application` object requires some values originated by the corresponding team.
So if no `Team`, no `Mentor::Application` ✔️

app/controllers/mentor/applications_controller.rb
```diff
@@ -8,6 +8,20 @@ def show
...
 
+  def signoff
+    @application = Application.find(application.id)
+    @application.sign_off! as: current_user
+    flash[:notice] = "Successfully signed-off #{@application.team.name}'s application."
```
klappradla: Not sure about how smart Rails is about having already queried for the `team`. To play save, I'd suggest to rather use
```ruby
"#{application.team_name}"
```
here, since this won't need an additional query.

Same for the flash in `fav` of course.

carpodaster: 👍

carpodaster:
@klappradla I made the requested changes.

mkalininait:
Looks really cool, I've checked on production (didn't test though).

I also like the comments to this PR.
I think it makes a lot of sense to: 

1. Implement the undo feature (people always want to undo! 😄  )
1. Show comments in the table (as @alicetragedy suggested)
1. Also, I'd love to see mentors' comments during the application rating process (application page, application table).

carpodaster:
@mkalininait thank you for your feedback. I will try to find time to add the undo and comments-in-index view feature tonight.

I'm very much against implementing item no. 3 this year: we've already stated that this would be a comment only for the eyes of the mentor. We would be violating this privacy promise.

alicetragedy:
I feel the same way about item nr. 3 — I don't like the idea of making something “public” which we specifically mark as private.

mkalininait:
@alicetragedy @carpodaster I agree with you, but maybe we can consider making those comments 'public' (available to the reviewers) next year. Let's see what lessons we will learn from this year's selection process.

F3PiX:
My personal vote would be: make comments public, and let a mentor have their own notepad to make notes.
This is part of my ongoing quest to let every web app in the world give you a space to make notes. :donquichote: :windmills:  😂

alicetragedy:
haha :) yeah I actually think giving the mentor the possibility to make personal notes as they go through the applications (especially if those notes are then visible for them in the application list/overview in the mentor dashboard, see https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/692) quite useful. 
Also see the value in them adding their comments to the selection committee about why they're faving teams, so having both would somehow make sense maybe..

PR #696, opened 2017-03-16
Service / task to flag applications for selection

klappradla:
Hey @mkalininait, this is the script you described in [selection#47](https://github.com/rails-girls-summer-of-code/selection_process/issues/47). Due to the _"freeness"_ of the application fields, the matching is rather _"fuzzy"_. It will for instance flag a team as `remote` if the students set these two locations:
```
["New York, United States", "Brooklyn, New York "]
```
The best trade off for me was to match the first and last word, no matter now they are separated. So some false matches will still be in.

Same goes for gender.

I ran it locally on a dump on the current production DB and found the results quite ok. It for instance also flags teams without locations as `remote` - which is fine in my opinion.

Codewise: I know the usage of instance vars within the `find_each` looks kind of nasty for people having seen the light of Elixir. But I think it's still an okish approach for a service object only exposing a single method to the outside.


F3PiX:
Looks great to me! 
I was thinking about flagging remote. In case it helps: We do geolocation, don't we? So we have coordinates? In that case something like a distance less than x would help with NYC and Brooklyn.
Else, maybe geonames can help make the matching less fuzzy. But, this is only trying to be helpful, if you would like to make the matches more precise.

klappradla:
Oh hell @F3PiX you're totally right! I completely forgot I added those coordinate fields 🙈  Thanx a ton for the reminder 🙏 

I should definitely use them for this - will close this one until I updated the matching accordingly.

mkalininait:
Thank you @klappradla! Just curious: How many applications have been flagged when you've run the task locally? Just to get a rough idea of what to expect on prod.

klappradla:
I cannot remember @mkalininait, but I can look it up once I'm on my home machine again.

klappradla:
So thanx to @F3PiX's tip, I changed the uber fuzzy city matcher to use lat-long coordinates instead. I put **70km** as a threshold for what I would consider _"remote"_ - open for suggestions for a better value here 👋 

Another thing about the `remote_team` flag: I skipped checking for the distance between the two locations if there's location data missing and respectively did not set the flag. Since this is coming from Google, it's definitely not the students fault not having it provided. It's **93** applications where this is the case. These are most probably cases where students could not find their location via autocomplete.

mkalininait:
@klappradla Could you please clarify what you mean by "location data missing"? It means that a location is entered but not associated with a Google location, right? Do you think it makes sense to use the distance approach for the applications with two associated Google locations and the name matching approach (what you implemented before) for other applications (since we have so many of them)?

> I put 70km as a threshold for what I would consider "remote"

I don't have a strong opinion about this. I'd say that this should be a distance, which one of the students can make on a **daily** basis, something like 1h of travel one way, because we expect our students to work together. The distance/traveling time ratio obviously depends on where people live, which kind of transportation they use, how busy the traffic is etc. I looked a little bit at this ratio for Delhi and Frankfurt (just two random ish places) and I'd probably reduce the maximum distance to 40-50km or so. Then again, it's just an opinion. What do you think @alicetragedy?

>/\A\s*male\z|\A\s*man\z|\A\s*men\z|\A\s*guy\z/i

Yeah, `boy` is probably a possible option too, I'd add it (I don't know why would anyone use it, but hey, who understands those students :D)

klappradla:
I don't know from the top of my head if **all location data** would be missing, have to look at the production data again tonight. If only lat-lng are missing then falling back to the text-maching would be indeed a good idea.

F3PiX:
It sounds like a quite complicated setup to use the location data, with the fuzzy text matching as a back up. Wouldn't it be cleaner to solve the location data where it is entered?

lib/selection/service/flag_applications.rb
```diff
@@ -0,0 +1,65 @@
...
+
+module Selection
+  module Service
+    class FlagApplications
+      FLAGS           = %w(remote_team male_gender age_below_18 less_than_two_coaches)
+      STUDENT0_COORDS = %w(student0_application_location_lat student0_application_location_lng)
+      STUDENT1_COORDS = %w(student1_application_location_lat student1_application_location_lng)
+      CITY_THRESH     = 100
+      GENDER_FIELDS   = %w(student0_application_gender_identification student1_application_gender_identification)
+      MALE_MATCHER    = ->(a) { a =~ /\A\s*male\z|\A\s*man\z|\A\s*men\z|\A\s*guy\z/i }
```
F3PiX: Can you write the regex like: `\A\s*(male|man|men|guy)\z/i` ? Or would it not be the same?

klappradla: Yes, could be - looks way cleaner actually 🎉 
I'm not a regex-ninja, just quickly hacked this one together...

klappradla:
Ok, I ran a script on production to fetch the missing `lat` and `lng` values. So when I run this task here on the updated production data, it flags **29** applications. I think that' fine. Would be good to merge...

beanieboi:
@klappradla the code looks good to me!

mkalininait:
Thank you Max @klappradla ! Should we now run this task on the production database to flag the applications? Is there anything we should know/prepare before running this task? And could you run it? (I have no idea how to do it yet).

klappradla:
No, the _preparation_ would have been the geo data update and I already ran them in production. I can run the task in production now if you want, the command is approx. like this btw:
```sh
heroku run --app rgsoc-teams-production bundle exec rake selection:flag_applications
```
Do you want me to run it right away?

mkalininait:
@klappradla yes please!

klappradla:
Ok, ran it.

mkalininait:
Great! Thank you!

PR #697, opened 2017-03-16
Add all the favicons

klappradla:
This adds all the favicons


klappradla:
I could also reduce it to just one per platform - but the generator gave me all, so I just took them all 🙈

klappradla:
Hm, wanna do this differently - closing this one again.

alicetragedy:
No problem, just ping me whenever you need a review/feedback @klappradla. 👍

klappradla:
So I first used a generator to create **all the favicons** (based on the .svg file from the design repo). However, after reading a bit more upon favicons, I decided to instead keep it basic. Covering all platforms is barely possible and also not really what I'd consider a useful feature for the teams app.

Having a favicon available in most browsers is an okish goal I'd say, so I skipped everything related to Windows tiles and what not. _"Decent"_ browsers like Chrome use the `.png` one, so there's also a `32x32` version for all those retina and <num>k displays - since I thought they'd be quite common.

The reason why I moved the files in `/assets/images` rather then having them in public: *"control"*. It's even recommended like this in [the Rails docs](http://api.rubyonrails.org/classes/ActionView/Helpers/AssetTagHelper.html#method-i-favicon_link_tag).

bitboxer:
One little note: make sure to have a favicon.ico in the root folder. Or a route that rewrites that to the real location. Otherwise your webserver will serve a 404 for some MS IE versions that have that check hardcoded (they don't look into the html for the correct location :( )

klappradla:
Bummer, but thanx @bitboxer - I thought they'd update their values once it's specified explicitly... But anyways, I'll just copy a fallback one over.

klappradla:
So after realizing I also mixed up branch names and issue numbers, I think it's simply the best to close this for good and start with a fresh one and maybe more sleep. Sorry for the hassles 🙈

Issue #655, opened 2017-02-09
Create a favicon


carpodaster:
Rollbar is full of warnings á la `No route matches [GET] "/favicon.ico"`

Anyone willing to play with a favicon? Personally, I would vote for one that's already rainbow-heart-schemed 😮 `#sneaky` 😉

alicetragedy:
@carpodaster omg yes! finally someone addresses this, I've been thinking about it forever everytime I see the Rollbar warning. We **do** have a favicon.ico already though and it's in `public/images/favicon.ico`. Sorry to rain on your (rainbow) parade 😂 

Not sure why or how it's not being found. Any ideas?

carpodaster:
@alicetragedy oh cool! There is a .png one, the .ico, sadly, is only 0 bytes. The png is very small (16x) and I was told by @axlwaii that you offer several sizes for different devices nowadays. Does anyone have an SVG version of the Rails Girls pixel heart? Or will it be a rainbow parade after all? 😉

alicetragedy:
@carpodaster in that case, it makes sense to create one, apologies! 🙈 
we do have the heart logo as an illustrator file I think ([link](https://github.com/rails-girls-summer-of-code/design/tree/master/logo)); I can look into making it an svg and pushing it so that whoever is interested in this task can take it :)

axlwaii:
http://www.favicon-generator.org/ does a great job creating all favicons you need :)

alicetragedy:
Update: an SVG of our current logo can be found [here](https://github.com/rails-girls-summer-of-code/design/blob/master/logo/rg_heart_logo.svg); @carpodaster feel free to add this in your original comment for anyone who wants to take over this task :)

anasofiapinho:
Hey everyone! I did the favicons of our logo in red and also rainbow colors because I was confused about what you wanted. Thanks @axlwaii for your tip about that website!

### Red logo favicon
The folder is [here](https://github.com/rails-girls-summer-of-code/design/blob/master/logo/favicon-rgsoc-logo-red.ico.zip).

### Rainbow logo favicon

The folder is [here](https://github.com/rails-girls-summer-of-code/design/blob/master/logo/favicon-rgsoc-logo-rainbow.ico.zip).  

### HTML code

In case you wanted, I added the code they suggested here:

```html
<link rel="apple-touch-icon" sizes="57x57" href="/apple-icon-57x57.png">
[...]

PR #698, opened 2017-03-18
Add / update favicons (heart icon)

klappradla:
3rd approach on an accurate PR trying to solve #655 .

I generated the favicon based on the `.svg` icon from the design repo. It's in my opinion out of scope for the Teams App trying to support *all* use cases for favicons (e.g. Windows Tiles, Android shortcuts, what not) while staying maintainable. However, since _"normal"_ browsers use the `.png` version and lots of people have retina displays, I at added a `32x32` one, since this already seems to be a standard use case.

Thanx to Bodo's hint, I also added a `.ico` version to the root (in `public/`) in order to allow old browsers that don't look at the HTML tags to get the favicon as well.


klappradla:
Anything to add upon this? Should I use more favicons or update to the rainbow one? If not, I'd merge it rather soon, keeping it up to date while lingering around is a bit of unnecessary work 😉

klappradla:
No worries, didn't want to stress anyone. Just thought it does not really make sense to have it lying around if there's nothing to add. Will do a merge party later today :v:

beanieboi:
@klappradla awweeesooomme 

awesome

PR #699, opened 2017-03-18
Fix sign off / fav per project

carpodaster:
This implements the fix w/o adapting the tests.

It also does not fix the existing data.

See Slack.

### DIFF
--- a/app/models/mentor/application.rb
+++ b/app/models/mentor/application.rb
@@ -5,8 +5,8 @@ module Mentor
     attr_accessor :id, :team_name
     attr_accessor :project_id, :project_name
     attr_accessor :project_plan, :why_selected_project
-    attr_accessor :signed_off_by, :mentor_fav
-    attr_reader   :signed_off_at
+    attr_accessor :signed_off_by
+    attr_reader   :signed_off_at, :mentor_fav
     attr_accessor :student0, :student1
     attr_accessor :first_choice
 
@@ -31,10 +31,26 @@ module Mentor
         user:             mentor)
     end
 
-    alias mentor_fav? mentor_fav
+    def mentor_fav?
+      !!mentor_fav
+    end
+
+    def mentor_fav=(value)
+      @mentor_fav = %w(1 true t).include?(value)
+    end
 
     def signed_off?
       !!signed_off_at
+    end
+
+    def mentor_fav!
+      self.mentor_fav = persisted_application.application_data["mentor_fav_project#{choice}"] = true
+      persisted_application.save!
+    end
+
+    def revoke_mentor_fav!
+      self.mentor_fav = persisted_application.application_data["mentor_fav_project#{choice}"] = false
+      persisted_application.save!
     end
 
     def sign_off!(as: nil)
--- a/app/models/mentor/application.rb
+++ b/app/models/mentor/application.rb
@@ -8,11 +8,11 @@ module Mentor
                   :project_plan, :why_selected_project,
                   :signed_off_by,
                   :student0, :student1,
-                  :first_choice
+                  :choice
     attr_reader :signed_off_at, :mentor_fav
 
-    def choice
-      first_choice ? 1 : 2
+    def first_choice?
+      choice == 1
     end
 
     def student0=(attrs)
@@ -165,7 +165,7 @@ module Mentor
           application_data -> :signed_off_at AS signed_off_at,
           (application_data -> :signed_off_by)::int AS signed_off_by,
           application_data -> :mentor_fav AS mentor_fav,
-          CASE WHEN :project_id::text = 'project1_id' THEN TRUE ELSE FALSE END AS first_choice
+          CASE WHEN :project_id::text = 'project1_id' THEN 1 ELSE 2 END AS choice
           FROM applications
           INNER JOIN teams
           ON teams.id = applications.team_id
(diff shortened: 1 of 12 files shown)


klappradla:
So I added the missing things to this PR as specified by @carpodaster on Slack:
* adapt / add tests for new structure now `fav` and `sign_off` are handled
* add a single-run task to *"fix"* the existing data

The data is *"fixed"* in a sense that **sign-offs are migrated** and **favs are simply dropped**. However, this does not really mean that the state of `sign_offs` is perfectly fine: before, a `sign_off` was per `application`, now it is per `project` with in the application`.

As far as I can reconstruct this mentally, this would mean, if previously **two** mentors signed an application - each for their project - only on of the `sign-offs` (the time-wise most recent one) is kept. The other one cannot be reconstructed from the data we have.

There was also a bug-report for mentors' comments / notes - but that is already fixed an no data has to be migrated.

Since @carpodaster is on an island 🏝, maybe @bitboxer, @pgaspar or @ramonh could have a look this codewise?

spec/models/mentor/application_spec.rb
```diff
@@ -207,26 +208,181 @@
...
+    subject { m_application.sign_off!(as: mentor) }
+
+    shared_examples :a_mentor_sign_off do |choice|
+      let(:other) { (choice % 2) + 1 }
+      let(:now)   { Time.now.utc.to_s }
+
+      before { Timecop.freeze(now) }
+      after { Timecop.return       }
+
+      it 'add the signoff time and user id to the application' do
```
pgaspar: Maybe change "add" to "adds" to conform with the "It does something" form, here?

app/models/mentor/application.rb
```diff
@@ -112,7 +162,10 @@ def sql_statement_all
           teams.name AS team_name,
           projects.name AS project_name,
           (application_data -> :project_id)::int AS project_id,
-          CASE WHEN :project_id::text = 'project1_id' THEN TRUE ELSE FALSE END AS first_choice
+          application_data -> :signed_off_at AS signed_off_at,
```
beanieboi: we might consider moving the existing indexes from the old columns to the new columns. I don't think it's urgent but we should not forget about it, since it can has an impact on the application performance.

old indexes
```
application_data -> project1_id
application_data -> project2_id
signed_off_by
```
possible new indexes, just theoretical, we should look at the query in production (ideally development staging :D)
```
application_data -> :signed_off_at
application_data -> :signed_off_by
application_data -> :mentor_fav
```

db/migrate/20170319180638_remove_mentor_fav_from_applications.rb
```diff
@@ -0,0 +1,9 @@
+class RemoveMentorFavFromApplications < ActiveRecord::Migration[5.0]
+  def up
+    remove_column :applications, :mentor_fav
```
beanieboi: I'm not super familiar with the data in the database. just as an idea.
should we can keep this for a while? maybe it's helpful to partially backfill the data?
if yes, I suggest renaming it to do `deprecated_mentor_fav` or something similar.

spec/controllers/mentor/applications_controller_spec.rb
```diff
@@ -126,40 +126,47 @@
...
         context 'undo\'ing the sign-off' do
-          before { application.sign_off! as: user }
+          before { m_application.sign_off! as: user }
 
-          it 'sets the sign-off timestamp' do
-            expect { put :signoff, params: { id: application.id } }
-              .to change { application.reload.signed_off_at }.to nil
+          it 'resets the sign-off timestamp' do
+            expect { subject }
+              .to change { application.reload.application_data['signed_off_at_project1'] }
```
beanieboi: should we test this also for `signed_off_at_project1 `?

spec/controllers/mentor/applications_controller_spec.rb
```diff
@@ -126,40 +126,47 @@
...
+              .to nil
           end
 
-          it 'set persists the mentor who signed-off' do
-            expect { put :signoff, params: { id: application.id } }
-              .to change { application.reload.signatory }.to nil
+          it 'resets the mentor who signed-off' do
+            expect { subject }
+              .to change { application.reload.application_data['signed_off_by_project1'] }
```
beanieboi: should we test this also for `signed_off_at_project2`?

klappradla:
Thanx for all the _(and quick)_ feedback 🎉
I'll try to address all this once I get home from work. It it's super urgent and this should have been deployed basically yesterday, someone else need to take over here, sorry.

db/schema.rb
```diff
@@ -83,7 +83,6 @@
     t.integer  "signed_off_by"
     t.datetime "signed_off_at"
     t.integer  "project_id"
-    t.boolean  "mentor_fav"
```
klappradla: Sorry @beanieboi, I don't really get it - where's the newly named column now? 🙈

beanieboi:
I forget to update schema.rb..... stupid me :)
Will update asap.

db/schema.rb
klappradla: If I pull the changes from master and migrate the DB it's there - that's what i wanted to say 😉

klappradla:
Oh, no worries @beanieboi, I'll update it since I'm already on it and merging back master into my other branches. Was just not sure if you finally wanted the column there or not :v:

mkalininait:
Checked on Prod. The issue seems to be fixed.

PR #700, opened 2017-03-19
Deprecate sign off columns

klappradla:
Ok, this would be ready to merge now.
I aligned this to go the @beanieboi  - way™ and just rename the deprecated columns, so that we can remove them all together after the selection phase.


klappradla:
Good to merge in my opinion then. Would do so tonight if no obligations.

PR #701, opened 2017-03-19
Update dependencies

klappradla:
After having certain moments of enlightment with updating dependencies in other apps, I had a look at the Teams App's dependencies.

Rails 5.1 is around the corner and certain deprecated things the Teams App is using / doing won't be possible anymore 😉  - such as, e.g
* non-kwargs syntax for controller specs
* have stubs access the DB

Also: providing the `valid_session` with Devise-related specs is no longer recommended since Rails 4 - there's no benefit in doing it now IIRC.

Still, this is WIP, I just pushed it in order to be able to continue on it from other machines and be reminded that it's still open 😉


klappradla:
Ok, so this would be ready to be merged. And since this PR touches a lot of files, I'd also love to merge it within not too long, because it's kind of pain to keep it up 2 date otherwise 😉 

I skipped updating `aws-sdk` and `feedjira` for now, it seemed out of scope for this PR to me: both gems a) interact with each other and b) are so outdated that there API changed significantly since then. I'd volunteer to update them and their integration into the Teams App, but I'd open up a separate issue for that and tackle this later someday in a separate PR.

spec/controllers/mailings_controller_spec.rb
```diff
@@ -42,15 +41,15 @@
     end
 
     context 'with user logged in' do
-      let(:user) { FactoryGirl.create(:student) }
```
beanieboi: is there a technical reason for that? does FactoryGirl not support the syntax anymore? the reason why I ask is that I really like the explicit method call. it obvious the everybody that this is `create()` from FactoryGirl. you don't need knowledge about the dependencies that are used in the app.

klappradla: **HUGE** ➕  **1** on that!
I removed it in a few places already 😉

beanieboi: yay and nay! 👯‍♂️

klappradla: Feel free to go ahead on it. I'm on a dirty branch right now and will stay there for some time as it seems ✌️

beanieboi: cool. will do it.

beanieboi:
@klappradla I updated the PR. it would be great if you can merge since I don't have access to Rollbar. dependency upgrade are usually good candidates for exceptions :)

klappradla:
Alright, will do in a few min. Thanx for the review 🙏

beanieboi:
👏🏼🤙🏾

PR #702, opened 2017-03-20
Cleanup/remove application sign-off

klappradla:
As @pgaspar already pointed out in #699 , there is a lot of unused / duplicated code related to signing-off `Applications` and `ApplicationDrafts` causing confusion for anyone not familiar with previous iterations of the Teams App's internal processes.

* Anything related to signing-off `Applications` is now moved over to the `Mentor::Application` model
* Signing-off `ApplicationDrafts` is not used any more _(none of the `ApplicationDrafts` from this season was signed-off and the code is not invoked from anywhere)_

Therefore, in order to not grow legacy, I think it should be fine to remove this logic and deprecate the attributes @beanieboi -way until the selection process is over and they can be fully removed.


klappradla:
Anyone with domain-knowlege of the current selection process wants to have a quick look at this( ping @alicetragedy , @ramonh )? Aside from that I think this should be good to merge then.

alicetragedy:
this makes sense to me @klappradla (in the context of the current sign-off and selection process) and it should be good to go. 👍

Issue #612, opened 2017-01-14
User Profile: tech fields for all users


mkalininait:
This issue originates from this one: https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/593

We would like to have more information about user technical skills primarily to make it easier for students and coaches to find each other.

**Suggested solution:**
Add the following fields (all required):
- "Technologies of expertise” - tags separated by comma (similar to the "Tags" field in Project)
- "Technologies of interest” - tags separated by comma (similar to the "Tags" field in Project)

Anagha22:
Hey, we are team Bundledore, we would like to work on this issue.

Anagha22:
I have submitted a PR for this issue . Please suggest changes if any.

Anagha22:
I have submitted the PR with all the changes mentioned. Please review it.

mkalininait:
Hi @Anagha22 Both of your PRs are closed. Are you going to make another one? I'm confused.

Anagha22:
@mkalininait Yeah, sorry for the delay, I got some technical issues. It's fixed now. Will send a PR soon.

branquinhoaa:
I'm not sure if this will be done by @Anagha22 or if I can take it. It is part of the community milestone, I would like to do if she won't. :smile:

branquinhoaa:
@klappradla  do you think this issue can be part of our community milestone?

klappradla:
Sure 👍  Good point!

branquinhoaa:
@mkalininait do you want to show the technologies fields in the user page? or this info will be there just for search?

klappradla:
Regarding the layout I already left some comments in #863 

For whether or not the fields should be mandatory:
I see some difficulties here...
1. From a technical perspective: it would make all existing user profiles invalid 🙈. I know we had something similar with the "city" or "country" field one, but I don't remember how we solved it...
2. From a human perspective: we may eventually insult or scare off groups of people unsure about the level of their knowledge and experience by "forcing" them to come up with some technologies and languages they are experienced in...

Maybe we have to come up with "softer" approaches to get this data. I could for instance imagine adding a reminder to fill out the profile information into the confirmation email or showing a friendly reminder when a user without a filled out profile logs in... there are several possibilities.

klappradla:
Closed via #863

PR #703, opened 2017-03-21
User Profile: tech fields for all users 

Anagha22:
Adding a PR for issue #612. Please review and suggest changes.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170319180638) do
+ActiveRecord::Schema.define(version: 20170321150557) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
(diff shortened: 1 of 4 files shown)


db/schema.rb
```diff
@@ -291,11 +291,13 @@
...
     t.float    "application_location_lng"
     t.text     "application_giving_back"
-    t.text     "application_diversity"
     t.string   "confirmation_token"
     t.datetime "confirmed_at"
     t.datetime "confirmation_sent_at"
     t.string   "unconfirmed_email"
+    t.text     "application_diversity"
+    t.string   "Expertise"
+    t.string   "Interests"
```
F3PiX: Question for reviewers: can the capitals stay?

alicetragedy: I feel like it's no big deal, but I'd prefer for this to be consistent (so: no capitals). @Anagha22 could you change this, please?

db/schema.rb
```diff
@@ -10,24 +10,24 @@
...
-ActiveRecord::Schema.define(version: 20170319180638) do
+ActiveRecord::Schema.define(version: 20170321150557) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
   enable_extension "hstore"
 
   create_table "activities", force: :cascade do |t|
     t.integer  "team_id"
-    t.string   "kind",         limit: 255
```
alicetragedy: I'm not sure why the `limit: 255` for most fields was removed in the schema file — I'm guessing this might have happened automatically with your newly added migration.I don't know how badly we need those limits on fields, but since this is not relevant to the PR, could you change all of them back to their previous state? (i.e. add `limit: 255` to the fields again where it's been removed). Thanks :)

db/schema.rb
Anagha22: @alicetragedy Okay!

F3PiX:
@Anagha22 Closing for now. Please feel free to reopen when you are ready to address the changes.

PR #704, opened 2017-03-22
Add volunteering-related columns to application table

ramonh:
As discussed for the new selection process, this adds the "Volunteer?" and "Volunteer hours" columns to the application rating table.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -4,6 +4,10 @@ module ApplicationHelper
 
   TIMEZONES = ActiveSupport::TimeZone.all.map{|t| t.tzinfo.name}.uniq.sort
 
+  def format_boolean(boolean)
+    boolean ? 'Yes' : 'No'
+  end
+
   def show_application_link?
     current_season.application_period? ||
       (Time.now.utc.between? current_season.applications_close_at, current_season.acceptance_notification_at)
--- a/app/views/rating/applications/_application.html.slim
+++ b/app/views/rating/applications/_application.html.slim
@@ -13,6 +13,8 @@ tr class=application_classes_for(application)
   td.coaching_company
     = application.coaching_company
   td = format_application_money(application)
+  td = application.application_data["voluntary"]
+  td = application.application_data["voluntary_hours_per_week"]
   td = application.total_picks
   td = rounded_points_for application
   - row.ratings.each do |rating|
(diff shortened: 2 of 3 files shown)


alicetragedy:
Something I just thought of when looking at this: do you think it would make sense to change the boolean values (`True` & `False`) to display as something like "yes" and "no" instead (using a locale or a helper method)? It's definitely not a requirement, more like an enhancement. :)

ramonh:
Hey @alicetragedy, thanks for the feedback! 

I thought of this too, but wasn't sure what would be preferred. `Yes` and `No` sound good to me though. Will add this today 👍

alicetragedy:
awesome, thanks @ramonh! And I think it's definitely something we can reuse — I took a quick look to see if we had anything like that but it doesn't look like it, and it might be useful for the [conferences table](https://github.com/rails-girls-summer-of-code/rgsoc-teams/blob/master/app/views/conferences/_table.html.erb#L21) too. ;)

ramonh:
Great idea! I'll make it a huge scoped one then :)

alicetragedy:
@ramonh did you get a chance to look at the helper method enhancement thing I suggested yesterday? or, better said: is there anything we can help you with there? I have some time today and over the weekend so am happy to jump in if needed <3

ramonh:
Hey @alicetragedy! I've got it ready to go, just had to step out a bit earlier. Will commit and push!

ramonh:
All set! Thanks for the patience.

alicetragedy:
@ramonh I'm sorry, but I wasn't thinking straight... I realised that application_data["voluntary"] actually returns a string, not a boolean, and so `format_boolean` will always show “Yes” in this case. The quickest way I can think of for fixing this is a conditional checking for the content of the value (ie. if true return yes) but maybe it's kind of "dirty"? what do you think?

ramonh:
Oh dear, good catch, @alicetragedy. I didn't think this through 😭 

I had initially misunderstood what you wrote, but I think that's a good idea! I don't think it's too dirty, because we're ensuring that we're passing the correct value type.

ramonh:
And done! Thanks for the feedback again :)

mkalininait:
@klappradla Do you think we can merge it (after solving the merging conflict) or does it not work well with your new application table model?

klappradla:
Hm, it's ok to merge it, but the benefit is not huge unfortunately (sorry @ramonh), since the two files are basically gone in the new view 🙈

ramonh:
Hey! No worries. If I understand correctly then, we should close this?

klappradla:
Yes @ramonh. Resolving the conflicts and then merging to master just for removing the files after all is not worth the effort - you know 😉

ramonh:
Fair point, thanks @klappradla! I'll catch up and try again

PR #705, opened 2017-03-23
additional info form changes

alicetragedy:
This fixes the first four points in [this issue](https://github.com/rails-girls-summer-of-code/selection_process/issues/52): removing some fields in the additional info form and changing flags as needed.

Note: I've also removed this attribute from the permitted list of params in the controllers, and from the Applications exporter. I'm not sure what the policy is on removing fields from the database though, so I'd love an opinion there: should I remove the `project_visibility` field from the `applications` table?

### DIFF
--- a/app/controllers/orga/applications_controller.rb
+++ b/app/controllers/orga/applications_controller.rb
@@ -71,7 +71,7 @@ class Orga::ApplicationsController < ApplicationController
 
   def application_params
     if params[:action] == 'update'
-      params.require(:application).permit(:misc_info, :project_visibility,
+      params.require(:application).permit(:misc_info,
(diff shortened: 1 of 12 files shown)


alicetragedy:
Note to self: `project_visibility`, since we don't need it at all, should also be removed from e.g. the list of permitted params in the `application_controller`. Not sure if it should also be removed from the db (we have other unused fields, like `sponsor_pick`, in the db, which are now not used anymore).

alicetragedy:
PR has been updated. I'm not sure what the policy is on removing fields from the database though, so I'd love an opinion there: should I additionally remove the `project_visibility` field from the `applications` table?
ping @carpodaster @klappradla @michaelem

klappradla:
I'd vote for removing it 👍

PR #675, opened 2017-02-26
Fixes github auth for users, closes #673

bitboxer:
Okay, this changes a few things in the OmniAuth flow.

It will redirect the user to the profile edit page when

* the user is freshly imported from github
* the user is not valid because data is missing

Additionally it shows a warning box if you haven't confirmed
your account yet with a button to confirm it manually.

PR #713, opened 2017-03-27
Fixes github auth for users, closes #673

### LINKED ISSUE #673: User confirmation: the "Resend" button
When user email address is not confirmed, we should display a warning about this with the “Re-send confirmation” button which triggers sending an email with a confirmation token.
**UPD from Bodo:** "we can add that to all pages to the top right below the navbar. that way no user can say that they haven’t seen it :slightly_smiling_face:"

**Context:** many users don’t get confirmation emails. In some cases we send them, but they don't reach users and in some cases we don’t even send them. I can’t say in which cases specifically we don't send emails, this needs to be tested thoroughly. But the Resend button would be a sufficient hot-fix for such cases for the moment.

beanieboi:
Okay, this changes a few things in the OmniAuth flow.

It will redirect the user to the profile edit page when

* the user is freshly imported from github
* the user is not valid because data is missing

Additionally it shows a warning box if you haven't confirmed
your account yet with a button to confirm it manually.


beanieboi:
this is a new PR that addresses the issue from https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/675

app/models/user.rb
```diff
@@ -112,7 +112,14 @@ def student
...
   # a preliminary user, e.g. when adding a non existant person
   # to a team using the github handle.
-  attr_accessor :github_import
+  attr_reader :github_import
+
+  # This informs devise that this user does not
+  # need a confirmation notification for now
+  def github_import=(import)
+    @github_import = import
+    skip_confirmation_notification! if @github_import
```
carpodaster: One very minor thing: I would swap the two lines so that it returns `@github_import`; one would expect an assignment to return its value.

beanieboi:
> I am wondering: is there any way a User record comes to life without github_import = true?

good question. sadly I can't answer it, because of my lack of knowledge of the codebase. but I would love to know the answer :)

app/models/user.rb
beanieboi: fixed

app/helpers/authentication/active_record_helpers.rb
```diff
@@ -6,32 +6,41 @@ def self.included(base)
 
     module ClassMethods
       def find_or_create_for_github_oauth(auth)
-        find_for_github_oauth(auth) || create_for_github_oauth(auth)
+        find_and_update_for_github_oauth(auth) || create_for_github_oauth(auth)
+      end
+
+      def find_and_update_for_github_oauth(auth)
+        user = find_for_github_oauth(auth)
+        update_user(user, auth) if user && user.email.nil? && user.unconfirmed_email.nil?
```
beanieboi: I used `unconfirmed_user?` for now. since I'm not 100% sure what this check is for. 
@bitboxer could you help me out with the naming here?

app/models/user.rb
```diff
@@ -117,8 +117,8 @@ def student
   # This informs devise that this user does not
   # need a confirmation notification for now
   def github_import=(import)
-    @github_import = import
     skip_confirmation_notification! if @github_import
```
carpodaster: Now we need to change `@github_import` to ref. the local variable `import`

beanieboi:
@carpodaster updated the PR accordingly

carpodaster:
@beanieboi shall we merge this?

beanieboi:
I'm up for merging!

alicetragedy:
yes, do it 🙌 🎉 🎈

PR #715, opened 2017-03-27
Remove Student location from application view

beanieboi:
this PR takes care of removing the location data from the application as requested in https://github.com/rails-girls-summer-of-code/selection_process/issues/51


beanieboi:
totally agree, we should have 1 place where we define all available fields and which fields we should display. I honestly see little value that we automatically show new fields. I'd prefer that we explicitly add fields that we want to show

alicetragedy:
+1 in what you both said — I think @mkalininait also agrees with this.

klappradla:
Yes. I wouldn't even go with a view object, etc., but just use a plain old Rails view rendering out the things. Same as I did for the mentor view on applications. I think this one is okish regarding readability.

app/models/application.rb
```diff
@@ -119,11 +119,18 @@ def data_for(role, subject)
     Data.new(application_data, role, subject).extract || {}
   end
 
-  def sorted_application_data
+  def application_data_for_view
```
mkalininait: 👍 👍 👍

mkalininait:
@alicetragedy oh yes, I do!!

PR #716, opened 2017-03-27
WIP: selection table

klappradla:
Hey @mkalininait as promissed: a wip version of the selection table, filters, ordering, etc. not working.


mkalininait:
First read it as: "selection table, filters, ordering, etc. [all] not working." 

Gonna test now. Thank you so much @klappradla!

PR #717, opened 2017-03-28
show student names

beanieboi:
Request: https://github.com/rails-girls-summer-of-code/selection_process/issues/51

Show the name of Student 1 and Student 2.

### DIFF
--- a/app/helpers/applications_helper.rb
+++ b/app/helpers/applications_helper.rb
@@ -34,9 +34,18 @@ module ApplicationsHelper
       project = Project.find_by_id value
       link_to_if project, project.try(:name), project
     else
-      auto_link simple_format(value)
+      auto_link(simple_format(value))
     end
     content_tag :p, formatted.html_safe
+  end
+
+  def show_student_names(application)
+    output = ""
+    output << content_tag(:b, Application.data_label("student0_name"))
+    output << content_tag(:p, simple_format(application.team.students.first.name))
+    output << content_tag(:b, Application.data_label("student1_name"))
+    output << content_tag(:p, simple_format(application.team.students.last.name))
+    output.html_safe
   end
 
   def format_application_location(application)
--- a/app/helpers/applications_helper.rb
+++ b/app/helpers/applications_helper.rb
@@ -18,34 +18,8 @@ module ApplicationsHelper
     classes.join(' ')
   end
 
-  def formatted_application_data_value(key, value)
-    markdown_fields = %w(project_plan1 project_plan2)
-    value = value.presence || 'n/a'
-    formatted = case
-    when markdown_fields.include?(key.to_s)
-      render_markdown value
-    when /student._application_coding_level/ =~ key.to_s
-      value + '/5'
-    when /student._application_language_learning_period/ =~ key.to_s
-      value + ' months'
-    when /student._application_money/ =~ key.to_s
-      value + ' USD'
-    when /project._id/ =~ key.to_s
-      project = Project.find_by_id value
-      link_to_if project, project.try(:name), project
-    else
-      auto_link(simple_format(value))
-    end
-    content_tag :p, formatted.html_safe
-  end
-
-  def show_student_names(application)
-    output = ""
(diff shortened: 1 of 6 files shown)


mkalininait:
Testing in progress, please don't merge yet.

spec/controllers/rating/applications_controller_spec.rb
```diff
@@ -49,8 +50,11 @@
       end
     end
   end
+
   describe 'GET show' do
-    let(:application) { create :application }
+    let(:application) { FactoryGirl.create(:application) }
+    let!(:student0) { FactoryGirl.create(:student, team: application.team) }
```
klappradla: This is just personal preference, but I don't see the point it writing `let` blocks if they aren't actually used in the test itself. The fact that one needs to use the bang version since the lambda is otherwise never being called is also kind of an indicator for me that these `create` calls should either go into a `before` block or the example itself and do not need to be assigned with names.

However, just a style preference 💅

app/views/rating/applications/show.html.slim
```diff
@@ -43,6 +43,8 @@
         td = format_application_flags(@application)
 
   h2 Submitted Application
+  = show_student_names(@application)
```
klappradla: I personally don't see a huge benefit in outputting **both** names in the beginning and then just throwing in the whole application. In my opinion it would make more sense to have a layout like this:
```
h3 student0 name
panel
  heading question 1
  body answer 1
panel
  heading question 2
  body answer 2
h3 student1 name
panel
  heading question 1
  body answer 1
panel
  heading question 2
  body answer 2
```

But this might not be in the scope of this PR. Just came to my mind when I saw this line.

app/helpers/applications_helper.rb
```diff
@@ -34,11 +34,20 @@ def formatted_application_data_value(key, value)
...
-      auto_link simple_format(value)
+      auto_link(simple_format(value))
     end
     content_tag :p, formatted.html_safe
   end
 
+  def show_student_names(application)
+    output = ""
+    output << content_tag(:b, Application.data_label("student0_name"))
+    output << content_tag(:p, simple_format(application.team.students.first.name))
```
klappradla: I'm not working with Rails on a daily basis any more, but from what I remember, `simple_format` will already return a paragraph anyways. So you could reduce this to:
```ruby
# ...
output << content_tag(:b, Application.data_label("student0_name"))
output << simple_format(application.team.students.first.name)
# etc.
```

beanieboi: yes, you are right, I just copy & pasted from the current implementation. it's fixed in the unrolled partial.

app/views/rating/applications/show.html.slim
beanieboi: I just unrolled the current structure. it's now way easier to make those kind of changes. I won't include them in this PR to keep it focused on the core issue.

mkalininait:
Tested locally, looks good to me!

app/views/rating/applications/_application_data.html.slim
```diff
@@ -0,0 +1,142 @@
+- ad = application.application_data
```
klappradla: Depending on how easy it is to handle this large file, one could also extract a `_student` partial here. Would require string-interpolating the keys however...

app/views/rating/applications/_application_data.html.slim
```diff
@@ -0,0 +1,142 @@
...
+b = Application.data_label("student0_application_diversity")
+p = auto_link(show_or_na(ad["student0_application_diversity"]))
+
+b = Application.data_label("student1_application_diversity")
+p = auto_link(show_or_na(ad["student1_application_diversity"]))
+
+b = Application.data_label("student0_application_location")
+p = auto_link(show_or_na(ad["student0_application_location"]))
+
+b = Application.data_label("student1_application_location")
```
klappradla: Maybe add a separator / some headline between these sections (student0, student1, etc.)

beanieboi: great idea, added a horizontal line between all of them

PR #721, opened 2017-03-29
bump dependencies

beanieboi:
regular dependency bump


lib/feed/s3.rb
```diff
@@ -2,30 +2,32 @@ class Feed
...
+          access_key_id: options[:access_key_id], secret_access_key: options[:secret_access_key]
+        )
       end
     end
 
     attr_reader :s3, :url, :logger
 
     def initialize(url, options = {})
-      @s3 = AWS::S3.new
+      @s3 = Aws::S3::Resource.new(stub_responses: Rails.env.test?)
```
carpodaster: Cool, I wasn't aware of that!

lib/feed/s3.rb
```diff
@@ -2,30 +2,32 @@ class Feed
   class S3
     class << self
       def setup(options)
-        AWS.config(access_key_id: options[:access_key_id], secret_access_key: options[:secret_access_key])
+        Aws.config.update(
+          access_key_id: options[:access_key_id], secret_access_key: options[:secret_access_key]
```
klappradla: What @carpodaster said about the region could also go in here from what I remember.

beanieboi:
remove the S3.setup call, since the `aws-sdk` will pick those ENV vars automatically (see http://docs.aws.amazon.com/sdkforruby/api/#Configuration)

beanieboi:
fyi: @klappradla did set AWS_REGION on the Heroic apps to `us-east-1`

PR #724, opened 2017-03-30
Rating Table

klappradla:
Update of the rating table according to the specs in rails-girls-summer-of-code/selection_process#53 (still missing the median column though). I also added a CSV-export, since this seems to be a common use case. Note that the export always exports **the exact same data that's available in the table** - so it takes all filters and sorting, just like the table itself.

On the way, I also removed unnecessary view helpers, etc. and generally tried to reduce the usage of (global) view helpers.

There's still some more cleanup to, but this is at least a working version to get the selection process started.
The PR is pretty huge - sorry for that, but I did not know how to sanely split this into multiple smaller ones.


app/models/application.rb
```diff
@@ -149,4 +136,12 @@ def sponsor_pick?
       value.to_s != '0' ? flags.concat([flag.to_s]).uniq : flags.delete(flag.to_s)
     end
   end
+
+  def project1
+    Project.find_by(id: application_data['project1_id'])
```
klappradla: How do you imagine this? _"Just"_ as a data transfer object or as a model persisted to the database?

But related to that: it's super hard to render these types of application overview views. I'd need to preload the teams and projects associated with an application in order to make this more sane (snappy would be the wrong word). However, I only know how to do this in raw sql...

beanieboi: I would say as a data object first. if we have the data object, moving it to the database as persisted is already half way there. since we can keep the same abstraction. I will work on a proposal for this today, to better show what I mean :)

klappradla: Ok, just to quickly clearify my understanding of DTOs. The way I usually use them, they are POROs taking KWARGS and then internally just delegate to the underlying objects. So they are basically just a wrapper around objects that are passed around in business-logic as this one pack which then can be easily stubbed in specs.

So they usually just have a few private `attr_reader` and some large lists of `def_delegators` 😉 , no *"real"* methods.

But yeah, curios of what you have in mind here.

klappradla:
I'd love to merge this soon if there are no obligations.

mkalininait:
Hey @klappradla I've tested the PR locally.
The table looks great! 
I found a couple of issues with the csv export.

1. When the table is already exported, I continue seeing this button:
joxi_screenshot_1490999288291
I can make a new export only after I refresh the page.

2. I think something went wrong with the last table columns: I'm not sure what Picks1,2 and Likes 1,2 mean; what the next two columns mean (the average rate for one team is 1.55, but there is no such number in the exported table); the last two columns don't have a header.

screen shot 2017-04-01 at 00 31 19
screen shot 2017-04-01 at 00 31 56

---

Another thing: these changes https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/725 disappeared for me, when I tested this PR. I guess it's just me not merging with the master branch.

beanieboi:
@mkalininait I would like to address the changes within this PR. I can also help out. I will check with max, to not step on his toes. If fear if we merge this now and want to address it later that we pile up technical debt and maybe don't address it at all.
If it's really urgent, then we can merge it now, I don't want to block it.

Thanks for the PR and review <3

klappradla:
Oh, thanx for the info @mkalininait 🙏 

**Ad a) *(the button)***
I know it stays like that - I thought it would be a *"feature*" actually 😂 
So as the rendering takes hours, I want to prevent people pressing the button a million times, that's why I disable it after the first click. It then stays like this, since it's not an ajax request (the content type is `text/csv`).
**But**: pressing it again after that does not make any sense - the data would be exactly the same. Therefore I thought it would actually be useful to have it disabled all the time...

**Ad 2)**
Picks / Likes 1 and 2 are the number of `picks` and `likes` the application got from the mentors for `Project 1`and `Project 2´.
For the `average_points` being wrong and the last column overlapping I honestly don't know right from the top of my head now. Maybe it even has to do with #725 not being merged into this branch yet *(I did not have the time to do this yesterday)*.

And @beanieboi, feel free to go ahead with implementing the fixes if you're already on it. I won't have time until around 4pm CEST. I could then ping you again on Slack whether the things are already covered or not.

beanieboi:
@mkalininait I fixed the CSV output, it looks now the same as the HTML table. could you please check it again.

thank you!

app/views/rating/applications/index.csv.slim
```diff
@@ -10,12 +10,15 @@
...
-      'Likes1',
-      'Likes2',
+      'Signoff Project1',
+      'Signoff Project2',
+      'Fav Project1',
+      'Fav Project1',
+      'Picks',
+      'Likes',
       'AveragePoints',
-      'Ratings Misc'],
```
klappradla: 😂  oh hell, I already thought it must be something like this. I originally wrote a `%w()` style array but did not know how to use this in slim 🙈

klappradla:
Did not (yet) check it in the browser with current data, but looking at @beanieboi's changes, this should have been it @mkalininait. Good to merge I'd say.

mkalininait:
@klappradla Thanks for the clarification about the button. It makes sense.

@beanieboi I've tested the exported data. The spreadsheet looks good now. Thank you!

I agree that this PR is good to merge.

PR #725, opened 2017-03-31
todo refinments

beanieboi:
based on feedback in slack
- fixes a bug where we don't show the rated date
- changes redirect after the rating


mkalininait:
Tested locally, works! Thank you @beanieboi! 🥇

PR #727, opened 2017-04-02
Refactor ApplicationData

beanieboi:
This is meant as proof of concept. I started extracting application_data into it's own class.
Right now we just pass the hash around, which has a couple of drawbacks. We treat it like an AR Model but we are leaking the everywhere.
This introduce an API for ApplicationData which is easier to extend and maintain. It also brings us halfway to extracting application data into it's own AR model. 

it does not implement the setters yet. if we want to change application data, we still have to use the hash methods. Any input on that is appreciated. 

What do you think about the direction we are heading here?


app/models/application.rb
```diff
@@ -91,6 +30,11 @@ def self.data_label(key)
     ApplicationDraft.human_attribute_name(key)
   end
 
+  def data
+    # require 'pry'; binding.pry
```
klappradla: Debugging leftover.

app/models/application.rb
```diff
@@ -91,6 +30,11 @@ def self.data_label(key)
...
+    ApplicationData.new(self.application_data)
```
klappradla: This is personal preference, but I'd skip the `self` here. Being an AR model, there's a accessor for the attribute and it does not seem to be a pattern to use `self` in this app *(at least from what I see in this PR)* 😉

app/models/application_data.rb
```diff
@@ -0,0 +1,265 @@
+class ApplicationData
+  def initialize(data)
+    @data = data || {}
```
klappradla: Again personal preference, but I'd make this a default and keeping the assignments as simple as possible:
```ruby
def initialize(data = {})
  @data = data
end
```

app/models/application_data.rb
```diff
@@ -0,0 +1,265 @@
...
+  end
+
+  def student0_application_gender_identification
+    @data["student0_application_gender_identification"]
```
klappradla: Also a personal habit of mine with no impact on the functionality or visibility: I'd like to define a private `attr_reader` for things that get assigned within the constructor and then spare all the `@`s afterwards.

But as I said, just a personal habit of mine, no need to adapt this here.

spec/exporters/applications_spec.rb
```diff
@@ -22,7 +22,7 @@
     application
   end
 
-  let!(:new_application)           { create :application, :in_current_season }
+  let!(:new_application)  { create :application, :in_current_season }
```
klappradla: +1

klappradla:
Any objections for merging this? I think it's a nice improvement _(even reducing technical dept I'd say)_ with no impact on functionality, so I'd go for it 🚢

beanieboi:
yay, thank you

PR #728, opened 2017-04-02
Upgrade Dependencies - the second part

beanieboi:
I'm taking 1 dependency after another :)
This removes the version lock on Feedjira. The main problem is that feedjira switched from `curb` to `faraday`as HTTP framework and therefor does not support `file://` URL's anymore.
My workaround is to server the local files over HTTP via a small webrick server. This allows us to upgrade Feedjira and still maintain the same test coverage. 

I also included a regular `bundle upgrade` which upgraded the AWS gem, but this is safe since we switched to the `2.x` version.

Let me know what you think about that approach.

Thanks
Ben


spec/lib/feed_spec.rb
```diff
@@ -1,17 +1,29 @@
...
+require 'webrick'
 
 describe Feed do
   let(:out)    { StringIO.new }
   let(:logger) { Logger.new(out) }
 
+  before :all do
+    # serve the local directory for faraday
+    root = File.expand_path("spec/stubs/feeds")
+    @server = WEBrick::HTTPServer.new :Port => 8000, :DocumentRoot => root
```
klappradla: Nifty af 🤘 would not have come up with this 🙇

carpodaster: That is indeed *very* cool!

PR #729, opened 2017-04-04
Sentry config

mkalininait:
Can't test it locally, we may want to try it on Staging first. Or on Production - and see what happens :)

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -28,5 +28,13 @@ module RgsocTeams
     config.action_mailer.default_url_options = { host: 'teams.railsgirlssummerofcode.org' }
 
     config.active_job.queue_adapter = :sucker_punch
+
+    # Sentry config
+    Raven.configure do |config|
+      config.dsn = 'https://744f7c7520bf4217ac2d5cc60e5ef202:eee47d2a388b4d56b61b11536e8f9726@sentry.io/153582'
+    end
+
+    config.action_dispatch.show_exceptions = false
+
   end
 end
--- a/config/application.rb
+++ b/config/application.rb
@@ -28,13 +28,5 @@ module RgsocTeams
     config.action_mailer.default_url_options = { host: 'teams.railsgirlssummerofcode.org' }
 
     config.active_job.queue_adapter = :sucker_punch
-
-    # Sentry config
-    Raven.configure do |config|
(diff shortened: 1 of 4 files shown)


config/application.rb
```diff
@@ -28,5 +28,13 @@ class Application < Rails::Application
     config.action_mailer.default_url_options = { host: 'teams.railsgirlssummerofcode.org' }
 
     config.active_job.queue_adapter = :sucker_punch
+
+    # Sentry config
+    Raven.configure do |config|
+      config.dsn = 'https://744f7c7520bf4217ac2d5cc60e5ef202:eee47d2a388b4d56b61b11536e8f9726@sentry.io/153582'
```
carpodaster: This looks like we should set it via the `ENV`. And we need a new pair of username/password since it's now public.

carpodaster: Ah, apparently `sentry-raven` will automatically pickup a `SENTRY_DSN` environment variable so we don't actually have to configure it manually. Let's hope that it doesn't complain when it can't find any DSN – if it doesn't, we can just keep on hacking in development without bothering about configuring an exception tracker.

config/application.rb
```diff
@@ -28,5 +28,13 @@ class Application < Rails::Application
...
+    end
+
+    config.action_dispatch.show_exceptions = false
```
carpodaster: Do we really need this? It's not mentioned in https://docs.sentry.io/clients/ruby/#configuration

carpodaster: Oh wait, it does mention it in https://docs.sentry.io/clients/ruby/integrations/rails/, but it also states that it should be the default for production. I'd rather try it without this setting first.

config/application.rb
carpodaster: We can set: 

```ruby
Raven.configure do |config|
  config.environments = %w[production]
end
```

But let's do so in an initializer (`config/initializers/sentry.rb`).

config/application.rb
mkalininait: Yeah, I was not sure about this, too. I'll remove it for the moment, and we'll see how it works.

config/application.rb
mkalininait: Good points @carpodaster! I regenerated username/password in Sentry, added the SENTRY_DSN variable to Heroku and removed this line from `config/application.rb`.  I also set `config.environments` in `config/initializers/sentry.rb`.

mkalininait:
@carpodaster All changes are done 👍

carpodaster:
Superb! Thx 🌻

PR #731, opened 2017-04-04
Fix display of mentor-favs within ratings

klappradla:
Updated the application table and application-rating view to show a
heart for each mentor-fav (updated the csv export to contain the
respective "true" and "false" values as well).

closes rails-girls-summer-of-code/selection_process#57


beanieboi:
looks good to me, can't wait to move the ApplicationData into the database and have real data types 👯‍♂️

klappradla:
Did you find the time to take a look at this already @mkalininait ? If it's fine, I'd actually straight merge it together with #732 _(which does basically the same just as a "non-fix")_ and run the rake task to clean up the DB (cc @carpodaster ).

app/views/rating/applications/_application.html.slim
```diff
@@ -18,8 +18,8 @@ tr
   td.text-left = application.flags.map { |f| content_tag(:span, f, class: 'label label-default') }.join('<br>').html_safe
   td.text-center = icon('check') if application.application_data['signed_off_at_project1']
   td.text-center = icon('check') if application.application_data['signed_off_at_project2']
-  td.text-center = icon('check') if application.application_data['mentor_fav_project1']
-  td.text-center = icon('check') if application.application_data['mentor_fav_project2']
+  td.text-center = icon('heart') if application.application_data['mentor_fav_project1'] == 'true'
```
carpodaster: What about `[…].to_s == 'true'`? That way we also support real booleans.

klappradla: Sound defensive and *"future-proof"* love this 💚 
I have it in several places, will update it.

Does it need to be within this PR? It does not directly impact rating, so I could also do it a day later...?

carpodaster: We can do it later, of course!

klappradla: Ok. Will take care of merging and adding the respective task tonight. Nice, thanx 🙏

mkalininait:
Tested locally. Looks good to me, thank you @klappradla!

app/views/rating/applications/_application.html.slim
klappradla: Ok, already addressed it when merging `master` into #732 - it was only a few places where it made sense.

PR #732, opened 2017-04-04
Cleanup/align mentor fav

klappradla:
Hey @carpodaster,

we had some misunderstandings when preparing the rating views about how `mentor_fav` and `sign_off` work when being revoked (rails-girls-summer-of-code/selection_process#57).
In my personal opinion, it makes sense to align their behavior and just fully *reset* them to `nil` (as in just removing the key from the hash as if nothing ever happened) or do you see a value in having a `"false"` string there instead?

I also added one of these one-off rake tasks to correct the production data.


beanieboi:
I'm up for just deleting the key from the hash.

lib/tasks/single_run.rake
```diff
@@ -52,4 +52,17 @@ namespace :single_run do
...
   end
+
+  desc '2017-04-05: Set non-mentor favs to nil'
+  task set_non_mentor_favs_to_nil: :environment do
+    Application.rateable.each do |application|
+      fav1 = application.application_data['mentor_fav_project1']
+      fav2 = application.application_data['mentor_fav_project2']
+      if [fav1, fav2].include?('false')
+        application.application_data.delete('mentor_fav_project1') if fav1 == 'false'
+        application.application_data.delete('mentor_fav_project2') if fav2 == 'false'
```
bitboxer: Small info: my import used the methods `revoke_mentor_fav!` and `revoke_sign_off!`. Both are changed in the pr. Maybe it is also good to change the signed off here, too.

klappradla: `revoke_sign_off!` was ever since resetting both attributes (user ID and timestamp) to `nil` from what I remember, so it should be fine, shouldn't it?

bitboxer: Yes, that is true. Hope there is no false somewhere in the database for this. Maybe worth a check 👍

klappradla: Yep, will for sure do a check :v:
Will do tonight though, when I'm back on my home machine.

klappradla: Ok, had a look. Only `nil` and stringified IDs for the sign_off values. Should be fine 👌

klappradla:
Any objections on merging this? I just merged `master` back in.

This PR does not introduce dramatic changes to the way how mentor favs and sign_offs are handled, it just aligns their behavior which in my opinion makes it easier to understand for people who were not involved in the implementation process.

alicetragedy:
sounds good to me @klappradla 👍 
makes sense to align the behaviour.

klappradla:
Ok, going to merge it and run the rake task then.

PR #733, opened 2017-04-05
Render application fields with line breaks when rating

plexus:
Run all the bigger free text fields through simple_format(). This turns newlines into <br>, and also runs the text through sanitize().


plexus:
Ok, I think it's good now :)

2017-04-05-095759_168x113_scrot

carpodaster:
Awesome, thanks!

PR #735, opened 2017-04-07
Update puma config

klappradla:
I looked a bit into the `R14 - Memory Quota Exceeded` problem on Heroku *(production app)*. It looks like I configured too many Puma workers. See e.g. [this section](https://devcenter.heroku.com/articles/ruby-memory-use#too-many-workers) in Heroku's docs around the topic.

Using two workers is *theoretically* possible with a Rails app running on one dyno - but it does not work for every app.

TLDR: I think reducing the workers to just one should fix the problem of exceeding the memory quota basically 10min after app start.

### DIFF
--- a/config/puma.rb
+++ b/config/puma.rb
@@ -1,4 +1,4 @@
-workers Integer(ENV['WEB_CONCURRENCY'] || 2)
+workers Integer(ENV['WEB_CONCURRENCY'] || 1)
 threads_count = Integer(ENV['RAILS_MAX_THREADS'] || 5)
 threads threads_count, threads_count
 


klappradla:
Hm, I'm actually not really a Heroku user, no idea. I did
```sh
heroku config --app rgsoc-teams-production | grep CONCURRENCY
```
and it did not return anything...

klappradla:
**Edit:** seems one can enable it to be set (by opting into certain [defaults](https://devcenter.heroku.com/changelog-items/618) ) but it's not set automatically.

klappradla:
Ok, let's give this a shot then 🐒

PR #737, opened 2017-04-09
Fix for back-button behavior in Chrome

klappradla:
This prevents Chrome from responding with the wrong format when using
the back button. Background: Chrome's cache does not distinguish between
the formats otherwise.

Video fuck yeah:
header


klappradla:
The `activities` page is actually the only page where I ran into that problem. If someone knows other places (cc @alicetragedy ), please tell me.

mkalininait:
Wow, this is awesome! <3

PR #739, opened 2017-04-11
show median points for applications

beanieboi:
as requested in https://github.com/rails-girls-summer-of-code/selection_process/issues/53 this shows the median points nest to the mean (average) points

### DIFF
--- a/app/models/concerns/rateable.rb
+++ b/app/models/concerns/rateable.rb
@@ -5,7 +5,19 @@ module Rateable
   end
 
   def average_points
-    total_points.round(2)
+    return 0 if rating_points.empty?
+    mean(rating_points)
+  end
+
+  def median_points
+    return 0 if rating_points.empty?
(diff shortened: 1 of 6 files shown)


app/models/concerns/rateable.rb
```diff
@@ -5,7 +5,19 @@ module Rateable
   end
 
   def average_points
-    total_points.round(2)
+    return 0 if rating_points.empty?
+    mean(rating_points)
```
klappradla: Nifty!

beanieboi:
Yolo! Let's ship it!

mkalininait:
Hey @ben, I’ve tested it, thanks! My two comments: 1) the average points are not rounded anymore (while the median points are), could you please fix it? 2) Could you please add "Median points" to the csv export too? 🙏 

joxi_screenshot_1491941582874

PR #741, opened 2017-04-12
Fair reviews: an attempt to adjust rating by reviewer strictness

carpodaster:
Targets rails-girls-summer-of-code/selection_process#60.

Big shoutout and credit to @neumanrq 💚 

Oh and yes, no tests 😿 Not proud of it. But since it's so dependend on the actual data, we skipped it for now.

### DIFF
--- a/app/controllers/rating/applications_controller.rb
+++ b/app/controllers/rating/applications_controller.rb
@@ -14,6 +14,7 @@ class Rating::ApplicationsController < Rating::BaseController
 
   def index
     @table = applications_table
+    @strictness = Rating::Strictness.in_current_season
     respond_to do |format|
       format.html
       format.csv do
--- a/app/models/rating/strictness.rb
+++ b/app/models/rating/strictness.rb
@@ -5,7 +5,7 @@ class Rating::Strictness
     @ratings ||= Rating
                   .joins('JOIN applications ON ratings.rateable_id = applications.id')
                   .where(rateable_type: 'Application')
-                  .where(season_id: Season.current.id)
+                  .where('applications.season_id' => Season.current.id)
   end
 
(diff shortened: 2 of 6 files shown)


mkalininait:
I've tested it quickly. Do I get it right that we added one column with the corrected average to the table?
I'd suggest to round the numbers to two decimal places. Other than that, it looks interesting and promising. I'd probably merge it. However, I still need time to look at the numbers more thoroughly.

Thank you Carsten and Robin!

beanieboi:
I think we should add some very high level tests. I would create 2 reviewers, 2 application and some arbitrary ratings. then we run the `Rating::Strictness` and record the outcome and do assertions on that. the algorithm looks deterministic, so we should always get back the same results. this way we run the code at least and it allows for some basic refactoring. it also does not complicates the tests too much.

what do you think?

carpodaster:
Thank you for your feedback, @mkalininait and @beanieboi. 

@mkalininait yeah, rounding it would make sense. I did not add any rounding because the average and mean values were also not rounded. Would it be okay to round all floats in the rating table after after we've merged this PR?

@beanieboi Doing a couple of high-level tests is a very good suggestion. We've started already and will continue later.

carpodaster:
Thank you for your feedback, @klappradla! 

I've just pushed our basic coverage (and the rounding to the 2nd decimal). I'm gonna merge now since this is non-destructive and just adds another column to the ratings table.

@mkalininait, @neumanrq and I had a quick call and looking at how the adjusted ratings fare with actual production data and Maria's expertise in interpreting its effects. We had mixed feelings. We discussed taking the variance into account; Robin and I have scetched out an implementation but we'll leave that for another PR.

PR #742, opened 2017-04-19
submit application comments in ratings via ajax

michaelem:
I was annoyed during the rating process that every time I submitted a comment before rating an application all my entered data would be lost, so this should fix that.

Maybe it would be wise to have some integration testing for stuff like this, then again it is a bit hard to implement those, any opinions?

### DIFF
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -8,14 +8,20 @@ class CommentsController < ApplicationController
   PATH_PARENTS = []
 
   def create
-    comment = Comment.new(comment_params)
+    @comment = Comment.new(comment_params)
 
-    if (comment.text.present? && comment.save)
-      anchor = ActionView::RecordIdentifier.dom_id(comment)
+    if (@comment.text.present? && @comment.save)
+      anchor = ActionView::RecordIdentifier.dom_id(@comment)
     else
       flash[:alert] = "Oh no! We can't save your comment. Please try again?"
     end
-    redirect_to commentable_path(comment.commentable, anchor || nil)
+
+    @new_comment = self.class::PATH_PARENTS + [Comment.new(commentable: @comment.commentable)]
+
+    respond_to do |format|
+      format.html { redirect_to commentable_path(@comment.commentable, anchor || nil) }
+      format.js { }
+    end
   end
 
   private
--- a/app/views/comments/_comments.html.slim
+++ b/app/views/comments/_comments.html.slim
@@ -1,0 +1,13 @@
+  #comments
+    h2 Comments
+
+    - commentable.comments.each do |comment|
+      hr(id="#{dom_id comment}")
+      h5 = "#{ user_for_comment(comment) }, #{ l(comment.created_at, format: :short) }".html_safe
+      p = render_markdown(comment.text).html_safe
+
+    = simple_form_for new_comment, remote: true do |f|
+      = f.input :commentable_id, as: :hidden
+      = f.input :commentable_type, as: :hidden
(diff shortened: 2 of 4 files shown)


michaelem:
👋 @klappradla

app/controllers/comments_controller.rb
```diff
@@ -8,14 +8,20 @@ class CommentsController < ApplicationController
...
+      anchor = ActionView::RecordIdentifier.dom_id(@comment)
     else
       flash[:alert] = "Oh no! We can't save your comment. Please try again?"
     end
-    redirect_to commentable_path(comment.commentable, anchor || nil)
+
+    @new_comment = self.class::PATH_PARENTS + [Comment.new(commentable: @comment.commentable)]
+
+    respond_to do |format|
+      format.html { redirect_to commentable_path(@comment.commentable, anchor || nil) }
```
klappradla: Is there still a non `remote true` form somewhere? If not, one could get rid about the `format.html` as well as everything related to `anchor` and `flash` since this won't be used for the ajax request anyways.

app/views/comments/_comments.html.slim
```diff
@@ -0,0 +1,13 @@
+  #comments
+    h2 Comments
+
+    - commentable.comments.each do |comment|
+      hr(id="#{dom_id comment}")
+      h5 = "#{ user_for_comment(comment) }, #{ l(comment.created_at, format: :short) }".html_safe
```
klappradla: Hm, maybe out of the scope of this PR, but I kinda don't like that we here first query for the comments and then **again** for the comment's user *(even though we in the end only want his or her name...)*. Do you think it would make sense to add something like a `comments_with_author` method to the concern used to make model commentable that would do a `includes(:user)`? One could then eventually also simplify the helper method used here a bit...

app/views/comments/create.js.erb
```diff
@@ -0,0 +1 @@
+$("#comments").html("<%= j render 'comments', commentable: @comment.commentable, new_comment: @new_comment %>")
```
klappradla: Personal preference, but I prefer to use data attributes, e.g. `data-behavior="comment-container" for things related to scripting in order to be able to fully leave IDs and classes to styling.

app/views/comments/_comments.html.slim
```diff
@@ -0,0 +1,13 @@
...
+      p = render_markdown(comment.text).html_safe
+
+    = simple_form_for new_comment, remote: true do |f|
+      = f.input :commentable_id, as: :hidden
+      = f.input :commentable_type, as: :hidden
+      = f.input :text
+      = f.submit 'Add comment'
```
klappradla: I'd add a `disable_with` here _(probable need to switch the input to a button then)_, users will probably hit the button multple times if there's no direct feedback and I currently don't see an option to delete a comment again.

app/controllers/comments_controller.rb
michaelem: Yes, the CommentsController is used all over the app, fore example projects are comment-able:
https://github.com/rails-girls-summer-of-code/rgsoc-teams/blob/master/app/views/projects/show.html.slim#L81

The following models seem support comments:
- Application
- Activity
- Project
- Team

app/views/comments/_comments.html.slim
michaelem: Generally in favor, but I would do this in a separate PR. (Also affects all sorts of parts of the app.

klappradla: Yep, totally agree 👍

PR #744, opened 2017-04-23
Simplify database config

klappradla:
Looking at the configuration for the production and staging apps on Heroku, I found the current database config unnecessarily complicated. We do not provide a `database.yml` with a `production` section, so Heroku just creates a default one after each deploy. However, recent versions of Rails allow to specify the DB connection via a url *(which Heroku provides us)* - see e.g. the comment in a fresh Rails 5 app's `database.yml` file:
```yaml
# On Heroku and other platform providers, you may have a full connection URL
# available as an environment variable. For example:
#
#   DATABASE_URL="postgres://myuser:mypass@localhost/somedatabase"
#
# You can use this database configuration with:
#
#   production:
#     url: <%= ENV['DATABASE_URL'] %>
#
```

When deploying the app with a `database.yml` file including an `production` section, Heroku skips creating a own one and uses this configuration. This will then allow us to have more specific settings, such as the pool size just specified globally for all environments, which is more transparent in my opinion. Rails 5 now also even suggests this in their default `database.yml`:

```yaml
# The specified database role being used to connect to postgres.
# To create additional roles in postgres see `$ createuser --help`.
# When left blank, postgres will use the default role. This is
# the same name as the operating system user that initialized the database.
```

However, feel free to close this PR if you think the old setup makes more sense, just wanted to propose this alternative :v:

PS: I adapted the Readme and `travis.yml` accordingly. There was a special *warning* section about Postgres on OSX - I could not verify this being an issue any more, so I removed it.


klappradla:
You mean because people want to e.g. use their own specific user with a password or use a completely different database name @carpodaster? I honestly did not think about this use case 🙈.
I expected this not to be a common thing to do... Or at least I'd expect the people who do this to also be used to taking the necessary actions to not accidentally commit this file..

Do you see the problem for newcomers to Rails / the project or for anyone?

A tiny disadvanta I see with not having the file under source control would be not being able to set things like the pool size. Heroku also has a [blog post](https://devcenter.heroku.com/changelog-items/709) on the topic of `database.yml` files where they mention this.

klappradla:
Hm, does anyone have an idea how to solve this in a better way? I'm not too familiar with Heroku...
Otherwise, I think it's fine to close this again :v:

klappradla:
Ok, could not come up with a better solution, let's close this 👋

carpodaster:
Sorry, @klappradla, didn't mean to sit this one out. I couldn't come up with another solution, either

klappradla:
No worries @carpodaster - not that important after all 😉

PR #745, opened 2017-05-14
Use expected values to improve strictness computation

neumanrq:
When @carpodaster  and me have proposed the [first version of reviewer strictness computation](https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/741) I made a mistake because of oversimplification. As @mkalininait correctly pointed out, our current formula gets slightly unfair when the amount of ratings is not equal for all reviewers.

Because I was a bit confused at all I started formalizing our problem en détail [here](https://github.com/neumanrq/fair_reviewers/blob/master/theory/index.pdf), where the new formula is taken from (It is the formula for `S^*`, currently Definiton 9 on page 5).

The short version is: After merging this, the strictness will be computed by determining the expectation value for each reviewer and comparing it to the "overall" computation value, instead of a more simple comparison of "amount of points assigned by reviewer A" with "amount points assigned by an average reviewer". 

I'd love to take @carpodaster into account: Maybe we could pair on this during next week and check it against real data. Of course, everyone is invited to ask sceptical questions on this 👀  😄

### DIFF
--- a/app/models/rating/strictness.rb
+++ b/app/models/rating/strictness.rb
@@ -76,9 +76,9 @@ class Rating::Strictness
     # given reviewer_ids only
     rating_subgroup = ratings.where(user_id: ids)
 
-    # Count the frequencies: How often did a specific
-    # reviewer given a specic value? The relative frequencies
-    # are used as probabilities
+    # Count the frequencies: How often has a specific
+    # reviewer assigned a specic point value? The
+    # distribution hash stores the absolute frequencies
     distribution = rating_subgroup.each_with_object({}) do |rating, hash|
       hash[rating.points.to_s] ||= 0
       hash[rating.points.to_s]  += 1
--- a/app/models/rating/strictness.rb
+++ b/app/models/rating/strictness.rb
@@ -72,8 +72,6 @@ class Rating::Strictness
   # @return [Float] for expected value for application raring of
   # given set of reviewers
   def expected_value_for(ids)
-    distribution = {}
-
     # We are working with the ratings of
     # given reviewer_ids only
     rating_subgroup = ratings.where(user_id: ids)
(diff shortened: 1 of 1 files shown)


mkalininait:
Related to this issue: https://github.com/rails-girls-summer-of-code/selection_process/issues/60

mkalininait:
Sorry, I closed this PR by accident. Reopened it already.

mkalininait:
Hi @neumanrq First of all, sorry for not showing up for some time. We had an exhausting selection and then were exhaling for a couple of weeks :D 

I've tested this PR on the production data and prepared a spreadsheet where you can compare the OLD corrected average and NEW corrected average. I'll gladly share it with you if you give me your gmail account. Briefly speaking, we have 186 applications; 9 applications have absolute difference between the old and new corrected averages more than 0.5; two versions of top-20 applications (sorted by new and old corrected averages) differ only in 1 application, which is not a lot, but can be significant for those specific people from flesh and blood.

I'm still digging through the theoretical part of the research and will leave my feedback here once I finish.

Thanks again for your work! We really appreciate it!

klappradla:
Thanx a million for your efforts on this @neumanrq 🙏 
And sorry for leaving this lingering around for so long 🙈 

I tbh cannot really say whether or not this fixes the problem we have (plus I don't even recall what the problem was) - but I trust your brain, so I'd just 1, 2, 3 merge this is no one is against it (cc @mkalininait )

neumanrq:
Thanks @klappradla for taking care :) 

Oh yes, I'm completely aware that this is hard to jump in. I'm still convinced these changes will fix the problem described above. I should take the time and do a little write-up to clarify and document the motivation behind these parts for you, the other maintainers, and also future-me 😬

klappradla:
Anyways ❤️💛💚💙💜

PR #746, opened 2017-05-14
UX: Show that country is a required field

F3PiX:
The country is a required field, and now the user can tell before they get a warning `on submit` .


carpodaster:
🎊 Thanks!

PR #747, opened 2017-05-15
Add validation for the chronological order of conference dates

### LINKED ISSUE #503: Conferences - small improvements for adding new

F3PiX:
Closes #503 (bullet point 2)
Replacement for #552 

- Add custom validator method + specs + factory
- Minor cosmetics

### DIFF
--- a/app/controllers/orga/conferences_controller.rb
+++ b/app/controllers/orga/conferences_controller.rb
@@ -12,8 +12,11 @@ class Orga::ConferencesController < Orga::BaseController
   end
 
   def update
-    conference.update_attributes(conference_params)
-    redirect_to orga_conference_path(conference)
+    if conference.update_attributes(conference_params)
+      redirect_to orga_conference_path(conference)
+    else
+      render action: :edit
+    end
   end
 
   def destroy
@@ -35,9 +38,9 @@ class Orga::ConferencesController < Orga::BaseController
 
   def conference_params
     params[:conference] ? params.require(:conference).permit(
-      :name, :url, :location, :twitter, :tickets, :flights, :accomodation,
-      :'starts_on(1i)', :'starts_on(2i)', :'starts_on(3i)', :round,
-      :'ends_on(1i)', :'ends_on(2i)', :'ends_on(3i)', :lightningtalkslots,
+      :name, :url, :location, :twitter,
+      :tickets, :flights, :accomodation,
+      :starts_on, :ends_on, :round, :lightningtalkslots,
       attendances_attributes: [:id, :github_handle, :_destroy]
     ) : {}
   end
--- a/app/controllers/orga/conferences_controller.rb
+++ b/app/controllers/orga/conferences_controller.rb
@@ -12,7 +12,7 @@ class Orga::ConferencesController < Orga::BaseController
   end
 
   def update
-    if conference.update_attributes(conference_params)
+    if conference.update(conference_params)
       redirect_to orga_conference_path(conference)
     else
       render action: :edit
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -16,12 +16,7 @@ module ApplicationHelper
       link_to 'Apply now', apply_path
     end
   end
-
-  def conference_tweet_link(conference)
-    tweet = "I really want to go to #{conference.name} this year! #{conference.twitter} @Railsgirlssoc"
-    "https://twitter.com/intent/tweet?text=#{URI.escape(tweet)}"
-  end
-
+  
   def avatar_url(user, size: 200)
     image = if user_avatar = user.avatar_url.presence
               "#{user_avatar}&s=#{size}"
@@ -67,10 +62,11 @@ module ApplicationHelper
     date && l(date.to_date, format: format) || '-'
   end
 
+  ## Conferences
   def format_conference_date(starts_on, ends_on)
-    return unless [starts_on, ends_on].all?(&:present?)
-    starts_on = starts_on.strftime('%d %b %y')
-    ends_on = ends_on.strftime('%d %b %y')
+    return unless starts_on && ends_on
+    starts_on = starts_on.strftime("%d %b %y")
+    ends_on = ends_on.strftime("%d %b %y")
     starts_on == ends_on ? starts_on : [starts_on, ends_on].join(' - ')
   end
 
@@ -86,10 +82,28 @@ module ApplicationHelper
     result
   end
 
+  def conference_tweet_link(conference)
+    tweet = "I really want to go to #{conference.name} this year! #{conference.twitter} @Railsgirlssoc"
+    "https://twitter.com/intent/tweet?text=#{URI.escape(tweet)}"
+  end
+
   def format_conference_twitter(twitter)
     twitter.to_s.starts_with?('@') ? link_to(twitter, "http://twitter.com/#{twitter.gsub('@', '')}") : twitter
   end
 
+  # @param conferences [Array<Conference>] a list of Conference records
+  # @return [Array<String>] a list of HTML anchor tags to conferences
+  def links_to_conferences(conferences)
+    conferences.map do |conference|
+      text = conference.name
+      extra = [conference.location, format_conference_date(conference.starts_on, conference.ends_on)].reject(&:blank?).join(' – ')
+      text += " (#{extra})" unless extra.blank?
+      link_to(text, conference)
+    end
+  end
+  
+  ## end Conferences
+  
   def if_present?(user, *attrs)
     yield if attrs.any? { |attr| user.send(attr).present? }
   end
@@ -123,21 +137,7 @@ module ApplicationHelper
       link_to(user.name.present? ? user.name : user.github_handle, user, class: attendance.confirmed? ? 'confirmed' : '')
     end
   end
-
-  # @param conferences [Array<Conference>] a list of Conference records
-  # @param verbose [Boolean] appends location and date info if true (defaults to false)
-  # @return [Array<String>] a list of HTML anchor tags to conferences
-  def links_to_conferences(conferences, verbose: false)
-    conferences.map do |conference|
-      text = conference.name
-      if verbose
-        extra = [conference.location, format_conference_date(conference.starts_on, conference.ends_on)].reject(&:blank?).join(' – ')
-        text += " (#{extra})" unless extra.blank?
-      end
-      link_to(text, conference)
-    end
-  end
-
+  
   def links_to_user_teams(user)
     user.teams.map do |team|
       link_to(team.name || team.project, team, class: "team #{team.sponsored? ? 'sponsored' : ''}")
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -62,9 +62,8 @@ module ApplicationHelper
     date && l(date.to_date, format: format) || '-'
   end
 
-  ## Conferences
+  # Conferences
   def format_conference_date(starts_on, ends_on)
-    return unless starts_on && ends_on
     starts_on = starts_on.strftime("%d %b %y")
     ends_on = ends_on.strftime("%d %b %y")
     starts_on == ends_on ? starts_on : [starts_on, ends_on].join(' - ')
@@ -96,13 +95,13 @@ module ApplicationHelper
   def links_to_conferences(conferences)
     conferences.map do |conference|
       text = conference.name
-      extra = [conference.location, format_conference_date(conference.starts_on, conference.ends_on)].reject(&:blank?).join(' – ')
-      text += " (#{extra})" unless extra.blank?
+      details = [conference.location, format_conference_date(conference.starts_on, conference.ends_on)].reject(&:blank?).join(' – ')
+      text += " (#{details})"
       link_to(text, conference)
     end
   end
   
-  ## end Conferences
+  # end Conferences
   
   def if_present?(user, *attrs)
     yield if attrs.any? { |attr| user.send(attr).present? }
--- a/app/views/orga/teams/show.html.slim
+++ b/app/views/orga/teams/show.html.slim
@@ -79,7 +79,7 @@ nav.btn-group
   ul.conferences
     - @team.members.each do |user|
       - if user.conferences.in_current_season.any?
-        li = "#{link_to user.name, user}: #{links_to_conferences(user.conferences.in_current_season, true).join(',')}".html_safe
+        li = "#{link_to user.name, user}: #{links_to_conferences(user.conferences.in_current_season).join(',')}".html_safe
 
 - if @team.activities.any?
   h3 Activities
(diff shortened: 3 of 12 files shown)


spec/controllers/orga/conferences_controller_spec.rb
```diff
@@ -3,6 +3,8 @@
 RSpec.describe Orga::ConferencesController do
   render_views
 
+  it_behaves_like 'redirects for non-admins'
+
```
F3PiX: Okay to add this?

spec/helpers/application_helper_spec.rb
```diff
@@ -115,14 +115,14 @@
...
       end
 
       context 'with location' do
         let(:conference) { build_stubbed :conference, location: 'Melée Island, Carribean' }
 
-        it 'will add the location in brackets' do
+        xit 'will add the location in brackets' do
           link_text = "#{conference.name} (Melée Island, Carribean)"
           expect(subject).to eql [link_to(link_text, conference)]
         end
```
F3PiX: These tests are related to formatting the location and the dates in several views, handling the layout of conferences with and without info with dates.  
Now with the `presence of dates validations`, is that code still relevant? (For example, existing records.) Or should it be removed.

spec/factories/conferences.rb
```diff
@@ -1,7 +1,9 @@
 FactoryGirl.define do
   factory :conference do
     name { [FFaker::CheesyLingo.title, 'Conf'].join ' ' }
-    tickets 2
```
F3PiX: I removed the tickets, because they are not a requirement for validation. Plus, the only test that refers to the ticket attributes, overrides it.
(Copied from old PR)

app/controllers/orga/conferences_controller.rb
```diff
@@ -35,9 +38,9 @@ def conference
...
     params[:conference] ? params.require(:conference).permit(
-      :name, :url, :location, :twitter, :tickets, :flights, :accomodation,
-      :'starts_on(1i)', :'starts_on(2i)', :'starts_on(3i)', :round,
-      :'ends_on(1i)', :'ends_on(2i)', :'ends_on(3i)', :lightningtalkslots,
+      :name, :url, :location, :twitter,
+      :tickets, :flights, :accomodation,
+      :starts_on, :ends_on, :round, :lightningtalkslots,
       attendances_attributes: [:id, :github_handle, :_destroy]
     ) : {}
   end
```
F3PiX: (Regrouped and reformatted for better readibility.)

F3PiX:
Re: my refactor fest. I'd really like to refactor the Conference's `(:starts_on, :ends_on)` to a DateRange object. I think the correct order is do the refactoring first, and the new validation afterwards. 
So, two questions: 1) Do you agree with the refactoring itself?
2) If so, should I do that first?

klappradla:
On refactoring vs. feature first: do as you wish ✌️ 
It totally depends on the state of the code base and how easy it is for you to work with its current state. The "classic" approach is feature first, then refactoring. But of course, if that part of the codebase has not been touched in a while or whatever other reason one could find, it may of course also be way more efficient to do the refactoring first.

Sometimes you even know upfront, that you cannot use any of the existing code - it makes of course sense to just throw it all away and implement your new way of doing things from scratch then.

**TL;DR** do however it suits your workflow to best archive what you want to.

spec/helpers/application_helper_spec.rb
klappradla: If the view helper is no longer used, yes, remove both, the helper and the tests. If it is still used, I don't see a point in removing the tests for it.

spec/models/conference_spec.rb
```diff
@@ -7,6 +7,22 @@
   it { is_expected.to have_many(:attendees) }
   it { is_expected.to validate_presence_of(:name) }
   it { is_expected.to validate_presence_of(:round) }
+  it { is_expected.to validate_presence_of(:starts_on) }
+  it { is_expected.to validate_presence_of(:ends_on) }
+
+  describe 'it checks the chronological order of dates' do
+    subject { FactoryGirl.build(:conference) }
```
klappradla: It's a common best practise and in my opinion also increases readability to leave a blank line between `subject` and the first example.

I'm also in general not sure about the use of `subject` here? I'm not working with Rails on a day to day basis, but from what I remember, the `errors` field is only appended to a model and updated accordingly if `model.valid?` is called... So I'm not sure if your expectation here actually does what it should.

spec/models/conference_spec.rb
```diff
@@ -7,6 +7,22 @@
...
+    it 'raises an error with an incorrect order of dates' do
+      subject.starts_on = '2017-07-15'
+      subject.ends_on = '2017-07-07'
+      expect(subject).to have(1).error_on(:ends_on)
```
klappradla: Personal preference, but I'm not a huge fan of the `rspec-collection-matchers` gem (I would love to remove it from the Teams App 😉  ) since it also continues to ship matchers from RSpec < v3, such as `errors_on`... In my opinion there's probably a reason why this is no longer in RSpec.

How about just using plain RSpec here?
```ruby
subject.valid?
expect(subject.errors[:ends_on]).not_to be_empty
```
Shouldn't it do the same?

spec/models/conference_spec.rb
carpodaster: IIRC,  the has errors matcher internally calls `#valid?`

klappradla: Ah, would make sense @carpodaster, thanx 🙏

spec/helpers/application_helper_spec.rb
F3PiX: Which reminds me: are there currently records of conferences without the dates set?
I don't have access to the database. 

(As far as I can see, the input form that prepopulate the dates was present [from the beginning of the Conferences model  ](https://github.com/rails-girls-summer-of-code/rgsoc-teams/commit/9bf58943bb7bc49c52ea2569e25962b5bf81c974#diff-da8e2af5531074e5eb6389d13add128a) )

app/controllers/orga/conferences_controller.rb
```diff
@@ -35,9 +38,9 @@ def conference
 
   def conference_params
     params[:conference] ? params.require(:conference).permit(
```
F3PiX: Also ancient code. I am not sure when I should and shouldn't touch existing code that is not explicitly part of the issue I am working on. 
I really _like_ the concept of leaving code cleaner than you found it, but isn't that the opposite of doing one thing at the time? (Refactor hat or Feature hat, IIRC) . 
How do I know when to clean up, or when to leave it alone?

spec/helpers/application_helper_spec.rb
klappradla: I had a quick look into the current production data - could not find any conferences with `starts_on` or `ends_on` set to `nil`. So I guess that's not a problem...

app/controllers/orga/conferences_controller.rb
klappradla: You usually know it when you remove it, run the tests and everything still works 😉  But yeah, risky, let's just focus on what you actually want to change.

I also had a quick look at the whole controller code. There's this line:
```ruby
@conference ||= params[:id] ? Conference.find(params[:id]) : Conference.new(conference_params)
```
This for instance causes looking for a non-existing conference to just create a new one instead of raising an error (Record not Found) - so that's why the hack with the empty hash exists 💡

spec/models/conference_spec.rb
F3PiX: :thinking: It looks is doesn't after all. 
Without `subject.valid?` `subject.errors[:ends_on]` is always empty. 
I can either add `subject.valid?`, or do 
`expect(subject).not_to be_valid`. Does it matter?

spec/helpers/application_helper_spec.rb
F3PiX: Thanks for checking 👍

spec/factories/conferences.rb
```diff
@@ -1,7 +1,9 @@
...
+    starts_on { Time.utc(Date.today.year, 7, 7) }
```
F3PiX: (Noticed the CheesyLingo lib in the FFaker gem ^ ? That was created by Team Cheesy during Soc15, and dedicated to SoC.) #thingsrailsgirlsdo

F3PiX:
I leave the daterange refactor for after this one is done.
Re: Removing/simplifying the application helper method: see the [commit message](https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/747/commits/de7333c4e79d84f8cc1e7191c37f34bccbb48e95)

F3PiX:
new application helper

spec/models/conference_spec.rb
klappradla:
```ruby
expect(subject).not_to be_valid
```
would generally be fine, if you can make sure that the error is not triggered from a completely different validation. So if doing it like this maybe also quickly check if the subject otherwise is valid

```ruby
subject { thing } 

it { is_expected.to be_valid }

# more specs
```

db/seeds.rb
```diff
@@ -6,7 +6,7 @@
 FactoryGirl.create(:team, :last_season, kind: "sponsored")
 FactoryGirl.create(:team, :last_season, kind: "voluntary")
 
-# Users with different roles
+# # Users with different roles
```
klappradla: Please remove the extra '#' *(or tell me what's the purpose of this 😉 )*

db/seeds.rb
```diff
@@ -16,7 +16,7 @@
 FactoryGirl.create(:mentor)
 FactoryGirl.create(:supervisor)
 
-# To explore use cases where user has no role yet
+# # To explore use cases where user has no role yet
```
klappradla: Again: Please remove the extra '#' *(or tell me what's the purpose of this 😉 )*

app/helpers/application_helper.rb
```diff
@@ -67,10 +62,11 @@ def format_date(date, format = :long)
     date && l(date.to_date, format: format) || '-'
   end
 
+  ## Conferences
   def format_conference_date(starts_on, ends_on)
-    return unless [starts_on, ends_on].all?(&:present?)
-    starts_on = starts_on.strftime('%d %b %y')
-    ends_on = ends_on.strftime('%d %b %y')
+    return unless starts_on && ends_on
```
klappradla: Is this a scenario that could ever happen? As far as I understand it, there are presence validations on these fields already...?

app/views/orga/teams/show.html.slim
```diff
@@ -79,7 +79,7 @@ nav.btn-group
   ul.conferences
     - @team.members.each do |user|
       - if user.conferences.in_current_season.any?
-        li = "#{link_to user.name, user}: #{links_to_conferences(user.conferences.in_current_season, true).join(',')}".html_safe
+        li = "#{link_to user.name, user}: #{links_to_conferences(user.conferences.in_current_season).join(',')}".html_safe
```
klappradla: Since the helper method is already doing quite a lot, I would suggest to also move `join` and `html_safe` there. It looks kind of weird to me, having a rather long method called `links_to_conferences` not returning the actual links in a *"view-ready*" format.

What do you think?

It would also make this list item a bit cleaner:
```ruby
li = "#{link_to user.name, user}: #{link_to_conferences(user.conferencs.in_current_season)}
```

app/views/teams/show.html.slim
```diff
@@ -91,7 +91,7 @@ nav.page
   ul.conferences
     - @team.members.each do |user|
       - if user.conferences.in_current_season.any?
-        li = "#{link_to user.name, user}: #{links_to_conferences(user.conferences.in_current_season, verbose: true).join(', ')}".html_safe
+        li = "#{link_to user.name, user}: #{links_to_conferences(user.conferences.in_current_season).join(', ')}".html_safe
```
klappradla: Same as for the other conferences link: I would suggest to move things into the helper if it's already there.

db/seeds.rb
F3PiX: It's a feature ;-) Like an H2, under H1 Users. But maybe MaudFlavoredMarkdown doesn't work that good if I have to explain it. Will delete it.

app/views/orga/teams/show.html.slim
F3PiX: I totally agreed, but I didn't find a way to solve it. Spent too much time on it, I'll make a note and will try it another time. Okay?

klappradla: Sure ✌️ 

The easiest way would have been to just append the `join` etc. to the already existing method:
```ruby
def links_to_conferences(conferences)
  conferences.map do |conference|
    text = conference.name
    extra = [conference.location, format_conference_date(conference.starts_on, conference.ends_on)].reject(&:blank?).join(' – ')
    text += " (#{extra})" unless extra.blank?
    link_to(text, conference)
  end.join(',').html_safe
end
```
But yeah, there's probably a prettier way.

app/helpers/application_helper.rb
```diff
@@ -96,13 +95,13 @@ def format_conference_twitter(twitter)
   def links_to_conferences(conferences)
     conferences.map do |conference|
       text = conference.name
-      extra = [conference.location, format_conference_date(conference.starts_on, conference.ends_on)].reject(&:blank?).join(' – ')
-      text += " (#{extra})" unless extra.blank?
+      details = [conference.location, format_conference_date(conference.starts_on, conference.ends_on)].reject(&:blank?).join(' – ')
+      text += " (#{details})"
       link_to(text, conference)
```
klappradla: Well, if the condition is gone now, I'd suggest to make this all a bit smaller:
```ruby
def links_to_conferences(conferences)
  details = [conference.location, format_conference_date(conference.starts_on, conference.ends_on)].reject(&:blank?).join(' – ')
  link_to "#{conference.name} (#{details})", conference
end
```

But well, optional 😉

app/views/orga/teams/show.html.slim
F3PiX: Yeah, I expected that to work too. And it didn't. That's how my quest started ;-)

app/helpers/application_helper.rb
F3PiX: I hate it that I forgot that..... :-)
I'll merge on green. Thanks again a lot!!!

PR #748, opened 2017-05-22
Refactor Fest: Extract Parameter Object from Conference dates 

F3PiX:
Introduce a DateRange object, which makes it possible to pass a date range instead of a start date and an end date.
For now: only implemented for Conference dates.
Bonus: makes formatting the dates in the view easier and more consistent.
So now we have 'compact ranges' , like `15 - 16 Jul 2017` instead of repeating the months and years. Yay!

In the table:
screen shot 2017-05-22 at 15 37 20

In team profile:
screen shot 2017-05-22 at 15 38 07

PS Part of an ongoing series of polishing the Conferences


app/models/date_range.rb
```diff
@@ -21,10 +21,10 @@ def compact
     
     same_years = @start_date.year == @end_date.year
     same_months = (@start_date.month == @end_date.month) && same_years
-    elements = elements.gsub(' %Y', '') if same_years
-    elements = elements.gsub(' %b', '') if same_months
-    
+    elements = elements.chomp(' %Y') if same_years
+    elements = elements.chomp(' %b') if same_months
```
F3PiX: Theoretically, we can omit the `same year` check in `same months`, because `chomp` will only remove the month-substring if it is at the end of the string. 
I left it in though, because it is easier to understand. Agree?

app/models/date_range.rb
```diff
@@ -0,0 +1,30 @@
...
+  def initialize(start_date, end_date)
+    @start_date = start_date
+    @end_date = end_date
+  end
+  
+  # Accepts dates
+  # Returns a string with one date or a formatted range:
+  # 12 - 13 Jul 2017 or 30 Jul - 1 Aug 2017 or
+  # 31 Dec 2017 - 5 Jan 2018 or 1 Dec 2017 - 12 Dec 2018
+  def compact
```
F3PiX: The method will raise an error when the input is not a `Date`. So I didn't an `raise Error` at entrance. Should I?

spec/models/date_range_spec.rb
```diff
@@ -0,0 +1,59 @@
+require 'spec_helper'
+
+RSpec.describe DateRange do
+
+  describe '#dates' do
+      let(:conference) { FactoryGirl.build_stubbed(:conference)}
```
klappradla: Indentation is slightly messed up here.

spec/models/date_range_spec.rb
```diff
@@ -0,0 +1,59 @@
+require 'spec_helper'
+
+RSpec.describe DateRange do
+
+  describe '#dates' do
```
klappradla: Since this test targets a method on `Conference`, rather then on `DateRange`, I think this would be a better fit for the `Conference` model specs.

spec/models/date_range_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+    it 'creates a date range for calling Models' do
+      expect(conference.dates).to be_an_instance_of DateRange
+    end
+  end
+  
+  describe '#compact' do
+    let (:conference)  { FactoryGirl.build_stubbed(:conference) }
+    subject { described_class.new(start_date: Date.parse('2015-12-31'), end_date: Date.parse('2015-12-31')) }
+
+    it 'responds to other models' do
```
klappradla: This test appears slightly *"off topic"*, as it does not really test the `DateRange` class. In my opinion all it does, is to see if `#compact` is a public method on a `DateRange`, which is already obvious through all the other specs in this block... So I'd vote for skipping this all remove the conference here completely.

spec/models/date_range_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+      let(:conference) { FactoryGirl.build_stubbed(:conference)}
+    
+    it 'creates a date range for calling Models' do
+      expect(conference.dates).to be_an_instance_of DateRange
+    end
+  end
+  
+  describe '#compact' do
+    let (:conference)  { FactoryGirl.build_stubbed(:conference) }
+    subject { described_class.new(start_date: Date.parse('2015-12-31'), end_date: Date.parse('2015-12-31')) }
```
klappradla: Just a matter of style, but I think the tests are more meaningful if the `subject` reflects the actual "subject" of all expectations. So in order to dry this up a bit and make it more expressive I'd suggest to switch it to be actually `compact`:

```ruby
let(:start_date) { Date.parse('2015-12-31') }
let(:end_date)   { Date.parse('2015-12-31') }

subject { described_class.new(start_date: start_date, end_date: end_date).compact }

# etc.
```

And then, for different contexts, all you have to do is change the values for `starts_at` and `ends_at`:
```ruby
context 'when multiple dates in same month' do
  let(:start_date) { Date.parse('2015-12-30') }
  let(:end_date    { Date.parse('2015-12-31') }

  it 'returns compact days' do
    expect(subject).to eq ('30 - 31 Dec 2015')
  end
end
```

What do you think?

spec/models/date_range_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+    it 'responds to other models' do
+      expect(conference.dates.compact).to be
+    end
+
+    context 'depending on date range' do
+      it 'returns one day range' do
+        expect(subject.compact).to eq subject.end_date.to_date.strftime('%-d %b %Y')
+      end
+      
+      context ' when multiple dates in same month' do
```
klappradla: Extra whitespace.

app/models/conference.rb
```diff
@@ -12,12 +12,16 @@ class Conference < ActiveRecord::Base
   scope :ordered, ->(sort = {}) { order([sort[:order] || 'starts_on, name', sort[:direction] || 'asc'].join(' ')) }
   scope :in_current_season, -> { where(season: Season.current) }
 
+  def dates
```
klappradla: So, looking at the latter usage, this method only serves as a view helper and is never used solo, but *always* concatinated with `#compact`. So why offering direct access to `dates` at all, if it's never actually used? This can eventually make things more complex later on... imagine someone forgets to call `compact` - nothing useful will be rendered.

I would suggest to just rename this method to `date_range` and directly return the compact version of it.

This can of course be done in multiple ways:
a) call the public method in the model
```ruby
DateRange.new(start_date: starts_on, end_date: ends_on).compact
```

b) directly return the compact version, e.g.
```ruby
class DateRange
  # ...
  def self.call(start_date:, end_date:)
    self.new(start_date: start_date, end_date: end_date).compact
  end
end

# in the model:
DateRange.call(start_date: starts_on, end_date: ends_on)
```

app/models/date_range.rb
```diff
@@ -0,0 +1,30 @@
+class DateRange
+  attr_reader :start_date, :end_date
+  
+  # Step 1: only in use with Conference date range.
+  # Validated in Conference: presence and chronological order
+  #  because the input is done there.
+  
+  def initialize(start_date: start_date, end_date: end_date)
```
klappradla: Copy paste error *(or otherwise I don't understand the syntax)* 😉 

If you want to use required KWARGS, this should look like this:
```ruby
def initialize(start_date:, end_date:)
```

spec/models/conference_spec.rb
```diff
@@ -41,6 +41,16 @@
       end
     end
   end
+  
+  describe '#dates' do
+    subject { FactoryGirl.build_stubbed(:conference) }
+    
+    it 'has a date range' do
```
klappradla: Not sure if the expectations here really fit the test. It does not test whether or not the return value is a `DateRange` object. Whether or not the internal values of the `DateRange` object match those of the conference is not really relevant I would say - that's internal logic of the `DateRange` class.

You could test the call to the constructor:
```ruby
it "has a date range" do
  expect(DateRange)
    .to receive(:new)
    .with(start_date: subject.starts_on, ends_on: subject.ends_on)
  expect(subject.dates).to be_a(DateRange)
end
```

app/models/date_range.rb
```diff
@@ -0,0 +1,30 @@
+class DateRange
+  attr_reader :start_date, :end_date
```
klappradla: What's the reason these attributes are public? The `DateRange` class is a simple Service object or "View Helper", I don't see the necessity for these to ever be called outside the scope of this class. So I'd suggest to move them to private:

```ruby

private

attr_reader :start_date, :end_date
```

app/models/conference.rb
```diff
@@ -12,12 +12,16 @@ class Conference < ActiveRecord::Base
...
+    DateRange.new(start_date: starts_on, end_date: ends_on)
+  end
+  
   def chronological_dates
     unless starts_on <= ends_on
       errors.add(:ends_on, 'must be a later date than start date')
     end
   end
-
+
```
klappradla: Trailing whitespaces.

F3PiX:
Thank you so much!! I am on it.

spec/models/conference_spec.rb
F3PiX: I tried everything I could think of, but it wouldn't pass. (I used the correct keyword `end_date`.)
It boils down to that after the `.with`-part, `subject.dates` is nil. 

Later, I changed the `dates` method to `date_range` to include compact, and then of course the method does not return a DateRange anymore, but a formatted string. 

Help? I think it makes sense to check if the outgoing message is being sent - if I understand the Sandi Metz rules correctly. But how?

app/models/conference.rb
F3PiX: I opt for a). 
One other question: shouldn't I store/memoize the `date_range`?

klappradla: Ah yes, would make good sense here I think 👍 
The method is used several times while rending some views and it's unlikely that start or end will change (and require recomputation) in between.

Good idea!

spec/models/conference_spec.rb
klappradla: Hm, ok.

So the last expectation on its own should work. So if you remove the first one,
```ruby
it "bla bla" do
  expect(subject.dates).to be_a(DateRange)
end
```
this should pass.

For the rest, maybe you have to add `and_call_original``
```ruby
it "bla bla" do
  expect(DateRange)
    .to receive(:new)
    .with(start_date: subject.starts_on, end_date: subject.ends_on)
    .and_call_original
  # etc.
end
```

carpodaster: (The Pipeline Operator Influence   `|>` is strong in this one :wink: :heart_decoration:)

F3PiX: Are you going Elixir on me? 😨

spec/models/conference_spec.rb
```diff
@@ -41,12 +41,21 @@
...
     end
   end
+  
+  describe '#dates' do
+    subject { FactoryGirl.build_stubbed(:conference) }
+    
+    xit 'has a date range' do
+      expect(subject.date_range).to be_an_instance_of DateRange
+    end
+  end
```
F3PiX: The new `date_range` method returns a formatted string, so now it is not a DateRange anymore. 
How about: 
`expect(subject.date_range).to be`

F3PiX:
@klappradla Back to you! 🙇

app/models/date_range.rb
```diff
@@ -0,0 +1,33 @@
...
+    @start_date = start_date
+    @end_date = end_date
+  end
+  
+  # Accepts dates
+  # Returns a string with one or more dates, with
+  # the following formats:
+  # 12 - 13 Jul 2017 or 30 Jul - 1 Aug 2017 or
+  # 31 Dec 2017 - 5 Jan 2018 or 1 Dec 2017 - 12 Dec 2018
+  def display
```
klappradla: Minor thing: after you defined an `attr_reader`, you'd access `start_date` and `end_date` without the `@` *(that's the purpose of the reader)*

spec/models/conference_spec.rb
```diff
@@ -41,12 +41,21 @@
       end
     end
   end
+  
+  describe '#dates' do
+    subject { FactoryGirl.build_stubbed(:conference) }
+    
+    it 'has a date range' do
+      expect(subject.date_range).to be_present
```
klappradla: Hm, I'm not a huge fan of `be_present` or `be` alone, I think it does not tell anything while still being as slow as most other expectations... How about at least do a
```ruby
be_a(String)
```
?

spec/models/date_range_spec.rb
```diff
@@ -0,0 +1,49 @@
+require 'spec_helper'
+
+RSpec.describe DateRange do
+  
+  describe '#compact' do
+    let(:start_date) { Date.parse('2015-12-31') }
+    let(:end_date)   { Date.parse('2015-12-31') }
+    subject { described_class.new(start_date: start_date, end_date: end_date).display }
```
klappradla: Leaving a blank line between the `let` statements and the subject would be 💚 - makes things way better to read.

spec/models/date_range_spec.rb
```diff
@@ -0,0 +1,49 @@
+require 'spec_helper'
+
+RSpec.describe DateRange do
+  
+  describe '#compact' do
```
klappradla: This should now be `#display` instead.

app/models/date_range.rb
```diff
@@ -0,0 +1,33 @@
...
+    same_years = @start_date.year == @end_date.year
+    same_months = @start_date.month == @end_date.month
+    elements = elements.chomp(' %Y') if same_years
+    elements = elements.chomp(' %b') if same_months
+    first_day = @start_date.strftime(elements)
+    
+    first_day << ' - ' << last_day
+  end
+
+  private
```
klappradla: It's best practise and a plus for readability to leave a blank line after the `private` macro.

app/models/date_range.rb
carpodaster: May I suggest naming this method `#to_s`? That way a `DateRange` instance can be used in `"string #{interpolations}"` and it follows a common interface amongst Ruby object. Reusing method names for similar functionality is very important in terms of polymorphism (Duck Typing in Ruby's case).

spec/models/date_range_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+    let(:start_date) { Date.parse('2015-12-31') }
+    let(:end_date)   { Date.parse('2015-12-31') }
+    subject { described_class.new(start_date: start_date, end_date: end_date).display }
+    
+    context 'depending on date range' do
```
carpodaster: These are very good test cases! 💚

app/models/conference.rb
```diff
@@ -12,14 +12,18 @@ class Conference < ActiveRecord::Base
   scope :ordered, ->(sort = {}) { order([sort[:order] || 'starts_on, name', sort[:direction] || 'asc'].join(' ')) }
   scope :in_current_season, -> { where(season: Season.current) }
 
+  def date_range
+    @date_range ||= DateRange.new(start_date: starts_on, end_date: ends_on).display
```
carpodaster: If the `DateRange#display` is to be renamed to `#to_s` (see other comment), I would return the `DateRange` instance here, not the string. It's just a minor thing, but to me having the method match a (snake_cased) Ruby class name and also semantically _meaning_ the same thing but returning a string is a bit confusing.

app/models/date_range.rb
```diff
@@ -0,0 +1,33 @@
...
+    last_day = @end_date.strftime(elements)
+    return last_day if @start_date == @end_date
+    
+    same_years = @start_date.year == @end_date.year
+    same_months = @start_date.month == @end_date.month
+    elements = elements.chomp(' %Y') if same_years
+    elements = elements.chomp(' %b') if same_months
+    first_day = @start_date.strftime(elements)
+    
+    first_day << ' - ' << last_day
```
carpodaster: This will fail with frozen string literals (Ruby 2.3 feature) as it modifies a String in place. Can you change the ``<<`` to a `+`?

app/models/date_range.rb
```diff
@@ -0,0 +1,33 @@
...
+  # Returns a string with one or more dates, with
+  # the following formats:
+  # 12 - 13 Jul 2017 or 30 Jul - 1 Aug 2017 or
+  # 31 Dec 2017 - 5 Jan 2018 or 1 Dec 2017 - 12 Dec 2018
+  def display
+    elements = "%-d %b %Y" # 1 Jul 2017
+    last_day = @end_date.strftime(elements)
+    return last_day if @start_date == @end_date
+    
+    same_years = @start_date.year == @end_date.year
```
carpodaster: How do you feel about removing the local variables and using the comparison expression directly in the `if` modifiers below?

app/models/date_range.rb
```diff
@@ -0,0 +1,33 @@
...
+  
+  # Accepts dates
+  # Returns a string with one or more dates, with
+  # the following formats:
+  # 12 - 13 Jul 2017 or 30 Jul - 1 Aug 2017 or
+  # 31 Dec 2017 - 5 Jan 2018 or 1 Dec 2017 - 12 Dec 2018
+  def display
+    elements = "%-d %b %Y" # 1 Jul 2017
+    last_day = @end_date.strftime(elements)
+    return last_day if @start_date == @end_date
```
carpodaster: (Personally, I try to avoid `return` statements in the middle of a method, but I don't have an idea how to improve it here)

app/models/date_range.rb
klappradla: +1 thought about this as well @F3PiX 

This would allow to skip the `display` part in the model (where it semantically doesn't belong) - makes a lot of sense here I'd say.

F3PiX: I didn't know that, thanks.

app/models/date_range.rb
F3PiX: Let me try something else

app/models/conference.rb
F3PiX: I removed them, not sure why it still shows up here.

F3PiX:
Thanks @carpodaster and @klappradla!!! Look at how much better it is now!!
I really love the `to_s`, and all the changes were super easy to implement because of all the other changes in the beginning of the PR. Even in the specs!! 
I couldn't find a better way for the [postponed return statement  ](https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/748/files#diff-56e7f37be264728c800fc00646920fb7R20). At least not without either repeating the 'elements' or move the 'same day'-check to the end of the method in a ternary.

app/views/conferences/_details.html.slim
```diff
@@ -3,7 +3,7 @@
     dt Location
     dd = conference.location
     dt Date
-    dd = format_conference_date(conference.starts_on, conference.ends_on)
+    dd = conference.date_range.to_s
```
carpodaster: I think you can remove the `to_s`, I'm pretty sure it is called automatically.

klappradla:
Hm, you could move `elements` and `last_day` out into their own private methods @F3PiX  - then the early return statement would become an actual guard clause. You'd then have to adapt the later mutating of your local `elements` variable...

F3PiX:
I tried that. I think I can't use an instance method, because that will return not the string "%-d ..." but a "4 Jun 2017" string. 
I did came up with this, but I didn't like it that much. WDYT?
```
ELEMENTS = "%-d %b %Y"
(...)
def to_s
    return last_day if start_date == end_date
 
    elements = ELEMENTS
    elements = elements.chomp(' %Y') if start_date.year == end_date.year
    elements = elements.chomp(' %b') if start_date.month == end_date.month
    first_day = start_date.strftime(elements)
    first_day + ' - ' + last_day
  end

def last_day
    end_date.strftime(ELEMENTS)
end 
```

klappradla:
Hm, looks fine to me (aside from the 4 spaces intend... 😉 ). This is what I originally meant:
```ruby
def to_s
  return last_day if start_date == end_date
  elements = base_format
  elements = elements.chomp(' %Y') if start_date.year == end_date.year
  elements = elements.chomp(' %b') if start_date.month == end_date.month
  start_date.strftime(elements) + ' - ' + last_day
end

private

def base_format
  @base_format ||= "%-d %b %Y"
end

def last_day
  end_date.strftime(base_format)
end
```

But using a constant is of course also a valid solution.

F3PiX:
Thanks Max! 🎈 🌷

PR #749, opened 2017-05-23
Update to Rails 5.1

klappradla:
Sorry, too tired for a long description.

This basically updates all dependencies and switches to Rails v.5.1.1. There were some special difficulties with deprecations warnings on AR::Dirty, see my commit message in 2a4c1f7


klappradla:
Well @carpodaster these are the minimum changes to get us running with **5.1.1**. While upgrading I saw some room for maybe cleaning up config files, etc. a bit. But I wanted to do this in a separate PR because I a) did not have the time to dive into every detail and b) don't think it is a core part of *"update dependencies"* 😉 

There was for instance this one thing I wanted to ask you and @michaelem for you opinion:

While upgrading I realized that Rails changed its default serializer for cookies from `:marshal` to `:json` *(probably already in Rails 5)*. I didn't do the change to the `config/initializers/cookie_serializer.rb` file because it will break all existing cookies out there 🙈 

But I wanted to ask you, if you'd be ok with this change maybe in a separate PR? It would basically just be this line:

```ruby
Rails.application.config.action_dispatch.cookies_serializer = :json
```

The change is not particularly necessary, I just think JSON is a way better and more futureproof format then "Marshal".

What do you think?

carpodaster:
> While upgrading I realized that Rails changed its default serializer for cookies from :marshal to :json (probably already in Rails 5). I didn't do the change to the config/initializers/cookie_serializer.rb file because it will break all existing cookies out there 🙈

[I had some bad experience with it](http://devnull.absolventa.de/2014/08/30/migrating-back-to-marshal-ed-cookie-serialization-in-rails-4-1/)

klappradla:
Oh, 🙈  makes sense!

I thought that in our case it would only be auth-related things from Devise. But yes, Github's OAuth gem is probably also accessing the session 🤔

michaelem:
Since inheriting directly from ActiveRecord::Migration now raises an error (https://github.com/rails/rails/commit/249f71a22ab21c03915da5606a063d321f04d4d3):
How do you feel about updating old migrations?
Note that the lowest version supported is [4.2], so it would be strange as well. In any case, as of now the migrations are no longer able to run.

klappradla:
Oh, good point @michaelem !
I tbh never ever ran all migrations, I just do a `db:setup`...

What about just removing all the old migrations @carpodaster @michaelem ? We could of course also update all of them to inherit from e.g. `ActiveRecord::Migration[5.0]`, but I personally don't see a huge benefit in keeping them around.

Issue #752, opened 2017-05-28
Migrations not running


F3PiX:
Problem:
Since the upgrade to rails 5.1 in #749  , `rails db:migrate` results in an error, as reported by @michaelem https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/749#issuecomment-304333342 
```
StandardError: Directly inheriting from ActiveRecord::Migration is not supported. Please specify the Rails release the migration was written for:
  class RatingsAddLike < ActiveRecord::Migration[4.2]
```

I had to run `rails db:schema:load` instead, so I could still use the app.

Solution: 
How should this be solved?  

@michaelem @carpodaster @klappradla

klappradla:
Hey @F3PiX,

thanx for reporting 🙏 

For "resetting" the database, or setting up the project from scratch, I usually do the following:
```sh
# drop the old dbs if there are any
dropdb <dev-db-name>
dropdb <test-db-name>

# set them up fresh including seed data:
bundle exec rails db:setup
```

But yeah, we'll need to come up with a solution for the old migrations as well...

F3PiX:
👍 
I usually do `bundle exec rails db:reset`. Which is db:drop and db:setup in one. Is `rails db:drop`  a different drop than `dropdb` ? 

But still: any ideas for a solution for the old migrations?

klappradla:
`dropdb` is a Postgres command, but probably what `rails db:drop` does under the hood anyways.

For the old migrations, I could either update all of them to no longer directly inherit from `ActiveRecord::Migration` or just remove them. I would personally vote for the latter, since in my opinion the migration files **should not be** the source of truth for a Rails app's database schema. There is the `schema.rb` file under version control and that's the one I care about.

I would however love to have a second opinion on that, since that's simply how I interpret certain features of the framework.

F3PiX:
( PR to squash migrations from last year: https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/381 )

F3PiX:
🌟 🎉

PR #753, opened 2017-06-01
More 5.1 adjustments

klappradla:
This adjusts some of the minor things I did not consider a core part of the *pure update* in #749 

1. **Update all dependencies that changed since then**
  Since the 5.1 update, some gems had minor updates - since I get notifications from Gemnasium anyways, I just added it to this PR while I'm already on it.

1. **Adjust the config files**
  I adjusted the config files to better resemble the current state of the art in Rails 5.1. Some options, such as `load_defaults` of course don't work with our app. I also did not opt into auto-reloading via the `listener` gem, since I a) only heard bad things about machines running out of processes and b) it would add another dependency to the app 😉 
I think it's generally helpful to keep the config files clean and up to date if we consider further maintaining and updating the app. Aside that, I also feel it would be more helpful, especially to beginners, to find up 2 date comments and an up 2 date general structure there - since that's what people currently learning the framework are familiar with *(and it's what helps them most to find the respective places in the docs)*

1. **Remove migrations**
  As I explained in #752 , I don't see migrations as the *"source of truth"* for the database schema, that's why there's a `schema.rb` file in version control. And since I don't think we want to rollback migrations forever, it in my opinion should be fine to just remove them. If we want changes, we can still add new migrations...

**Note:** I also updated the Puma config. So before this gets deployed to either staging or production, I will set the `WEB_CONCURRENCY` env var there, in order to have the exact same setup as we are running now (1 worker).


klappradla:
Also: we "soft-deprecated" certain columns during the preparation for this year's selection process with the promise to then remove them. I would after this is merged go ahead and do this in order to get rid of these "legacy columns" before this year's team starts working on the app.

klappradla:
😂  @carpodaster 

But yes, I actually cannot remember ever deleting migrations - thought it would often make sense. Therefore nice that Rails forces us into these new waters :v:

😂  ✌️  @carpodaster 
I also cannot remember ever deleting migrations... though it actually makes sense! So thanx Rails broke the backwards compatibility, *"winkd

PR #755, opened 2017-06-05
Use Postgres 9.5 on Travis

klappradla:
As the `production` app runs Postgres 9.5 and there have been some changes regarding `hstore` *(and other JSON related things)* between 9.3 and 9.5, I thought it would be saver to also run our CI with the respective version.

However: it's not available on Travis' default container infrastructure, so I had to switch to the (beta) [trusty linux](https://docs.travis-ci.com/user/getting-started/#Selecting-infrastructure-(optional)).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: ruby
 cache: bundler
 sudo: false
+dist: trusty
 rvm:
 - 2.4.1
 bundler_args: "-j4 --without development production --quiet"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,6 +1,7 @@
 language: ruby
 cache: bundler
 sudo: false
+dist: trusty
```
carpodaster: I _trust_ (get it? :rofl:) Travis doesn't have Xenial?

klappradla: Apparently not @carpodaster 😞  - the default stack is on 12.04 🐧 ...

PR #756, opened 2017-06-05
Move users_info controller and view into orga namespace

F3PiX:
Some prep work before working in the User Info views.

Users Info is a view for orga's only, but lived outside the Orga namespace.
Now it is in the Orga namespace.
Added a link to the Orga dashboard.


carpodaster:
Oh wait, CI is failing. The failing routing spec has be moved – other than that: yeah!

F3PiX:
Sorry, I forgot to run the tests.... 😊

PR #758, opened 2017-06-10
Remove deprecated columns from DB

klappradla:
This removes columns deprecated during the last selection phase. They weren't used anymore, but we decided to keep them around "deprecated" during the selection phase to remove them later on.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,194 +10,187 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170330225217) do
+ActiveRecord::Schema.define(version: 20170610162757) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
   enable_extension "hstore"
 
-  create_table "activities", force: :cascade do |t|
-    t.integer  "team_id"
-    t.string   "kind",         limit: 255
-    t.string   "guid",         limit: 255
-    t.string   "author",       limit: 255
-    t.string   "title",        limit: 255
-    t.text     "content"
-    t.string   "source_url",   limit: 255
+  create_table "activities", id: :serial, force: :cascade do |t|
+    t.integer "team_id"
+    t.string "kind", limit: 255
+    t.string "guid", limit: 255
+    t.string "author", limit: 255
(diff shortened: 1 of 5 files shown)


db/schema.rb
```diff
@@ -10,194 +10,187 @@
...
 
-  create_table "activities", force: :cascade do |t|
-    t.integer  "team_id"
-    t.string   "kind",         limit: 255
-    t.string   "guid",         limit: 255
-    t.string   "author",       limit: 255
-    t.string   "title",        limit: 255
-    t.text     "content"
-    t.string   "source_url",   limit: 255
+  create_table "activities", id: :serial, force: :cascade do |t|
```
F3PiX: Where are the `serials` coming from? Is that standard behaviour now after migration?

klappradla: Yes, there have been changes to the `create_table` method, such as that the definition of primary keys is now part of this method. The Postgres adapter then uses [this function](https://github.com/rails/rails/blob/c6b4b4a52c7ec66d1923ef44dcbc71b582d7b165/activerecord/lib/active_record/connection_adapters/postgresql/schema_definitions.rb#L44) to define it *(it could for instance also be a UUID or a big-serial)*.

F3PiX: Ok, I knew about AR now using `bigint` for id's, but my own Rails 5.1 app doesn't add the type explicitly. 
So, does that mean that we now have to define a `:serial` explicitly when we create a migration? I just ask this because if so, it would be helpful if you can update the readme accordingly. 
Also: if people have an older version of pg running, will that be a problem?  The Readme says: '9.3' or higher.

klappradla: No, should not be necessary to explicitly cite the ID type, unless you need something different like`UUID` of course. From what I remember AR just adds the information based on the underlying adapter and its version *(in my case 9.6.3)*

So for your second question: technically speaking, it is no problem to run the app on Postgres versions < 9.5 as long as the people doing it do not always commit their changed `schema.rb` if they did not add any new migrations themselves. Therefore, in order to prevent a row of commits, e.g. removing the `limit` option on every string column, etc. I'm gonna update the Readme to require a Postgres version approxing the one run on production (9.5) for development 😉 
Good point, thanx!

PR #750, opened 2017-05-25
[BLOCKED] Polishing conferences table

F3PiX:
Still working on polishing the Conferences. 

- Reverse toggle, so now it shows Round 1 by default, and toggle Round 2
- Update button
- Move table title from namespace view to shared partial.

Goal: making working on the Conferences table easier and nicer.

PR #761, opened 2017-06-15
Prepare conferences table for upcoming changes 

F3PiX:
Instead of #750 
Now that the Conference raffle rounds are gone, and we are preparing for a different setup, this PR is preparing the conferences table. 

- Moved title to partial
- Removed toggle in view for round 1 vs 2
- Removed repeated table for second round
- Temp: added round (sortable) to table
- Restored sortable functionality (sortability used to be there, but got lost somewhere)
- Updated button style
- cleanup controller, use regular CRUD setup

### DIFF
--- a/app/controllers/conferences_controller.rb
+++ b/app/controllers/conferences_controller.rb
@@ -9,7 +9,6 @@ class ConferencesController < ApplicationController
   end
 
   private
-
     def conferences
       @conferences ||= Conference.ordered(sort_params).in_current_season
     end
@@ -22,7 +21,7 @@ class ConferencesController < ApplicationController
 
     def sort_params
       {
-        order: %w(name location starts_on).include?(params[:sort]) ? params[:sort] : nil,
+        order: %w(name location starts_on round).include?(params[:sort]) ? params[:sort] : nil,
         direction: %w(asc desc).include?(params[:direction]) ? params[:direction] : nil
       }
     end
--- a/app/controllers/conferences_controller.rb
+++ b/app/controllers/conferences_controller.rb
@@ -3,6 +3,14 @@ class ConferencesController < ApplicationController
 
   def new
   end
+ 
+  def index
+    @conferences = conferences
+  end
+ 
+  def show
+    @conference = Conference.find(params[:id])
+  end
 
   def redirect
     redirect_to orga_conferences_path
@@ -11,14 +19,8 @@ class ConferencesController < ApplicationController
   private
  
   def conferences
-    @conferences ||= Conference.ordered(sort_params).in_current_season
+    Conference.ordered(sort_params).in_current_season
   end
-  helper_method :conferences
-
-  def conference
-     @conference ||= Conference.find(params[:id])
-  end
-  helper_method :conference
 
   def sort_params
     {
--- a/app/controllers/conferences_controller.rb
+++ b/app/controllers/conferences_controller.rb
@@ -9,20 +9,21 @@ class ConferencesController < ApplicationController
   end
 
   private
-    def conferences
-      @conferences ||= Conference.ordered(sort_params).in_current_season
-    end
-    helper_method :conferences
+ 
+  def conferences
+    @conferences ||= Conference.ordered(sort_params).in_current_season
+  end
+  helper_method :conferences
 
-    def conference
-       @conference ||= Conference.find(params[:id])
-    end
-    helper_method :conference
+  def conference
+     @conference ||= Conference.find(params[:id])
+  end
+  helper_method :conference
 
-    def sort_params
-      {
-        order: %w(name location starts_on round).include?(params[:sort]) ? params[:sort] : nil,
-        direction: %w(asc desc).include?(params[:direction]) ? params[:direction] : nil
-      }
-    end
+  def sort_params
+    {
+      order: %w(name location starts_on round).include?(params[:sort]) ? params[:sort] : nil,
+      direction: %w(asc desc).include?(params[:direction]) ? params[:direction] : nil
+    }
+  end
 end
--- a/app/controllers/orga/conferences_controller.rb
+++ b/app/controllers/orga/conferences_controller.rb
@@ -47,7 +47,7 @@ class Orga::ConferencesController < Orga::BaseController
 
   def sort_params
     {
-      order: %w(name location starts_on).include?(params[:sort]) ? params[:sort] : nil,
+      order: %w(name location starts_on round).include?(params[:sort]) ? params[:sort] : nil,
       direction: %w(asc desc).include?(params[:direction]) ? params[:direction] : nil
     }
   end
--- a/app/controllers/orga/conferences_controller.rb
+++ b/app/controllers/orga/conferences_controller.rb
@@ -1,9 +1,14 @@
 class Orga::ConferencesController < Orga::BaseController
 
+  def new
+    @conference = Conference.new
+  end
+  
   def create
-    conference.season = current_season
+    @conference = Conference.new(conference_params)
+    @conference.season = current_season
     respond_to do |format|
-      if conference.save
+      if @conference.save
         format.html { redirect_to orga_conference_path(conference) }
       else
         format.html { render action: :new }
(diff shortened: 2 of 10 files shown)


klappradla:
Just to let you know @F3PiX : reviewing this is on my agenda, but I'm in the process of moving, so it will probably take me some days until I find the time...

F3PiX:
In a previous commit, [a long existing method came up](https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/747#discussion_r116864600), and it was not clear why it was there.
It looks like it causes trouble for my importer method in an upcoming PR, so I added it to the Spring Cleaning Fest.

app/controllers/orga/conferences_controller.rb
```diff
@@ -1,9 +1,14 @@
 class Orga::ConferencesController < Orga::BaseController
 
+  def new
+    @conference = Conference.new
+  end
+  
   def create
-    conference.season = current_season
+    @conference = Conference.new(conference_params)
```
klappradla: This may come down to personal preference (+ I don't know if this still works with Rails 5.1), but I usually try to have all params in place before initializing:
```ruby
@conference = Conference.new(conference_params.merge(season: current_season)
```

app/controllers/orga/conferences_controller.rb
```diff
@@ -1,9 +1,14 @@
 class Orga::ConferencesController < Orga::BaseController
 
+  def new
+    @conference = Conference.new
```
klappradla: The forms used for `new` and `edit` currently use the `conference` helper. Since it seems like there's a `@conference` previously initialized by the controller method anyways, I'd vote for getting rid of the `helper_method` macro and instead use the normal `@conference` variable in the views to spare this weird back and forth between view and controller methods.

app/controllers/conferences_controller.rb
```diff
@@ -9,21 +9,21 @@ def redirect
   end
 
   private
+ 
+  def conferences
+    @conferences ||= Conference.ordered(sort_params).in_current_season
+  end
+  helper_method :conferences
```
klappradla: Personal preference, but since this will probably not be used outside the `index` action, I'd really like to see the `@conferences` being explicitly defined there, rather then using the `helper_method` macro in the view. I generally don't like this pattern and this something like this is way more obvious, especially for newcomers to the app:

```ruby
def index
  @conferences = confereces
end

private

def conferences
  Conference.ordered etc
end

# in the view:
= render 'conferences/table', namespace: :orga, conferences: @conferences
```

app/views/conferences/_table.html.erb
```diff
@@ -1,61 +1,31 @@
...
   </thead>
 
   <tbody>
-  <% conferences.where(round: 1).each do |conference| %>
+    <% conferences.each do |conference| %>
       <tr>
         <td> <%= link_to conference.name, [namespace, conference] %></td>
         <td> <%= conference.location %></td>
         <td> <%= conference.date_range %></td>
+        <td> <%= conference.round if conference.round.present? %> </td>
```
klappradla: What's your intention here? If `round` is not present, so either `nil` or `""`, this will render an empty string. So the behavior would be exactly the same if you'd simply leave:
```ruby
<td><%= conference.round %>
```

app/views/conferences/_table.html.erb
```diff
@@ -1,61 +1,31 @@
...
-<br>
-<p><%= link_to "View as Calendar", calendar_path %>
-</p>
-
-<p><%= link_to "Show/Hide First Round Conferences", "#", onclick: "$('.first-conferences').toggle()" %></p>
-
-<div class="first-conferences" style='display:none;'>
-<h1 class=" header">Conferences (1st round)</h1>
-
+<h1 class="header">Conferences</h1>
```
klappradla: Why is this an `.erb` file if the rest of the app is using slim?

app/views/conferences/_table.html.erb
```diff
@@ -1,61 +1,31 @@
...
 <table class="table-striped table-bordered table-condensed conferences">
   <thead>
     <tr>
-      <th> <%= sortable "Conference" %> </th>
-      <th> <%= sortable "Location" %> </th>
-      <th> <%= sortable "Dates" %> </th>
+      <th> <%= sortable :name, "Conference" %> </th>
```
klappradla: Why is there a whitespace before and after the content? If this is done for layout reasons, then this should be done via css if necessary at all in my opinion.

app/views/conferences/_table.html.erb
F3PiX: The excuse is that I thought we were going to get rid of slim. The real reason is that I found slim very confusing, so I always convert to and from erb, and that is not time I feel is well spent. Can you live withe erb here?

klappradla: Sure - was just wondering if it's on purpose or not.

I'm generally against dropping slim as I think it gives an interesting tiny little "transfer-thinking" effort to the Teams App: HTML is limited and not a programming language. So it just takes a few looks at the docs *(and of course some trial and error)* to express the same things one has in mind for HTML in slim.

But yeah, can imagine it being tedious if you start with a full .erb implementation and try to "convert" it to .slim 😉

F3PiX:
@klappradla Back to you 🙇

app/controllers/orga/conferences_controller.rb
```diff
@@ -1,53 +1,58 @@
 class Orga::ConferencesController < Orga::BaseController
+  before_action :find_resource, only: [:show, :edit, :update, :destroy]
```
klappradla: Comes down to personal preference, but I'd prefer this method to be named `find_conference` as it's more tied to what's actually happening. With before filters being a tricky thing in general, I think being as explicit as possible can still be helpful here.

app/controllers/orga/conferences_controller.rb
```diff
@@ -1,53 +1,58 @@
...
+    @conference.destroy!
     redirect_to orga_conferences_path, notice: 'The conference has been deleted.'
   end
 
   private
 
-  def conferences
-    @conferences ||= Conference.ordered(sort_params).in_current_season
+  def find_resource
+    @conference ||= Conference.find(params[:id])
```
klappradla: Since we're now no longer using the helper method in the views, it should be fine to switch this to
```ruby
def find_resource
  Conference.find(params[:id])
end
```

app/controllers/orga/conferences_controller.rb
```diff
@@ -1,53 +1,58 @@
...
-    @conference ||= params[:id] ? Conference.find(params[:id]) : Conference.new(conference_params)
+  
+  def conferences
+    Conference.ordered(sort_params).in_current_season
   end
-  helper_method :conference
 
   def conference_params
-    params[:conference] ? params.require(:conference).permit(
+    params.require(:conference).permit(
```
klappradla: 👍

app/views/conferences/_table.html.erb
klappradla: Any news on the whitespace story here? I'd vote for removing them and if necessary fix any spacing issues with CSS *(with bootstrap)* ✌️

app/views/orga/conferences/_form.html.slim
```diff
@@ -1,9 +1,9 @@
-= simple_nested_form_for ([:orga, conference]) do |f|
-  - if conference.errors.any?
+= simple_nested_form_for ([:orga, @conference]) do |f|
```
klappradla: What's the purpose of this whitespace here? Shouldn't it just be
```ruby
simple_nested_form_for([:orga, @conference]) do |f|
  # do stuff
```
?

app/controllers/orga/conferences_controller.rb
F3PiX: ... but we still use it in the controller actions. Right?

F3PiX:
@klappradla 🎉

app/controllers/orga/conferences_controller.rb
klappradla: Yes we do, but not in that *"clever"* way it was used before with utilizing memorization to spare some lines in the controller code. Now, with doing it more or less explicit (I'm not the hugest fan of before filters...) we can spare the memorization part - the helper method will always just be called once.

PR #775, opened 2017-07-03
Conferences Importer

### LINKED ISSUE #762: [Conferences] #1: Import a longlist of conferences
Goal:
Orga user wants to show the students a 'long-list' of the conferences. The conferences are currently managed with a Google Doc. 
Later on, students will interact with the list, to choose their preferences, suggest other conferences etc. The scope of this issue is only uploading the file. 

- Add a file importer, so that an orga can upload a csv-file. The contents will be shown in the existing Conference table view and saved in the conferences db table. 
- Update the existing conference table structure to show the relevant fields.* 
- Write specs. 

* @mkalininait :   
- I am thinking: conference `name`, `link to` their websites, `dates`, `location`, for now. 
Maybe also: what the conference is about. Do you collect that kind of data in your Google Doc? For example , front/back end, language specific, devops. Otherwise we can add a simple 'notes' column. 
- In Google Docs, can you make a csv-export with just the columns we want to show? Or should we make it so that the not-listed columns are ignored. 
  
**UPD by @mkalininait:**
 * The entire idea: https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/751#issuecomment-308574743
 * Full list of conference user stories: https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/751#issuecomment-309138513

F3PiX:
Step 1 in New Conferences functionality
closes #762 : "Import a longlist of conferences"

- [x]  Would it be better to instantiate the `Conferences::Importer`?
Answer: I decided not to, because there is no state left after the import has finished. 
Furthermore, but I am not sure about this one:  handling the `@tempfile` that is created by `ActionDispatch::Http::UploadedFile` on import, seems to make things overly complicated with an instantiated importer.

### Purpose
Showing the students a long list of conferences. 
(Managing the conferences is done by the orga's in a Google Spreadsheet.)  

### What's new?
Orga has a Google Spreadsheet, exports to .csv, and uploads the file in the Teams App
The new `Conferences::Importer` processes the file
The importer uses a `gid` ('Google ID' LOL ) to reference the conference, and to decide on updating an existing conference or creating a new one. 
The imports file is a long list of conferences, so not all the details are known upfront. Some have unknown dates; that's why I removed the presence-validations on the dates. 
The 'old' way of adding and editing the conferences is removed. 

ToDo's:
- [x] More specs. Under construction. 
- [x] Cleaning up the views 
- [x] Before merging: Remove csv content from spec file


app/assets/stylesheets/global.sass
```diff
@@ -151,8 +151,11 @@ html, body
         height: 20px
 
     &.conferences
+      font-size: 0.9em
       td
```
F3PiX: @alicetragedy Okay? I like it when the table font is a bit smaller than the regular text.

alicetragedy: yes, works for me! 👍

config/environments/production.rb
```diff
@@ -75,6 +75,9 @@
 
   # Use default logging formatter so that PID and timestamp are not suppressed.
   config.log_formatter = ::Logger::Formatter.new
+  
+  # MEMO: check if imports_logger gives output if this is not set?
+  # config.log_level = :info
```
F3PiX: This is more a question than a memo ;-)

app/models/conference/importer.rb
```diff
@@ -0,0 +1,73 @@
...
+    def info_log(arg, file)
+      notices = {
+        started: "Started importing file #{file.original_filename}",
+        finished: "Finished updating/creating #{count_conferences_in(file)} conferences"
+      }
+      imports_logger.info "\n***" + notices[arg] + "***"
+    end
+   
+    def imports_logger
+      Logger.new("#{Rails.root}/log/imports_logger.log")
```
carpodaster: Logging to a file isn't really working on Heroku. For one, the log file will get purged due to Heroku's Ephemeral Filesystem logic and secondly, the file will only be available on the dyno the receives the import request – and we can't tell which one that will be. I would just use `Rails.logger` as the logging facility.

app/models/conference/importer.rb
```diff
@@ -0,0 +1,73 @@
...
+  # - Dates should be formatted with dd mm yyyy
+  # - UID should be unique and formatted with 'current season' + id: 2017001
+  # - The headers should not be changed
+  # Output:
+  # Conferences will be updated or created
+  # Conferences deleted in the input file, are not removed from the table
+  # Import errors are logged in a separate 'imports_logger' file
+  
+  class << self
+    def import(file)
```
carpodaster: (Disclaimer: Nitpicking / Tangent :wink: )

I always find mirroring the class name in its only public method a bit odd: `Runner.run`. _of course_ it runs, hence the name of the class :upside_down_face: 

My personal approach is mostly to name the method `call`. `run` would work, too, but `call` is available on Lambdas/Procs so the architecture - theoretically - also gets more losely coupled.

db/migrate/20170628104617_add_columns_to_conferences.rb
```diff
@@ -0,0 +1,9 @@
+class AddColumnsToConferences < ActiveRecord::Migration[5.1]
+  def change
+    add_column :conferences, :gid, :integer
```
carpodaster: What does `gid` stand for? Group ID? :) (I honestly don't know)

app/assets/stylesheets/global.sass
```diff
@@ -151,8 +151,11 @@ html, body
...
-        white-space: nowrap
+        max-width: 300px
+        //white-space: nowrap
```
carpodaster: (Commented code)

config/routes.rb
```diff
@@ -87,7 +87,10 @@
     end
     resources :teams
     resources :seasons
-    resources :conferences
+    resources :conferences, only: :show
```
carpodaster: Can you add the `do … end` block of the following line to this one and remove lines 91..93? I think otherwise you'll lose the `only: :show`

app/controllers/orga/conferences_controller.rb
```diff
@@ -1,31 +1,21 @@
...
+  
   def index
     @conferences = conferences
   end
   
   def new
-    @conference = Conference.new
   end
   
   def create
```
carpodaster: The create action is not routed anymore so the action can go away

app/controllers/orga/conferences_controller.rb
```diff
@@ -1,31 +1,21 @@
...
+  def import
+    Conference::Importer.import(params[:file])
+    redirect_to orga_conferences_path, notice: "Import finished! Check log for errors."
+  end
+  
   def index
     @conferences = conferences
   end
   
   def new
```
carpodaster: The new action is not routed anymore so the action can go away

app/controllers/orga/conferences_controller.rb
```diff
@@ -1,31 +1,21 @@
...
   
-  def update
-    if @conference.update(conference_params)
-      redirect_to orga_conference_path(@conference)
-    else
-      render action: :edit
-    end
-  end
-
   def destroy
```
carpodaster: The destroy action is not routed anymore. The implementation is still here though and I think it makes sense to be able to delete a conference. Can you add it to the `routes.rb` again ?

config/routes.rb
carpodaster: Actually, we also need `:index` and `:destroy`

app/views/conferences/_details.html.slim
```diff
@@ -1,24 +1,12 @@
...
     dt Location
-    dd = @conference.location
+    dd #{@conference.city}, #{@conference.country}
     dt Date
     dd = @conference.date_range
     dt Twitter
     dd = format_conference_twitter(@conference.twitter)
     dt Homepage
     dd = link_to @conference.url, @conference.url
-    dt Tickets total
```
carpodaster: Just for comprehending the feature: we don't work with this data anymore? If not, should we remove the corresponding database columns as well?

app/models/date_range.rb
```diff
@@ -1,7 +1,9 @@
 class DateRange
   # Step 1: only in use with Conference date range.
-  # Validated in Conference: presence and chronological order
-  #  because the input is done there.
+  # Validation is supposed to be in input (e.g. Conference.rb), but
+  # with the new Importer for adding conferences, the presence validation
+  # of the dates is disabled. Validation of chronological order is still present.
+  # Importing a Conference without dates will log an error in the imports_logger file
```
carpodaster: I would remove this line (reference to implementation detail in another file/class)

app/models/conference/importer.rb
```diff
@@ -0,0 +1,73 @@
...
+    
+    def count_conferences_in(file)
+      CSV.foreach(file.path, headers: true).count
+    end
+    
+    def process_csv(file)
+      CSV.foreach(file.path, headers: true, col_sep: ';' ) do |row|
+        begin
+          conference = Conference.find_or_initialize_by(gid: row['UID'])
+          # TODO check: when extra columns are added in the Google Doc, what happens?
```
carpodaster: See my comment `check_valid(file)`

app/models/conference/importer.rb
```diff
@@ -0,0 +1,73 @@
...
+          conference_hash = {
+            gid: row['UID'].to_i,
```
carpodaster: UID is mapped to `gid` ? Why not also `uid`?

app/models/conference/importer.rb
```diff
@@ -0,0 +1,73 @@
...
+          fetch_season_id(row['UID'])
+          conference.update!(conference_hash.merge(season_id: @season_id))
+        rescue => e
+          imports_logger.error "Error in #{row['UID']}: #{e.message}"
+        end
+      end
+    end
+    
+    def fetch_season_id(uid)
+      year = uid.to_s[0,4]
```
carpodaster: And another question about `uid` / `gid`: this seems to be some magic string. Can you add a brief comment on what the format looks like?

app/models/conference/importer.rb
```diff
@@ -0,0 +1,73 @@
...
+            city: row['City'],
+            country: row['Country'],
+            region: row['Region'],
+            url: row['Website'],
+            notes: row['Notes'],
+          }
+          fetch_season_id(row['UID'])
+          conference.update!(conference_hash.merge(season_id: @season_id))
+        rescue => e
+          imports_logger.error "Error in #{row['UID']}: #{e.message}"
```
carpodaster: I have an idea how to get rid of log files entirely but this is probably best done in a pair programming session. Let's not do it in this PR.

app/models/conference/importer.rb
```diff
@@ -0,0 +1,73 @@
...
+        end
+      end
+    end
+    
+    def fetch_season_id(uid)
+      year = uid.to_s[0,4]
+      @season_id = Season.find_by(name: year).id
```
carpodaster: Since this is a class-level method, the `@season_id` defines a (global!) class-level variable. If you want to stick to class methods, this method should just return the season without binding it to a variable.

The call to this method can either bind it to a local var or use its return value directly like so:

```ruby
conference.update!(conference_hash.merge(season_id: fetch_season_id(row['UID']))
```

app/models/conference/importer.rb
```diff
@@ -0,0 +1,73 @@
...
+      }
+      imports_logger.info "\n***" + notices[arg] + "***"
+    end
+   
+    def imports_logger
+      Logger.new("#{Rails.root}/log/imports_logger.log")
+    end
+    
+    def check_valid(file)
+      raise "Oops! I can upload .csv only :-(" unless file.content_type == "text/csv"
```
carpodaster: (Disclaimer: Nitpicking)

`raise` with just a string as the error message will raise a `RuntimeError`. Since the method has a problem with its argument, I would go for raising an `ArgumentError` instead.

spec/fixtures/files/test.csv
```diff
@@ -0,0 +1,7 @@
+UID;Name;Start date;End date;City;Country;Region;Website;Notes
+2017001;Deccan RubyConf 2017;Aug 12, 17;Aug 12, 17;Pune;Belgium;Europe;http://www.deccanrubyconf.org/;
+2017002;JSFoo 2017;Sep 15, 17;Sep 16, 17;Bangalore;India;Asia Pacific;https://jsfoo.in/2017/;
+2017003;;;;Delhi;India;Asia Pacific;https://in.pycon.org/;1st week of Nov (dates are to be defined)
+2017004;GHC India 2017;17/11/2017;15/11/2017;Bangalore;India;Asia Pacific;https://ghcindia.anitaborg.org/;They have student scholarship applications. Deadline: June 30, 2017
+2017005;PyCon Pune 2018;;;Pune;India;Asia Pacific;https://pune.pycon.org/;Dates are to be defined
+2017006;CheesyRuby;07/07/2017;15/07/2017;Amsterdam;NL;Europe;www.example.com;Duly Noted
```
carpodaster: CheesyRuby! I wanna goooo! :cheese:

app/models/conference/importer.rb
F3PiX: Because I don't like to make promises I can't keep. There is no way we can be sure of the uniqueness of the ID in the imported Google Doc. If it is called `uid`, we suggest that we can trust the uniqueness.

app/models/conference/importer.rb
F3PiX: That would be great! Now I am curious.
I also considered making it a rake task.

config/routes.rb
F3PiX: Sorry, I forgot some changed I had stashed. 🙈

app/views/conferences/_details.html.slim
F3PiX: We don't use them in this table. They will be back in one of the upcoming user stories.
I think we shouldn't remove them from the database, at least not for now. I added a note to the project backlog. 👍

carpodaster: 👍

app/models/conference/importer.rb
F3PiX: Question: Do you mind if I don't (at least  in this iteration)? I checked it with a test file. Extra columns are just ignored, the order doesn't matter (`:process_csv` takes care of that), and empty columns will be empty in the table. 
The headers in the Google doc are the headers Maria is actually using in the Doc.

carpodaster: Sure, let's do it somewhen else. Or not at all :)

db/migrate/20170628104617_add_columns_to_conferences.rb
carpodaster: Nevermind, for some reason I completely overlooked what you've written in this PRs description 🙈

F3PiX:
Thank you so much @carpodaster! 💃  So much nicer.
Sorry for the routing and actions things, I had those changes stashed and forgot to unstash. 

I'll go and try to get the specs in place now.

app/controllers/orga/conferences_controller.rb
```diff
@@ -1,30 +1,14 @@
 class Orga::ConferencesController < Orga::BaseController
-  before_action :find_conference, only: [:show, :edit, :update, :destroy]
+  before_action :find_conference, only: [:show, :destroy]
+
+  def import
+    Conference::Importer.call_with(params[:file])
```
F3PiX: LOL Ruby is all about calling methods, so _of course_ it calls :wink:
I am very tempted to go with `Importer.do_your_magic_with(file)` but I went with `call_with(file)`.  Because `call(file)` looks weird to me.
Okay? Or is this undermining the architecture argument?

spec/controllers/orga/conferences_controller_spec.rb
```diff
@@ -16,20 +16,20 @@
       end
     end
```
F3PiX: Please ignore, not ready to be pushed

spec/models/conference/importer_spec.rb
```diff
@@ -0,0 +1,22 @@
+require 'spec_helper'
```
F3PiX: Please ignore, shouldn't be pushed

app/controllers/orga/conferences_controller.rb
carpodaster: > LOL Ruby is all about calling methods, so of course it calls 😉

Haha, you won this round 🤣 

But yeah, `call` is a thing when working with lambdas / Procs:

```ruby
l = ->(cheese) { "#{cheese} is tasty" }
l.call("Camembert") # => "Camembert is tasty"
```

Rack also only expects an arity-1 `call` so that you can either concrete classes/instances reponding to it or just pass a lambda/proc.

I would really prefer `call` instead of adding another name to the mix.

config/environments/production.rb
```diff
@@ -75,7 +75,7 @@
 
   # Use default logging formatter so that PID and timestamp are not suppressed.
   config.log_formatter = ::Logger::Formatter.new
-
+
```
carpodaster: (whitespace)

carpodaster:
👍  @F3PiX can you ping me again when you have the specs ready for review?

config/environments/production.rb
F3PiX: Shoot. I changed a setting in my IDE and can't find how to set it back. Sorry,

F3PiX:
@carpodaster I can't find how to test the `Conference::Importer.call`. Can you help please?

spec/controllers/orga/conferences_controller_spec.rb
```diff
@@ -16,20 +16,17 @@
...
-        post :create, params: {conference: attributes_for(:conference)}
-        expect(assigns(:conference).season.name).to eql Date.today.year.to_s
+      it 'posts a .csv file' do
+        post :import, params: { file: file  }
+        expect(response).to redirect_to (orga_conferences_path)
+        expect(flash[:notice]).to match(/Import finished/)
       end
     end
-
+
```
F3PiX: (whitespace)

carpodaster:
@F3PiX I would work with `expect { foo }.to change { bla }` when calling `.call` with a CSV file fixture. Testworthy scenarios:

1. Creates a new record from CSV data (check that season has been properly parsed)
2. Updates existing record from CSV data
3. Input not a CSV file (`expect { foo }.to raise_error BlaError`)

F3PiX:
The main problem was that I couldn't find a way to stub the content_type of the file, but I just found out that I don't need to do that. :-)

F3PiX:
I can't do anything with the file without raising the ArgumentError/csv error. 

This is what I have now:
```
require 'spec_helper'
require 'csv'

RSpec.describe Conference::Importer do

  describe "#call" do
    let!(:file) { file_fixture("test.csv") }
    subject { described_class.call(file).read }

    it { puts file_fixture_path } # => spec/fixtures/files

    it 'works' do
      expect { subject }.not_to raise_error(ArgumentError, /.csv/)  #=> green
    end

    context 'with valid file' do
      it 'imports the conferences' do # => red : raises ArgumentError, /.csv/ 
        expect { subject } .to change { Conference.count }.by(6)
      end
    end
  end
end
```
So, is there a way to set the content_type for the test file? Or is there a way to stub that?

carpodaster:
You should be able to also use `fixture_file_upload("spec/fixtures/files/test.csv", 'text/csv')`

app/models/conference/importer.rb
```diff
@@ -0,0 +1,74 @@
...
+    end
+    
+    def count_conferences_in(file)
+      CSV.foreach(file.path, headers: true).count
+    end
+
+    def fetch_season_id(uid)
+      # uid format : 2017001
+      year = uid.to_s[0,4]
+      Season.find_or_create_by(name: year).id #Q: can I do that here?
```
F3PiX: Question: can I do that here? (It is because maybe wants to add a yearly conference, a few years ahead. RailsConf 2018, 2019, 2020)

spec/models/conference/importer_spec.rb
F3PiX: Please stop ignoring :-)
I spent a lot of time solving the wrong problem 🙈 . Sorry that it took so long. 
In the specs, I use the conferences from the test file, and now I am not sure if that is the right level of abstraction for the specs.

F3PiX: Also, I added the content of the file to the bottom of this file, for your convenience in review. I'll remove them after review.

F3PiX:
@carpodaster Back to you.

app/models/conference/importer.rb
```diff
@@ -57,17 +63,12 @@ def process_csv(file)
             url: row['Website'],
             notes: row['Notes'],
           }
-          conference.update!(conference_hash.merge(season_id: fetch_season_id(row['UID'])))
+          conference.update!(conference_hash)
```
carpodaster: You can merge the `season_id` from the line below onto `conference_hash`, saving one DB write.

app/models/conference/importer.rb
carpodaster: Sure, that should work. Is there a `!` version of `find_or_create_by`? If so, maybe use that one instead in case the first four characters of `uid` don't return a proper year.

app/models/conference/importer.rb
```diff
@@ -0,0 +1,74 @@
...
+  # Output:
+  # - Conferences will be updated or created
+  # - Conferences deleted in the input file, are not removed from the table
+  # - Import errors are logged in a Rails Logger
+  # - UID is mapped to 'gid' ('google-id' LOL), and has the format: 2017001
+  
+  class << self
+    def call(file)
+      check_valid(file)
+      info_log(:started, file)
```
carpodaster: How about this?

```ruby
def call(file)
  check_valid(file)
  with_log(file) { process_csv(file) }  # could probably bind the outer `file` to be available within the block, too. But let's not ;)
end

# …

def with_log(file)
  logger.tagged("Importer") { logger.info "Started importing file #{file.original_filename}" }
  yield
  logger.tagged("Importer") { logger.info "Finished updating/creating #{count_conferences_in(file)} conferences" }
end
```
Disclaimer: not tested, written entirely in GitHub ;)

app/models/conference/importer.rb
F3PiX: I 'undid' merging because of the spec :
```it 'assign the season_id' do
        FactoryGirl.create(:conference, gid: 2017001, season_id: nil)
        FactoryGirl.create(:conference, gid: 2018005, season_id: nil)
        s_2017 = FactoryGirl.create(:season, name: "2017")
        s_2018 = FactoryGirl.create(:season, name: "2018")

        expect{subject}.to change{Conference.find_by(gid: 2017001).season_id}.to(s_2017.id)
        expect(Conference.find_by(gid: 2018005).season_id).to eq s_2018.id
      end
```
I did not find a way to check the change in the specs as well as merging in the season_id. 
I understand that is a bad reason to do an extra db action, but I also really want to doublecheck if the season_id has been changed.
Is there a way to do that? Or should I only care that the conference has the correct season_id?

spec/controllers/orga/conferences_controller_spec.rb
```diff
@@ -25,6 +25,11 @@
         expect(response).to redirect_to (orga_conferences_path)
         expect(flash[:notice]).to match(/Import finished/)
       end
+
+      it 'refuses other formats' do
+        post :import, format: :json, params: { file: file}
```
F3PiX: I _thought_ I added an extra check (on the Content_Type) in the POST request, after checking for .csv format in the form and before checking if the error is raised in the importer.rb. 
Now I am not sure if we need this in the first place? Or, I don't understand why to _not_ check the Content-Type here.

app/models/conference/importer.rb
F3PiX: Wow, I like that very much! Thanks.

F3PiX:
It's all yours, @carpodaster. Two questions re: your previous review (in the comments).
Thanks for all the help!!

PR #776, opened 2017-07-05
Include a new intermediate state (pending) for the projects

branquinhoaa:
View before changes:
before_changes

View after first changes:
after_changes

View after changing the action column to one:
only_one_col

- Include a new state for the project's model (pending)
- Include a new action to change the project to this state (action pending in projects controller)
- Change the reject and accept states to transiting from pending (not from proposed)
- Change the project fixture to transiting to accept and reject only after start_review.

### DIFF
--- a/app/controllers/orga/projects_controller.rb
+++ b/app/controllers/orga/projects_controller.rb
@@ -14,7 +14,7 @@ class Orga::ProjectsController < Orga::BaseController
     redirect_to [:orga, :projects]
   end
 
-  def pending
+  def start_review
     if @project.start_review!
       flash[:notice] = "Successfully marked project as pending!"
     else
--- a/app/controllers/orga/projects_controller.rb
+++ b/app/controllers/orga/projects_controller.rb
@@ -14,6 +14,15 @@ class Orga::ProjectsController < Orga::BaseController
     redirect_to [:orga, :projects]
   end
 
+  def pending
+    if @project.start_review!
+      flash[:notice] = "Successfully marked project as pending!"
+    else
+      flash[:alert]  = "There has been an error marking this project as pending."
+    end
+    redirect_to [:orga, :projects]
+  end
+
   def reject
     if @project.reject!
       flash[:notice] = "Project has been rejected!"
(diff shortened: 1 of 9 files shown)


carpodaster:
Awesome, first PR already! I'll look into it tonight (=in a few hours)

app/models/project.rb
```diff
@@ -21,13 +21,18 @@ class Project < ActiveRecord::Base
     state :proposed, initial: true
     state :accepted
     state :rejected
+    state :pending
+
+    event :start_review do
```
carpodaster: Great naming choice, I like it very much!

app/controllers/orga/projects_controller.rb
```diff
@@ -14,6 +14,15 @@ def accept
     redirect_to [:orga, :projects]
   end
 
+  def pending
```
carpodaster: Shall we change the name of the controller action to correspond to the acts-as-state-machine state, i.e. `start_review`?  I don't mind either option so I let you decide.

app/views/orga/projects/index.html.slim
```diff
@@ -10,7 +10,7 @@ table.table.table-striped.table-bordered.table-condensed.hidden-sm.hidden-xs
       th Submitted by
       th State
       th Tags
-      th colspan=3
+      th colspan=5
```
carpodaster: I wonder if we should display the state-related actions all in the same column; there will only ever be one button to change the state of the project record, so we might as well have them in the same `<td>` all the time. What do you think?

app/controllers/orga/projects_controller.rb
branquinhoaa: Yeah, it makes sense the action be renamed to start_review.

app/views/orga/projects/index.html.slim
carpodaster: @branquinhoaa is `colspan=5` still valid with your most recent changes?

carpodaster:
@branquinhoaa can you add another screenshot of how the action buttons look now, please? Thanks in advance! 🎁

carpodaster:
Ah, I only just realized that you've added the screenshot to the initial PR description. Looks perfect, thank you!

PR #777, opened 2017-07-05
redirect user to his page after update and creation

branquinhoaa:
- When update or create user, if successful, it will be redirected to user's page.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -38,7 +38,7 @@ class UsersController < ApplicationController
 
     respond_to do |format|
       if @user.save
-        format.html { redirect_to params[:redirect_to].blank? ? @user : params[:redirect_to], notice: 'User was successfully created.' }
+        format.html { redirect_to params[:redirect_to].presence || @user, notice: 'User was successfully created.' }
         format.json { render action: :show, status: :created, location: @user }
       else
         format.html { render action: :new }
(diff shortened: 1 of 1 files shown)


app/controllers/users_controller.rb
```diff
@@ -38,7 +38,7 @@ def create
 
     respond_to do |format|
       if @user.save
-        format.html { redirect_to params[:redirect_to] || @user, notice: 'User was successfully created.' }
+        format.html { redirect_to params[:redirect_to].blank? ? @user : params[:redirect_to], notice: 'User was successfully created.' }
```
carpodaster: That works!

However, there is a coding style referred to as _Tell, Don't Ask_ which can be applied here. What it means is that _asking_ a data structure something and then deciding on what to do can oftentimes be refactored to something where you just trust a data structure to know what to do when you send it a message (= calling a method on it). Here, the ternary clause _asks_ if the parameter is `blank?`  and branches off into two logical paths.

The previous implementation attempted to make use of the fact that `nil` is falsey. The `||` would use its left hand value first if it was something truthy, otherwise its right hand side would be chosen as the redirection target.

The bug that causes the behaviour is that `params[:redirect_to]` is actually never `nil` (you can inspect the values being send to the `users_controller#update` action to see what actual value it has).

Going back to the original implementation with the `||` operator, you can fix the bug using [a neat extension to Ruby objects provided by Rails' ActiveSupport](http://guides.rubyonrails.org/active_support_core_extensions.html#presence).

branquinhoaa: Nice!! I did not know this method... really cool. I saw that the parameter is not nil, but it is empty. I will fix it now. Thanks!

carpodaster:
Boom. Congrats on your first merged PR to this project :balloon: :smile_cat:

Issue #760, opened 2017-06-12
Remove 'event' views


F3PiX:
We have a directory with views for events, but there is no model and no controller. 

Can we remove the views? Or should we add the backend?

alicetragedy:
+1 for removing it; it's an old thing that was put in place to have the different seasons (see [here](https://github.com/rails-girls-summer-of-code/rgsoc-teams/commit/3d2298a8f4b6945a3112a9be2ff182ad3a770383)) but was later [removed](https://github.com/rails-girls-summer-of-code/rgsoc-teams/commit/e14322e421d3ab2b540db9170894182de2120177). I can only assume the view was accidentally left there. :)
Do you want to take this?

F3PiX:
I am collecting some beginner friendly issues. Not hard to do, but nice starters to see how our PR and code reviews work. :-)

carpodaster:
That is most awesome, @F3PiX ! 🌷

alicetragedy:
yeah that's perfect @F3PiX! Was going to add the `beginner-friendly` label but I saw that you beat me to it ;)

carpodaster:
It would seem that [someone](https://github.com/carpodaster) didn't do their job properly when [deleting code](https://github.com/rails-girls-summer-of-code/rgsoc-teams/commit/e14322e421d3ab2b540db9170894182de2120177).

![](https://images.duckduckgo.com/iu/?u=http%3A%2F%2Fstarecat.com%2Fcontent%2Fwp-content%2Fuploads%2Fwhen-you-delete-a-block-of-code-that-you-thought-was-useless-pink-panther-gif-animation.gif&f=1)

PR #780, opened 2017-07-06
remove the event views directory

branquinhoaa:
- Simply remove the event directory in the views as asked in the issue #760


carpodaster:
Awesome, thank you!

Issue #782, opened 2017-07-07
Show Off the Projects! Page 


F3PiX:
@anikalindtner requested a page that shows all the accepted projects with their tags.

I pushed a preliminary PR  #783 , with just the name, the url and the tags.
This is the minimal form, it surely has possibilities.
- [ ] Find a suitable icon (instead of the current 'users' icon)
- [ ] Check (with @anikalindtner) if it makes sense to add more info like: 
  - [ ] the teams that works on a project
  - [ ] the number of years the project is part of the SoC 
  - [ ] the level : beginner or not.
- [ ] Should it be accessible by the menu? All year, or during the summer only? 
- [ ] Add seeds

Another thing I found is that the ./projects#index page is publicly accessible. That page shows the number of first/second choice applications, and has links to the project page that shows the comments. 
I don't think all of that is meant to be publicly available? Although the index page only shows the accepted teams, it _may_ be possible to access the rejected projects via the url.
- [ ] How much of this info should be publicly available?
- [ ] Should the show views for rejected projects protected? Are they?

You can split this issue in several partial issues, if you want to work on part of them.

F3PiX:
This was premature. Closing until we know more.

PR #783, opened 2017-07-07
Quick Fix for a page with projects that have teams working on them

F3PiX:
It shows the :accepted projects in the current season, with the name of the project, the website and the tags.
Further Improvements are listed in issue #782 

I found that the {root}/projects  page, that lists all the (accepted) projects and all the details (including first /second choice applications, comments etc) are publicly available.
I guess that is not meant to be that way?
As a temporary solution, I re-routed `./projects` to `show_projects#index`. 

That means: the old projects show pages are not really protected from  ./projects/:id. And the `orga/projects` routes still show the extended projects page. 

Can we go with this solution first?


app/controllers/show_projects_controller.rb
```diff
@@ -0,0 +1,6 @@
+class ShowProjectsController < ApplicationController
+  
+  def index
+    @projects = Project.in_current_season.where(aasm_state: :accepted)
```
klappradla: If I recall correctly, AASM automatically adds a scope for each state. So you could shorten this to the even more expressive form of:
```ruby
@projects = Project.in_current_season.accepted
```

app/controllers/show_projects_controller.rb
```diff
@@ -0,0 +1,6 @@
+class ShowProjectsController < ApplicationController
```
klappradla: Hm, I think I don't really get the point here 🙈 
There is already a public projects page - which in my opinion also makes sense, students applying need to have an overview of projects available.

So what is the difference in information this new page now offers? If it's something only a specific group of users should be allowed to see, then I'd vote for a nested controller in the respective users' role namespace. If this is the general *"project page"* than I'd say it's fine to be non-restricted and therefore not nested into any role namespace.

config/routes.rb
```diff
@@ -3,6 +3,8 @@
 
   root to: 'activities#index'
 
+  get 'projects', to: 'show_projects#index'
```
klappradla: Again, I'd vote for making this a `resources` route and eventually nesting it into the respective namespace.

spec/controllers/show_projects_controller_spec.rb
```diff
@@ -0,0 +1,11 @@
+require 'spec_helper'
+
+RSpec.describe ShowProjectsController, type: :controller do
+
+  describe "GET #index" do
```
klappradla: I think it would make sense to have a few more tests here. One could for instance test for the `:index` template to be rendered or, in my opinion even more meaningful for the two scopes being called on `Project`.

The latter would again be something like
```ruby
projects_of_season = double(accepted: [])
expect(Project)
  .to receive(:in_current_season)
  .and_return(projects_of_season)
expect(projects_of_season)
  .to receive(:accepted)
```

Note that the above may not be executable, I just wrote it as an inspiration out of memory.

F3PiX:
Thanks Max! Interesting questions. Anika asked for " a page like this, with only the selected projects: [teams.railsgirlssummerofcode.org/projects](https://teams.railsgirlssummerofcode.org/projects), " for sponsors to see the projects and technologies.
I thought that that page showed too much information. But maybe it is not? I'll check with Anika.
 
Also, AFAIK it is not that urgent that it can't wait until Monday. I'll work on the improvements in the meantime. 💪

klappradla:
Oh ok, sound fine 👍 

I don't know the general plans on what information should be visible for which groups of users. I just initially would have suspected the projects to be "all public", since it's a) a core part of the program and b) I wouldn't know any privacy concerns there (it's links to public Github Repos...).

F3PiX:
Turns out we need to _filter_ the projects, to show the projects that have RailsGirls working on them only.
Here is a intermediary solution:
- From May - Sept (= `Season.active`) show the filtered ^ projects list
   Show the regular selection the rest of the year
- I used the applications table as a join table, to map the `projects.name` to the `teams.projects_name`.
  Instead, I'd rather add a foreign key to the teams table with the `project_id`. TBH I was surprised there wasn't. Is that for a reason I am not aware of? Or shall I do it?
  One of the disadvantages of the complicated query is that it is hard to get the test right.
 
So, before breaking my head over the specs, I wanted to check if the foreign key solution isn't better. 
Also, because I won't have time to work on this until late in the afternoon, is this safe enough a solution to merge it? And follow up with the PR with specs and maybe the foreign key solution.

klappradla:
Yes @F3PiX as I said before, I'm totally fine with merging an intermediate-solution if someone relies on the functionality to be in place, as long as someone is working on a succeeding PR 👍 
So feel free to merge it whenever you need it.

On your question about the foreign key: if I recall correctly it once was in place, but not used or only being set if manually done via a "hidden" form 😉  But I don't remember any details on why it is no longer there. It probably has to do with the projects' IDs only being part of the `application_data` htstore...

But I can take a look tonight if I find out what happened to it.

klappradla:
Just for quick notice @F3PiX : I had a look at production data
```ruby
Application.pluck(:project_id)
```
returns only `nil`, so the query you had in mind won't work. 

The `team` references its project - but sadly only by name (`project_name`) 😞 

But afaik this should be set for all teams that got accepted and got a project assigned.

So this query for instance would give you all projects that are worked on right now:
```ruby
project_names = Team
  .in_current_season
  .accepted
  .pluck(:project_name)

Project
  .in_current_season
  .where(name: project_names)
```

Based on the `project_name` as the only association between Team and Project one could even build up an association 🙈 Looks something along the lines of this as far as I remember (absolutely no guarantee):
```ruby
# in team
has_one :project, foreign_key: 'name', primary_key: 'project_name'
```

F3PiX:
O thanks!! With this info, I dived in the code again, and AFAIK the project_id is never been _saved_ on the Application 🤔  Which explains why the production db returns nil. 
Great query. I'll push that in a minute. 
I'll save the association for the follow-up 👍  #784

Issue #784, opened 2017-07-10
Sustainable solution for Projects Page (#783)


F3PiX:
#783 provides a quick fix projects page. 
Make it sustainable!

- [ ] Either move the application query to the application model or remove it and add a project_id to the teams table.
- [ ] Add tests
- [ ] Check if the 'seasonal' filter should stay, if not provide a different solution .

F3PiX:
Max said:
> On your question about the foreign key: if I recall correctly it once was in place, but not used or only being set if manually done via a "hidden" form 😉 But I don't remember any details on why it is no longer there. It probably has to do with the projects' IDs only being part of the application_data htstore...

The project_id was removed with the introduction of the Projects Proposal feature, here: 
https://github.com/rails-girls-summer-of-code/rgsoc-teams/commit/7bcef84b42c13858a4cb106a8cab22f42dab6a59.

klappradla:
Thanx for researching where the `project_id` got lost @F3PiX 🙏 

If it's really no longer used, I'd vote for a) either renaming the column to `deprecated_project_id` and then removing it later or b) straight away removing it *(in favor)*. This can in my opinion even be done within the scope of this issue - it's just a tiny migration after all.

The way it is set up now, it's only confusing people 🙈

F3PiX:
Maybe the `project_id` do exits for older applications in the db. (From before the projects proposal feature was implemented.) Wouldn't that be a problem?

klappradla:
Well, I personally don't consider it a problem. Previous seasons' data model don't work with the current schema and app anyways - we have snapshots of all previous years for this.

F3PiX:
@anikalindtner 
At the new 💥 projects page, we display the 'active' projects (= accepted projects, with teams working on them during SoC) during the 'active' season (that is: from the day the application confirmations are sent, until the last day of the coding summer.) 
The rest of the year, we show the accepted and proposed projects. 
Can we keep it that way? 
If so, I'll close this issue.

alicetragedy:
@F3PiX may I jump in with some thoughts?
I like the idea of still showing all projects that have been submitted, regardless of whether they are participating with a team or not. I'm not sure if in this case making it 
[...]

PR #786, opened 2017-07-10
Project Page, second try

F3PiX:
The query in the former PR #783 returned an empty table in production. Caused by querying the `applications` table for `project_id`. 
This query (by @klappradla ) does not use the applications table, so should work.
YOLO

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -31,6 +31,11 @@ class Project < ActiveRecord::Base
     end
   end
 
+  def self.selected
+    project_names = Team.in_current_season.accepted.pluck(:project_name).uniq
+    Project.in_current_season.where(name: project_names)
+  end
+
   def taglist
     tags.join(', ')
   end
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -24,7 +24,11 @@ class ProjectsController < ApplicationController
   end
 
   def index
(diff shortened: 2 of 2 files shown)


app/models/project.rb
```diff
@@ -31,6 +31,11 @@ class Project < ActiveRecord::Base
     end
   end
 
+  def self.selected
+    project_names = Team.in_current_season.accepted.pluck(:project_name).uniq
```
klappradla: Let's skip the `uniq` here - it will convert the whole thing to an Array (and actually execute the query), which is not really necessary (there can be loads of duplicates, the final resulting Projects will be the same :v:)

app/models/project.rb
```diff
@@ -31,6 +31,11 @@ class Project < ActiveRecord::Base
...
+    Project.in_current_season.where(name: project_names)
```
klappradla: I'm not 100% sure if it's necessary, but I'd have scoped it even more to this:
```ruby
Project.in_current_season.accepted.where(name: project_names)
```
This makes sure, we really only look into the accepted projects (maybe there's a rejected one with the same name...).
What do you think @F3PiX ?

F3PiX:
I didn't think of that. 🙈 
Can it wait for #784 (the sustainable solution)? I'll take that one after I finished the Conferences Importer.

Issue #772, opened 2017-06-24
[Conferences] #8: Accepting/rejecting a conference


mkalininait:
The idea: https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/751#issuecomment-308574743
Full list of conference user stories: https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/751#issuecomment-309138513

**User Story:**
As an RGSoC orga, I want to know if the students accept our offer or not.

**Things to do:**
1. Who can edit a conf status:
 - Students of the team
2. Who can see a conf status:
 - Students of the team
 - Supervisors
 - Orga
3. Available from:
 - Team view, h4 Conference offers (see Task #771)
4. Next to each offered conference show two links: "Yes, I will attend!" and "No, I cannot attend". You can press one link, but then change your mind and press another one (see how it looked like before: https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/789#discussion_r128508734)
5. Add new fields to the conference export (see Task #770):
 - Conference offer: Value (conference name)
 - Conference offer accepted: yes/no
 - Conference offer #2:  Value (conference name)
 - Conference offer #2 accepted: yes/no
 - etc

branquinhoaa:
@mkalininait How do you think that we can display the conf. status?
I did like this:

![conf-attendance](https://user-images.githubusercontent.com/15239141/29319856-d0dc25bc-81ab-11e7-87e7-ec7687d9aa3d.png)

Will we change the links style after the team had clicked on attend link?

Also, I think we need to merge the PR  #813 in order to finish this one (the export change part).

klappradla:
Just a short note to your screenshot above ☝️ @branquinhoaa: I think it would increase the readability if you'd always add a whitespace 🏳️ between the link and the label.
And for the orga comment, I guess it would make sense to add a linebreak and make the comment a [blockquote](https://getbootstrap.com/docs/3.3/css/#default-blockquote)

For the conference-links: I'd suggest to use a normal [list](https://getbootstrap.com/docs/3.3/css/#unstyled) without bullets (I don't see the point for them here...)

For the attendance-links: they're currently not distinguishable from the conference-links, even though they're a totally different call to action _(they change state, rather then "just" inform)_. It would make sense to have this difference somehow visible, e.g. by making them small buttons instead of just links.

[...]

Issue #771, opened 2017-06-24
[Conferences] #7: Assigning a conference


mkalininait:
The idea: https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/751#issuecomment-308574743
Full list of conference user stories: https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/751#issuecomment-309138513

**User Story:**
As an RGSoC orga, I want to see team's conference preferences and assign one or multiple conferences from the longlist which are actually willing to give them free tickets (with an optional comment).

**Things to do:**
1. Who can assign a conference:
 - Orga
2. Who can see offered conferences:
 - Students of this team
 - Supervisors
 - Orga
3. On the team edit form add:
 - h3 Conference offers
 - (dropdown) Offered conference (gets values from the conference longlist)
 - (button) Add another conference (adds another dropdown "Offered conference #2" etc) *
 - (text area) Comment (optional)
 - Change h2 Conference preferences to h3 Conference preferences
 - Add h2 Conferences (above h3 Conference preferences and h3 Conference offers)
4. On the team view page add:
 -  h4 Conference offers
 - Offered conference: Value
 - (if present) Offered conference #2: Value (etc) * 
 - Orga comment
 - Add h3 Conferences (above h4 Conference preferences and h4 Conference offers)

\* ⚠️ In most cases there will be only one offer, in some cases two, in very very very rare cases more than two. To make things simpler, we can implement two fields "Conference offer" and "Conference offer 2" instead of making the button.

Additionally, please pay attention to this comment: https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/771#issuecomment-315903651

klappradla:
Hey @mkalininait ,

I'm pretty late to the party, but @F3PiX brought this up and I think it would make sense to at least consider it before we this issue is picked up.

In order to implement the "assigning conferences" feature you describe here, it could make sense to migrate the current `Attendance` model to a `ConferencePreference` model. The way this is currently set up and being implemented in #789, students pick their preferred conferences within their settings / profile pages, however, under the hood the created entity is already called `Attendance`, even though it's *"just"*  a **"preference"** for now 😉 
Having create "attendances" for their preferences sound semantically wrong and is already causing some confusion in #789 ...

[...]

Issue #767, opened 2017-06-24
[Conferences] #2: Students' preferences


mkalininait:
The idea: https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/751#issuecomment-308574743
Full list of conference user stories: https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/751#issuecomment-309138513

**User Story:**
As a student, I can choose my team's primary and secondary choices from the conferences longlist.

**Things to do:**
1. Move the Conference block from User to Team.
2. Who can see the Conference block:
 - Students of this team
 - Supervisors
 - Orga
3. Who can edit the Conference block:
 - Students of this team
 - Orga
4. UI:
Block title: Conference preferences
Field 1 (dropdown): Conference (primary choice)
Field 2 (dropdown): Conference (secondary choice)
5. Contents of the dropdowns: longist of conferences (see Issue #762). If would be super great, if we could visually split the list into sub-lists according to the conferences' regions (Africa, Asia Pacific, Europe, North America, South America).

branquinhoaa:
I want to do this one, but we will need the previously story ready, right? We will need the list to choose the options. Or maybe not, we can put fake data on the two dropdown and start from there.

mkalininait:
Hi @branquinhoaa Yes, you are right, in order to complete this task, we need to have the list of conferences finished. But I agree that it should be possible to start with the things you listed. Plus you can implement permissions: only students who belong to a team can see and edit the entire Conference block  on the team's page. Unfortunately we can't provide conference tickets to coaches.

I'm not sure I understood your question about attendances. What do you mean? Do you refer to the user stories 7 and 8? If so, then yes, we will assign conferences which are willing to provide us tickets to teams as well.

PR #722, opened 2017-03-30
Adds structure to Application View

mkalininait:
Covers https://github.com/rails-girls-summer-of-code/selection_process/issues/51, "Split the application page into sections (and add Titles) like on the Application form"

@beanieboi I removed your `hr`s from the **Submitted application** section. I added sub-section titles there, and `hr` started adding a bit too much structure. I hope you don't mind.

I added `hr`s between different sections in the application view though. I don't know if it improved the view or not (not a designer) :D

PR #789, opened 2017-07-12
Conferences/students preferences

branquinhoaa:
- remove the relationship between user and conferences through attendances; 
- create a relationship between teams and conferences through attendances;
- remove conference options from user page;
- include conference options in teams page;
- include an option field in attendance model (to allow conferences prioritize choice)
- fix tests and create new tests 

teams\:id\edit
teams_edit


klappradla:
Ace @juuh42dias & @branquinhoaa 🎉 

Would you mind adding a reference to the issue(s) these changes relate to? It would help me to get  a better overview of the "problems" your PR is trying to solve :v:

juuh42dias:
All these commits are of  
#767

app/models/user.rb
```diff
@@ -215,6 +211,10 @@ def self.search(search)
     (q_user_names + q_team_names).uniq
   end
 
+  def student_team
+    self.teams.last if self.student?
```
klappradla: I'm not sure if this can guarantee to always select the "right" team... as far as I remember, `team` (hopefully) does not have a default scope.

What would be the team you want to retrieve here? The "current" one as in the one active in this current season? Why not use
```ruby
teams.in_current_season.last if student?
```

app/views/orga/teams/show.html.slim
```diff
@@ -74,12 +74,11 @@ nav.btn-group
...
   p = @team.supervisors.any? ? link_to_team_members(@team, :supervisor) : content_tag(:em, 'Will be added by RGSoC Organizers')
 
-- if @team.members.any? { |user| user.conferences.in_current_season.any? }
+- if @team.conferences.in_current_season.any?
   h4 Conferences
   ul.conferences
-    - @team.members.each do |user|
-      - if user.conferences.in_current_season.any?
-        li = "#{link_to user.name, user}: #{links_to_conferences(user.conferences.in_current_season).join(', ')}".html_safe
+      - if @team.conferences.in_current_season.any?
```
klappradla: Maybe I don't get it, but I think you can spare this line, since you already have this condition for the whole block 3 lines above.

app/views/orga/teams/show.html.slim
```diff
@@ -74,12 +74,11 @@ nav.btn-group
   h3 Supervisors
   p = @team.supervisors.any? ? link_to_team_members(@team, :supervisor) : content_tag(:em, 'Will be added by RGSoC Organizers')
 
-- if @team.members.any? { |user| user.conferences.in_current_season.any? }
+- if @team.conferences.in_current_season.any?
   h4 Conferences
```
klappradla: If I recall correctly, for the if-condition to actually work, these two lines (the h4 and ul) need to be indented one more level.

app/views/orga/teams/show.html.slim
```diff
@@ -74,12 +74,11 @@ nav.btn-group
...
   ul.conferences
-    - @team.members.each do |user|
-      - if user.conferences.in_current_season.any?
-        li = "#{link_to user.name, user}: #{links_to_conferences(user.conferences.in_current_season).join(', ')}".html_safe
+      - if @team.conferences.in_current_season.any?
+        li = "#{link_to @team.name, @team}: #{links_to_conferences(@team.conferences.in_current_season).join(', ')}".html_safe
```
klappradla: I did not look at this in the browser, maybe I'm getting it wrong. But in my understanding this should be a list of conferences. So something like:
```ruby 
ul
  li conference_1
  li conference_2
  # etc.
```

The ways your view code looks to me, this will only render a single `li` item (will all conferences in it...).

app/views/teams/_form.html.slim
```diff
@@ -69,6 +69,14 @@
               = s.link_to_remove 'Remove', class: 'btn btn-default'
       .form-group.form-btn-group
         = f.link_to_add 'Add another source', :sources, class: 'btn btn-primary col-sm-offset-3'
+  -if can?(:update_conferences, @team)
```
klappradla: This is probably personal preference, but I find "conferences" here a bit misleading, since the form is actually about "attendances" to conferences... I'd also suggest to more stick to the pattern of looking at the single user's abilities rather then the team - it's a in the end a single user who submits the form 😉 

Do you have a better idea for expressing this?

app/views/teams/show.html.slim
```diff
@@ -86,12 +86,11 @@ nav.page
     h3 Supervisors
     p = @team.supervisors.any? ? link_to_team_members(@team, :supervisor) : content_tag(:em, 'Will be added by RGSoC Organizers')
 
-- if @team.members.any? { |user| user.conferences.in_current_season.any? }
+- if @team.conferences.in_current_season.any?
```
klappradla: Same as for the orga teams view - I think there's something wrong with the list layout.

spec/controllers/teams_controller_spec.rb
```diff
@@ -184,7 +184,29 @@
 
             expect(Role.last.user).to eql existing_user
           end
+        end
+
+      #TODO: refactor this test
+      context 'selecting the conference options' do
+        let(:conference_1) { FactoryGirl.create(:conference, :in_current_season)}
+        let(:conference_2) { FactoryGirl.create(:conference, :in_current_season)}
+        let(:team)         { FactoryGirl.create(:team, :in_current_season) }
```
klappradla: Is it really necessary to create another (unrelated) team? If you already created a student, you could also just use this existing team I'd say:
```ruby
let(:team) { student.teams.first }
```

spec/controllers/teams_controller_spec.rb
```diff
@@ -184,7 +184,29 @@
...
+        let(:conferences) do
```
klappradla: Personal preferences, but I find the name a bit misleading... How about "team_params"?

spec/controllers/users_controller_spec.rb
```diff
@@ -73,8 +73,8 @@
     end
 
     context 'with conferences' do
-      let!(:attendance) { FactoryGirl.create :attendance }
-      let(:user)        { attendance.user }
+      let!(:attendance) { FactoryGirl.create :attendance, :student_attendance }
+      let(:user)        { attendance.team.members.first }
```
klappradla: Minor thing, but since you're looking at a "student", it would be more explicit to use
```ruby
attendance.team.students.first
```
here.

spec/factories/attendances.rb
```diff
@@ -1,6 +1,14 @@
 FactoryGirl.define do
   factory :attendance do
```
klappradla: Could you on the way please also fix the indentation for this factory? It's kinda hard to understand like this.

spec/models/ability_spec.rb
```diff
@@ -143,29 +145,28 @@
...
-      describe "a user should not be able to mark another's attendance to a conference" do
-
-        context 'when same user' do
-          let!(:attendance) { FactoryGirl.create(:attendance, user: user)}
+          let!(:user) { FactoryGirl.create(:student) }
+          let!(:attendance) { FactoryGirl.create(:attendance, team_id: user.student_team.id) }
 
           it 'allows marking of attendance' do
-            expect(ability).to be_able_to(:crud, attendance)
+            expect(ability).to be_able_to(:crud, user)
```
klappradla: What is this supposed to test? The expectations reads different to its description.

spec/models/ability_spec.rb
```diff
@@ -143,29 +145,28 @@
...
           end
 
-
           context 'when user is admin' do
             let!(:organiser_role) { FactoryGirl.create(:organizer_role, user: user)}
             it "should be able to crud attendance" do
-              expect(subject).to be_able_to :crud, attendance
+              expect(subject).to be_able_to(:crud, organiser_role)
```
klappradla: Again, I don't understand what this is testing 🙈 The description reads different to the expectation...

spec/models/ability_spec.rb
```diff
@@ -274,6 +275,14 @@
               expect(subject).to be_able_to :crud, student_team
             end
 
+            it 'allows select conferences on own team' do
+              expect(subject).to be_able_to :update_conferences, student_team
+            end
+
+            it 'does not allow select conferences on own team' do
```
klappradla: I think this should be something along
> does not allow to select conferences for other teams

spec/models/user_spec.rb
```diff
@@ -328,6 +326,14 @@
     end
   end
 
+  describe '#student_team' do
```
klappradla: Since this is a method on the `User` model, it would be cool to also have a scenario for when the user is not a student or has no team. What do you think?

spec/models/user_spec.rb
```diff
@@ -328,6 +326,14 @@
...
+     it 'return student current team' do
+      team    = FactoryGirl.create(:team)
```
klappradla: Afaik, it's not necessary to actually "create" all the records here. Building should be fine.

app/controllers/attendances_controller.rb
```diff
@@ -13,7 +13,7 @@ def destroy
   private
 
     def attendance
-      current_user.attendances.find(params[:id])
+      current_user.student_team.attendances.find(params[:id])
```
branquinhoaa: hey @klappradla, we are wondering that maybe this attendance controller it is not being used. We create/update attendances through the nested field in edit team action. Unless you need this data for some API...

klappradla: 😂 well I've never touched this area of the app. But if that's true, and the controller does not have **any** touching point with the rest of the app, than it should probably be removed altogether.

Can you maybe explore this a bit deeper if you can really confirm it's not being used?

spec/models/ability_spec.rb
branquinhoaa: That really doesn't make any sense. Forgive us. :joy:

spec/models/ability_spec.rb
branquinhoaa: It took a while for us to understand how this works. :joy_cat: sorry

klappradla: No worries ✌️ 
Do you need any more explanation on it or is everything clear now?

spec/models/ability_spec.rb
klappradla: No worries 😉 
Just tell me if you need any more information on how these tests are set up.

app/controllers/attendances_controller.rb
branquinhoaa: In fact, I found one place where this controller is used. I created 2 more test scenarios to simulate a user who is not a student and a student that is not in a team to try update attendance.
Now, I believe that is throwing the right error message.

app/views/orga/teams/show.html.slim
branquinhoaa: The additional indentation is not necessary (we tested)

app/views/teams/show.html.slim
branquinhoaa: ok, fixed.

spec/models/user_spec.rb
branquinhoaa: Excellent idea! we have added these scenarios.

spec/factories/attendances.rb
branquinhoaa: yeah, sorry about that. I think now is ok, right?

app/controllers/attendances_controller.rb
branquinhoaa: @klappradla  In fact, I found one place where this controller is used. I created 2 more test scenarios to simulate a user who is not a student and a student that is not in a team to try update attendance.
Now, I believe that is throwing the right error message.

branquinhoaa:
@klappradla  the corrections were made.

klappradla:
Yay, gonna have another look tonight @branquinhoaa 🎉 
*(don't know what time this will be in your parts of the world thought)*

juuh42dias:
kkkkkk here (Time.now) 2017-07-18 10:11:15 -0300

app/controllers/attendances_controller.rb
klappradla: Oh, ok, sorry, I wasn't clear on what I actually mean 🙈 It does not *necessarily*  need to be especially this certain exception. I just meant: if someone not being a student or on a team can access the controller action, the resulting error should somehow make clear what's wrong - and a 500 status could be anything 😉 

The way you put it now, sure it works, but it looks kinda weird: having the `student_team` fetched in two places and then explicitly raising the exception appears wrong in a controller, don't you think?

In which context is this controller used? If it's only used by the user actually owning the `attendance`, it could be simplified to
```ruby
def attendance
  current_user.attendances.find(params[:id])
end
```

If the user is also changing his or her teammates attendances, then one could maybe work with the abilities? Would look somewhat like this I think:
```ruby
class AttendancesController < ApplicationController
  load_and_authorize_resource

  def update
    @attendance.update_attributes!(attendance_params)
    redirect_back fallback_location: user_path(current_user), notice: 'Ok, successfully updated'
  end

  def destroy
    @attendance.destroy
    redirect_back fallback_location: user_path(current_user), notice: 'Ok, the ticket was added back to the pool of available tickets'
  end

  private

  def attendance_params
    params.require(:attendance).permit(:confirmed)
  end
end
```

I would of course prefer the first scenario :v:

db/migrate/20170707145338_remove_columns_from_attendance.rb
```diff
@@ -0,0 +1,11 @@
+class RemoveColumnsFromAttendance < ActiveRecord::Migration[5.1]
```
klappradla: Not necessary to change anything here. Just wanted to note that the indentation looks a bit off 👮 

Maybe check your editor config ✅

spec/controllers/teams_controller_spec.rb
```diff
@@ -184,7 +193,28 @@
 
             expect(Role.last.user).to eql existing_user
           end
+        end
 
+      #TODO: refactor this test
```
klappradla: Let's either do it or silently remove the comment 😉

spec/controllers/teams_controller_spec.rb
```diff
@@ -184,7 +193,28 @@
...
+            '0'=>{
+              option: 1, conference_id: conference_1.id
+            },
+            '1'=>{
+              option: 2, conference_id: conference_2.id
+            }
+          })
+        end
+
+          it 'team student user' do
```
klappradla: Please come up with a more meaningful description for the test and fix the intendation - it makes it kind of hard to read for someone new to it.

spec/factories/attendances.rb
klappradla: Perfect :tada: thanx!

app/controllers/attendances_controller.rb
branquinhoaa: @klappradla  I saw this controller being used just when a user (now the team) has to confirm her participation in the conference (after choosing the preferences). So, theoretically, just the students from the team should have access to them. The details around this will be developed better in this issue: https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/772e.

We have included the student_team in the private method attendance because the attendance belongs to the team, not to the user. But this page couldn't be visited by any one who is not a student from that specific team.

I think that this can change with the issue #772. What do you think?

branquinhoaa:
@mkalininait, could you take a look? we believe that now it is working. :dancer:

branquinhoaa:
@klappradla Could you take a look too? we have changed the things you asked for

mkalininait:
Hey @branquinhoaa What do you think about implementing @klappradla 's suggestions here https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/771#issuecomment-315555001 ?

app/models/ability.rb
```diff
@@ -25,6 +25,14 @@ def initialize(user)
       user.admin? or signed_in?(user) && team.new_record? or on_team?(user, team)
     end
 
+    can :update_attendances, Team do |team|
+      user.admin? or (user.student? && (team.students.include? user))
+    end
+
+    can :confirm_attendance, Team do |team|
```
mkalininait: As we discussed before, we don't need to confirm conference *preferences*.

app/views/orga/teams/show.html.slim
mkalininait: Can we rename it to "Conference preferences" please?

app/views/teams/show.html.slim
```diff
@@ -86,12 +86,20 @@ nav.page
...
   h4 Conferences
```
mkalininait: Can we rename this to "Conference preferences" please?

app/models/ability.rb
branquinhoaa: no, but this one is to team confirm its attendance, not its preference: 
attend
We check the ability to confirm attendance in the teams show view. When some user access the show team page, we check if the user can confirm attendance. If it can, we show these links.

mkalininait:
@branquinhoaa I've tested the PR again; it works well for me! A couple of issues I have:
1) The one I mentioned in my previous comment (https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/789#issuecomment-316636767).
2) Students should have the possibility to confirm or reject conferences yet, because they only fill out their *preferences* at this point. Their confirmation will be necessary in Issues #771 and #772. 
3) It would be great to rename the title "Conferences" to "Conference preferences" in the team views (both orga and non-orga). Or maybe make a subtitle "Conference preferences"... It can also be considered as part of 1).

app/models/ability.rb
mkalininait: My point was that we don't need these links at all here.

branquinhoaa:
Nice that it worked well!! :dancer:  
1. I think it is a great idea. Do you want us to do that in this PR?
2. The possibility for confirmation is it already there, but it is only visible for students from that specific team (we put this option on the action team show) 
attend
3. ok!

app/models/ability.rb
mkalininait: Students filled out their preferences, but we didn't offer them anything yet :)

branquinhoaa: aaaah, ok! now I get it.:sweat_smile: 
It is to remove it then.

mkalininait:
@branquinhoaa 
1. I think it makes sense to make it in this PR. What do you think @klappradla ?
2. Replied here :) https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/789#discussion_r128511565 But we will need the confirmation/rejection feature later as a part of #772

app/controllers/attendances_controller.rb
klappradla: Ok, thanx for the explanation. So let's just go with your initial approach for now (the one without explicitly raising an exception), since you're right, this controller will very likely change to be orga only in #722 and then we can probably get rid of `student_team` all together.

So let's not waste any more thoughts on this, just go back to the initial version ✌️

app/models/conference_preference.rb
```diff
@@ -0,0 +1,6 @@
+class ConferencePreference < ActiveRecord::Base
+  include GithubHandle
```
klappradla: This module only works if a `user` method is available in the class. With switching the `Attendance`/ `ConferencePreference` to be owned by the team, it will no longer work.

db/seeds.rb
```diff
@@ -38,6 +38,7 @@
   random_date = rand(1.year).seconds.from_now
   FactoryGirl.create(:conference, :in_current_season,
     location: FFaker::Venue.name,
+    region: ["Africa", "South America", "North America", "Europe", "Asia Pacific"].sample,
```
klappradla: Interesting, I did not know `FFaker` does not have continents _(maybe submit a PR and have another easy to grab contribution?)_

spec/models/conference_spec.rb
```diff
@@ -50,12 +50,12 @@
   describe '#tickets_left' do
 
     context 'ticket value defined' do
-      let(:attendance)    { FactoryGirl.build(:attendance, confirmed: true) }
+      let(:conference_preference)    { FactoryGirl.build(:conference_preference, confirmed: true) }
```
klappradla: Spacegame 🚀 🎮

mkalininait:
Tested. Works AWESOME!! One tiny tiny thing: the secondary choice goes before the primary choice
screen shot 2017-07-21 at 09 43 49
Maybe we can also label conferences (primary, secondary) in the team views, to avoid confusion.
screen shot 2017-07-21 at 09 50 50

For future PR related to Issue #771:
We now have the following headers in the team views:
...
h3 Supervisors
h4 Conference preferences
h3 Activities

The Conference preferences in one level lower, but if we later add h3 Conferences and h4 Conference offers (or something like that), it all will be very consistent :)

app/controllers/attendances_controller.rb
branquinhoaa: In fact, I removed this controller. Now that we don't need to confirm preference anymore, this controller makes no sense.

db/seeds.rb
branquinhoaa: Seems a nice idea! :+1:

branquinhoaa:
labels
Hi @mkalininait !!
I have changed the things you asked for. Hope it works perfectly now!! :dancing_men:

app/controllers/attendances_controller.rb
klappradla: Also fine 👍

app/controllers/orga/conferences_controller.rb
```diff
@@ -33,7 +33,7 @@ def conference_params
       :round, :lightningtalkslots, :tickets, :flights, :accomodation,
       :gid, #id in orga's Google Spreadsheet (format: 2017001)
       :notes,
-      attendances_attributes: [:id, :github_handle, :_destroy]
+      conference_preferences_attributes: [:id, :github_handle, :_destroy]
```
klappradla: Having `ConferencePreferences` associated with the team rather then the user, we can also get rid of the `github_handle` here (just a tiny cleanup thing) :v:

mkalininait:
Hey @branquinhoaa We are very very close to perfection! 
The order of conferences is correct on the edit page now, but still wrong (or sometimes wrong?) in the team views (orga and non-orga).

screen shot 2017-07-24 at 14 19 00

branquinhoaa:
@mkalininait :D done!

mkalininait:
@branquinhoaa Yay! Thank you! <3

Merge! Merge! Merge! @carpodaster Any objections?

PR #791, opened 2017-07-14
Improve projects page (step 1)

F3PiX:
Solves the first part of #784 (Projects Page)
This should provide a robust query for the Projects Page.

doublecheck: 
The ‘new’ projects page uses a filter to only show the projects that students are working on. This filter works during the ‘active’ season, i.e. from the day the acceptance letters are sent, to the end of the coding summer (≈ May through Sept).
The rest of the year the page shows the regular selection, as it was before, i.e. proposed and accepted projects.
- [ ] Is this the behaviour we want? 

ToDO:
- [ ] Remove project_id from applications table => move to separate issue

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -37,8 +37,8 @@ class Project < ActiveRecord::Base
   end
 
   def self.selected
-    project_names = Team.in_current_season.accepted.pluck(:project_name).uniq
-    Project.in_current_season.where(name: project_names)
+    project_names = Team.in_current_season.accepted.pluck(:project_name)
+    Project.in_current_season.accepted.where(name: project_names)
   end
 
   def taglist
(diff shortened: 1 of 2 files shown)


F3PiX:
Copied here for convenience: Notes by Max from the already merged PR

https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/786#pullrequestreview-49250816
> Let's skip the uniq here - it will convert the whole thing to an Array (and actually execute the query), which is not really necessary (there can be loads of duplicates, the final resulting Projects will be the same ✌️)

Done ✅ 

https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/786#discussion_r126738790
> I'm not 100% sure if it's necessary, but I'd have scoped it even more to this:

> `Project.in_current_season.accepted.where(name: project_names)`
> This makes sure, we really only look into the accepted projects (maybe there's a rejected one with the same name...).

Done ✅

F3PiX:
@klappradla I'd rather do the removal of the project_id in a separate issue. It is indeed a small migration, but it generates totally unrelated changes. Okay?

PR #793, opened 2017-07-17
Move mailing stuff into orgs namespace

### LINKED ISSUE #757: Move mailer into orga namespace
Orga users can send email to targeted groups via `.../mailings/new` . The use of this feature is suboptimal. Especially because the mailer is not suited (yet) to play a part in more complicate workflows.

As only orga users use the email functionality, it should live in the orga namespace.

F3PiX:
Closes #757 

Everything for sending, reviewing and resending mails to targeted groups is now in the orga namespace, and accessible from the orga dashboard 😋 

It looks like there are opportunities for follow-ups with more improvements. I restrained myself to do the moving part and relevant changes, and nothing else. Much.  
The submissions specs were not totally clear; the things I wanted to test are already in the mailer specs. While the mailer is fired after the `submission`. I left it as is. Mostly.

Some questions in the comments


app/models/ability.rb
```diff
@@ -49,11 +49,11 @@ def initialize(user)
       user.admin? || user == attendance.user
     end
 
+    #todo add mailing controller and view for users in their namespace, where applicable
```
F3PiX: I'd like to move the user's own mailings in their namespace/dashboard.
Also, I'll have to check if mailings will pop up in the activity page; and if we want them overthere in the first place. To be continued.

spec/controllers/mailings_controller_spec.rb
```diff
@@ -6,17 +6,14 @@
...
 
-  shared_context 'User logged in' do
-    before { sign_in user }
-  end
-
   describe 'GET index' do
     before { mailing }
 
     context 'with user logged in' do
-      include_context 'User logged in'
```
F3PiX: Why is the shared_context not working here? It does work on other places.

spec/models/mailing_spec.rb
```diff
@@ -1,26 +1,40 @@
...
       role = create :coach_role
       mailing.save!
       expect{mailing.submit}.to change{Submission.count}.by(mailing.emails.count)
     end
   end
 
-  describe '#recipients' do
-    skip 'TODO'
-  end
+  # Remove? This is tested in recipients_spec.rb
```
F3PiX: ??

spec/models/submission_spec.rb
```diff
@@ -12,4 +14,10 @@
       expect(Submission.unsent).not_to eql nil
     end
   end
-end
\ No newline at end of file
+
+  describe '#enqueue' do
+    it 'sent into delivery' do
+
+    end
```
F3PiX: I didn't know what to do with the existing specs. AFAIK we write specs for the public methods, and both 'valid' and 'sent' aren't methods in `submission.rb`. 
But what should be tested in the methods that _are_ present? 🤔

spec/routing/orga/submissions_routing_spec.rb
```diff
@@ -0,0 +1,33 @@
+require 'spec_helper'
+
```
F3PiX: Delete these as well?

app/controllers/orga/mailings_controller.rb
```diff
@@ -0,0 +1,67 @@
...
+
+  def destroy
+    @mailing.destroy!
+    redirect_to orga_mailings_path, notice: "The mailing has been deleted"
+  end
+
+  private
+
+  def find_mailing
+    @mailing ||= Mailing.find(params[:id])
```
klappradla: Why is this `||=` rather then just `=`? Maybe I'm overseeing it, but I don't find a scenario where this would be called twice within one request response cycle...?

app/controllers/orga/submissions_controller.rb
```diff
@@ -0,0 +1,29 @@
+class Orga::SubmissionsController < Orga::BaseController
+  before_action :find_mailing
+  before_action :set_submission, only: :update
```
klappradla: Again, if this is only used for **one specific** action, I'd suggest to instead make it an explicit call within the specific action. Especially given the fact that there is already another `before_action` for the respective action and they now depend on their order...

app/views/mailings/index.html.slim
```diff
@@ -1,24 +1 @@
...
-        th Subject
-        th Seasons
-        th Group
-        th To
-        th From
-    tbody
-      = render @mailings
-
-  = paginate @mailings, theme: 'twitter-bootstrap-3', pagination_class: 'pagination-sm'
+= render 'mailings/table', namespace: nil, mailings: @mailings
```
klappradla: Is the `table` partial supposed to be reused somewhere else or is this `index` view becoming more complex in the near future?  If not, I think it would make sense to the code from the partial into the actual view. Rendering things and looking up partials is probably one of the slowest parts of the framework, I don't see a point it utilizing it for no reason.

It also doesn't make things easier to grasp here in my opinion - correct me if I'm wrong.

app/views/mailings/_details.html.slim
```diff
@@ -0,0 +1,19 @@
+h1 class="mailing" = @mailing.subject
```
klappradla: I haven't been involved in Rails view code in some time, so please correct me if I'm wrong, but aren't partials supposed to work on "normal" variables rather then view-instance variables? As in one would pass `mailing: @mailing` and then have `mailing` available in the partial.

app/views/mailings/_table.html.slim
```diff
@@ -0,0 +1,18 @@
+h1 Mailings
+
+- if @mailings.any?
```
klappradla: Same as for the `_details` partial: how do you have `@mailing` available here?

app/views/mailings/show.html.slim
```diff
@@ -1,34 +1 @@
...
-  dt Sent at
-  dd = @mailing.sent_at ? @mailing.sent_at.to_formatted_s(:short) : '-'
-
-.mailing
-  .body
-    = render_markdown(@mailing.body).html_safe
-
-- if can?(:read, Submission)
-  = render partial: 'submissions', locals: { mailing: @mailing, submissions: @mailing.submissions }
+= render 'mailings/details', namespace: nil, mailings: @mailings
```
klappradla: Same as for the `index` view: if this partial ain't going to be reused for another mailer view, I'd suggest to simply put the view code in the `show` view.

config/routes.rb
```diff
@@ -63,16 +63,13 @@
   resources :roles, only: [] do
     put :confirm, on: :member
   end
+  resources :mailings
```
klappradla: As I wrote for the `MailingsController`: if you we should not have any more actions then `show` and `index` for this resource, this should probably be done on router level:
```ruby
resources :mailings, only: [:index, :show]
```

spec/controllers/mailings_controller_spec.rb
klappradla: What do you mean with *not working*? The way I understand the setup, it would try to get the mailing before it would sign in the user - order of `before` hooks... is that how it's supposed to be set up?

spec/controllers/mailings_controller_spec.rb
```diff
@@ -6,17 +6,14 @@
...
+      # include_context 'with user logged in' # => red
+      before { sign_in user }                 # => green
```
klappradla: If you want to have the call to `index` dryed out of the examples, you could also do
```ruby
before do
  sign_in user
  get :index
end
```

spec/controllers/mailings_controller_spec.rb
```diff
@@ -6,17 +6,14 @@
   let(:mailing)        { FactoryGirl.create(:mailing) }
   let(:user)           { FactoryGirl.create(:user) }
 
-  shared_context 'User logged in' do
-    before { sign_in user }
-  end
-
   describe 'GET index' do
     before { mailing }
```
klappradla: I'm afraid I don't understand this pattern. What's the difference to making line 6 a `let!(...)`?

spec/controllers/orga/mailings_controller_spec.rb
```diff
@@ -0,0 +1,37 @@
+require 'spec_helper'
+
+describe Orga::MailingsController do
+  render_views
+
+  it_behaves_like 'redirects for non-admins'
+
+  let(:mailing)        { FactoryGirl.create(:mailing) }
```
klappradla: Spacegame 🚀

spec/models/mailing_spec.rb
```diff
@@ -1,26 +1,40 @@
...
-  let(:mailing) { Mailing.new(from: 'from@email.com', to: Role::ROLES,
+  let!(:mailing) { Mailing.new(from: 'from@email.com', to: Role::ROLES,
     cc: 'cc@email.com', bcc: 'bcc@email.com', subject: 'subject', body: '# body') }
 
   it { is_expected.to have_many(:submissions).dependent(:destroy) }
 
   describe '#sent?' do
-    skip 'TODO'
+
+    it 'false before submission' do
```
klappradla: For the sake of readability, I'd suggest to name this *it 'returns false before submission' do*

spec/models/mailing_spec.rb
```diff
@@ -1,26 +1,40 @@
...
+      expect(mailing.subject).to eq 'subject'
+      mailing.save!
+      expect(subject.sent?).to eq false
+    end
+
+    it 'true after submitting' do
```
klappradla: For the sake of readability, I'd suggest to name this *it 'returns true after submitting' do*

spec/models/mailing_spec.rb
```diff
@@ -1,26 +1,40 @@
...
     cc: 'cc@email.com', bcc: 'bcc@email.com', subject: 'subject', body: '# body') }
 
   it { is_expected.to have_many(:submissions).dependent(:destroy) }
 
   describe '#sent?' do
-    skip 'TODO'
+
+    it 'false before submission' do
+      expect(mailing.subject).to eq 'subject'
+      mailing.save!
```
klappradla: What's the sake of calling `save` here once again after using a `let!` in the beginning? Does it do anything new?

spec/models/mailing_spec.rb
```diff
@@ -1,26 +1,40 @@
...
+      expect(subject.sent?).to eq false
+    end
+
+    it 'true after submitting' do
+      mailing.save!
```
klappradla: Same as for the negative case: what does `save` here do?

spec/models/mailing_spec.rb
```diff
@@ -1,26 +1,40 @@
...
+      mailing.submit
+      expect(mailing.sent_at).not_to be_nil
+      expect(mailing.submissions).not_to be_nil
+      expect(mailing.sent?).to eq true
+    end
   end
 
   describe '#submit' do
-    it 'should deliver seperate emails to all recipients (incl. cc and bcc)' do
+    it 'should deliver separate emails to all recipients (incl. cc and bcc)' do
```
klappradla: I don't see the description reflected in the expectation here...

spec/models/mailing_spec.rb
```diff
@@ -1,26 +1,40 @@
...
       role = create :coach_role
       mailing.save!
       expect{mailing.submit}.to change{Submission.count}.by(mailing.emails.count)
```
klappradla: I don't understand what the `by(mailings.emails.count)` part is for... Why don't we use a fixed number here? I personally would have no idea what number and context this is talking about.

spec/routing/orga/submissions_routing_spec.rb
klappradla: Why deleting them?

spec/models/submission_spec.rb
klappradla: The way I understand this, this is supposed to test the good path of submitting a mailing -> it will be send if validations don't fail. But I agree with you, this could be expressed somehow more obvious. A more explicit description could already be a start 😉

spec/controllers/mailings_controller_spec.rb
carpodaster: It should be a `let!`, yes. But it's been there before so technically not part of this PR ;)

F3PiX:
Thanks Max for your review! Much appreciated.

Three issues:
1) partials (https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/793#discussion_r127827825)
The `_table` and `_details` partials are shared partials, by the `orga/mailings_controller` and the public `mailings_controller`. I'd prefer to leave them as partials, and not duplicate the code. Okay? 

2) tests
I still don't understand all of the code. I added some basic specs, and I left comments inside the code 😱 , to explain my problems (to prevent GH noise with explanations). 
The main logic is from before ActiveJob, and [at one point](https://github.com/rails-girls-summer-of-code/rgsoc-teams/commit/71cb736d9f84dfd967ce099821d48bac62916590#diff-8b04dfc748ae7c73f82a96d00a3db3c2), a `Worker` was replaced with a Mailer.
Either there is some duplication, or I don't fully understand it. In both cases, it doesn't help to find the right specs to write.    
Of course, I will remove the comments in the next commit. 
I'd like to know if I am going in the right direction with the specs.

3) About the redirection and the routing([your comment])(https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/793#discussion_r127829287) : Instead of the complicated and confusing redirection, I just restored cancancan's `load_and_authorize_resource`. In order to get that to show a 'non authorised' notice (instead of `action not found`), the routes has to be present in the `routes` without constraints, and empty declarations in the controller. 
(For the empty declarations: I tend to add only new and edit.) 

  
@klappradla Before doing a code review, maybe first answer to this issues? Then I'll update accordingly.

klappradla:
Thanx for the tl;dr @F3PiX 🙏 

1. Sure, stay with partials here ✔️ 
1. Yep, tests go into the right direction ✔️  _(I cannot really answer the shared_context thing so, would have to try it out locally)_
1. 👍  for `load_and_authorize_resource` ✔️

spec/support/shared_contexts/with_user_logged_in.rb
```diff
@@ -1,5 +1,5 @@
 RSpec.shared_context 'with user logged in' do
-  let(:current_user) { build(:user) }
+  let(:current_user) { create(:user) }
```
F3PiX: LOL Turned out: Just a few months ago, you updated from `build_stubbed` to `build`.  🤣

F3PiX:
@klappradla Back to you! 🍪 
I think the specs are covering the basic functionality now. If this functionality gets more use during the summer, it may be worthwhile to look into improving and optimising the code.

app/controllers/orga/mailings_controller.rb
```diff
@@ -0,0 +1,63 @@
...
+  before_action :normalize_params, only: [:create, :update]
+  before_action :find_mailing, only: [:show, :edit, :update, :destroy]
+
+  def index
+    authorize! :read, :mailing
+    @mailings = Mailing.order('id DESC').page(params[:page])
+  end
+
+  def show
+    authorize! :read, :mailing
```
klappradla: Has eventually the same outcome, but for making it more future-proof: this should probably be
```ruby
authorize!, :read, @mailin
```

app/controllers/orga/mailings_controller.rb
```diff
@@ -0,0 +1,63 @@
...
+  def create
+    @mailing = Mailing.new(mailing_params)
+    if @mailing.save
+      redirect_to orga_mailing_path(@mailing), notice: 'Mailing was successfully created.'
+    else
+      render action: :new
+    end
+  end
+
+  def update
```
klappradla: Why no `authorize!` for the update, edit and destroy actions?

Is it enough to know the user is a member of `orga`? If so, one could savely remove the `authorize!` calls for show and index as well.

spec/support/shared_contexts/with_user_logged_in.rb
klappradla: Good catch!

F3PiX:
@klappradla Can you take one more look please? Apart from deleting the redundant `authorize` calls, I did two more commits. Because somewhere in the PR I introduced a `DelegationError`, that I didn't notice before 🙈 . It is solved with setting a default for `from`.

app/controllers/mailings_controller.rb
```diff
@@ -4,7 +4,7 @@ class MailingsController < ApplicationController
 
   def index
     @mailings = Mailing.order('id DESC').page(params[:page])
-    authorize! :read, @mailing
+    authorize! :read, :mailing
```
klappradla: 👍

app/controllers/mailings_controller.rb
```diff
@@ -1,59 +1,17 @@
...
-  end
-
   def index
+    @mailings = Mailing.order('id DESC').page(params[:page])
     authorize! :read, :mailing
   end
 
   def show
-    authorize! :read, :mailing
+    @mailing = Mailing.find(params[:id])
```
klappradla: Same as for `edit`, `load_and_authorize_resource` does already do this for you - you have the `@mailing` instance variable already available via some before hooks it automatically adds to the controller. So you should skip that line in order to spare the extra call to the database ✌️

F3PiX: Thanks for catching this! It makes so much sense, given that they called the method LOAD_and_.. ;-)

F3PiX:
@klappradla \o/

app/controllers/mailings_controller.rb
```diff
@@ -1,17 +1,15 @@
...
   def index
     @mailings = Mailing.order('id DESC').page(params[:page])
     authorize! :read, :mailing
   end
 
-  def show
-    @mailing = Mailing.find(params[:id])
-  end
-
+  # These actions are here to enable the cancancan 'not authorised' notice
```
klappradla: Thanx for the comment 🙏 
Will for sure help people looking at this in the future 🖤

Issue #792, opened 2017-07-14
In conferences table: handle empty url's in views


F3PiX:
Conference are imported by csv. If the url field in the csv file is empty, the teams app shows a useless url, like `orga/conferences`. 

Find out how the url is set in the index and show views, and show something usefull (maybe an empty field?).
Make sure that `sanitizing` imported url's stays in place, if needed.

branquinhoaa:
:+1:  what about the other empty fields? that will be allowed?

PR #795, opened 2017-07-17
show url just when the field is not empty

branquinhoaa:
PR related with the issue #792

conferences index - before change:
conf_index_before

conferences index - after change:
index_after

conference show - before change:
conference_show_before

conference show - after change:
show_after

- verify if URL is present in a conference before display it.

## update in the view (let the label even if the field is empty):

after_change_including_title

### DIFF
--- a/app/views/conferences/_details.html.slim
+++ b/app/views/conferences/_details.html.slim
@@ -6,8 +6,8 @@
     dd = @conference.date_range
     dt Twitter
     dd = format_conference_twitter(@conference.twitter)
-    -if @conference.url
-      dt Homepage
+    dt Homepage
+    -if @conference.url.present?
       dd = link_to @conference.url, @conference.url
     dt Notes
     dd = @conference.notes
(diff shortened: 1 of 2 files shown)


app/views/conferences/_details.html.slim
```diff
@@ -6,7 +6,8 @@
     dd = @conference.date_range
     dt Twitter
     dd = format_conference_twitter(@conference.twitter)
-    dt Homepage
-    dd = link_to @conference.url, @conference.url
+    -if @conference.url
```
carpodaster: Can you add a `present?` please? When we later edit the conference and leave the url field empty, it will end up with an empty string (which is _truthy_).

app/views/conferences/_table.html.erb
```diff
@@ -20,7 +20,7 @@
         <td><%= conference.city %></td>
         <td><%= conference.country %></td>
         <td><%= conference.region %></td>
-        <td><%= link_to conference.url, conference.url %></td>
+        <td><%= link_to conference.url, conference.url if conference.url %></td>
```
carpodaster: Add `present?` – see above.

app/views/conferences/_details.html.slim
F3PiX: Cool! Thanks Team! ✨ 
I'd vote for leaving the label in, even when the url value is empty. Same as we do with the other labels.

Issue #794, opened 2017-07-17
Github Handle saving empty fields


juuh42dias:
GithubHandle returning false or nil class when edit/update your team.

![erro](https://user-images.githubusercontent.com/1828204/28287152-febb46ae-6b10-11e7-9f35-015d10026a2e.png)

carpodaster:
Hey @juuh42dias – good call. I've seen this error around but never really looked into it.

The problem is line 9: it does not return / assign the `user` instance variable. This causes the error in the following line (it tries to call a method on `false`).

Can you transform line 9 to something that safely returns a user instance and/or guards against illegal values for `github_handle` alltogether?

juuh42dias:
I believe that method first_or_initialize (return find or new) should return an instance to save(blank or not). I think github_handle.present? maybe don't necessary. What do you think?

carpodaster:
I'm closing this one, I'm positive it has been fixed with the referenced PR

PR #796, opened 2017-07-17
Delete github_handle when empty field

juuh42dias:
Related #794


carpodaster:
Hello @juuh42dias! I don't think this PR will solve the underlying problem. See my comment at issue #794.

carpodaster:
Hey @juuh42dias that looks very good now. Do you think you could add a test for that?

app/controllers/teams_controller.rb
```diff
@@ -74,6 +74,7 @@ def set_users
 
     def team_params
       params[:team].fetch(:sources_attributes, {}).delete_if { |key, source| source[:url].empty? }
+      params[:team].fetch(:roles_attributes, {}).delete_if { |key, role| role.present? && role[:github_handle].empty? }
```
carpodaster: I'm not really sure what this line and the one above do. That is, I don't know if `#fetch` returns a _copy_ of the nested hash or the actual instance. Depending on what object it returns, deleting keys from it probably doesn't do anything.

In any event, I would remove the line again. The problem (or well: at least _one_ of the problems 😉 ) should be solved by your change in the github_handle concern.

carpodaster:
@juuh42dias are you working on a test? I'm fine with merging without one but if you'd like to go for it that's of course highly appreciated. But we can also add one later.

juuh42dias:
I'll create tests for this feature, is better.

klappradla:
Since this is now lying around for quite a while, what's the state of this @juuh42dias ?

Is the feedback @carpodaster gave implemented / discussed? Are there any open questions or any blockers? Or - also possible of course 😉  - is this no longer relevant?

carpodaster:
@juuh42dias Are you ok if we merge this? I can add the test later as a PR and ping you for a review / walkthrough.

I would remove the last two commits though, the shared example doesn't test anything right now.

Issue #788, opened 2017-07-12
Add T-Shirt cut to User profile


mkalininait:
Next to "T-Shirt size" we need another dropdown "T-Shirt cut" with possible values:
 - [blank]
 - Straight cut
 - Fitted cut

mkalininait:
@alicetragedy Are you ok with it?

alicetragedy:
👍

carpodaster:
Closes in #797

mkalininait:
Tested on prod, works well. Thank you @branquinhoaa and @carpodaster !

PR #797, opened 2017-07-18
User/tshirt cut

branquinhoaa:
PR related to the issue  #788

- include column tshirt_cut in the user model
- include a select dropdown to choose the t-shirt cut
- show the selected t-shirt cut where t-shirt size is shown

### orga user info
orga_user_info_before
### show user
show_user_before
### edit user
tshirt_before

### orga user info
orga_user_info_after

### show user
show_user_after

### edit user
tshirt_after

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -115,7 +115,7 @@ class UsersController < ApplicationController
       params.require(:user).permit(
         :github_handle, :twitter_handle, :irc_handle,
         :name, :email, :homepage, :location, :bio,
-        :tshirt_size, :postal_address, :timezone,
+        :tshirt_size, :tshirt_cut, :postal_address, :timezone,
         :country,
         :hide_email,
         :is_company, :company_name, :company_info,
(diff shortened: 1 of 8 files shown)


mkalininait:
@carpodaster Yes, I think we should allow the blank option.

mkalininait:
@branquinhoaa Could you please add the new field to `app/exporters/users.rb`? Thanks!

mkalininait:
@branquinhoaa The exporter fix looks good to me!
The 'allow blank value' change I didn't test. But I trust your testing.
Interesting note: the 'blank' value of 'tshirt_size' is set to false, but I still can set the blank value:
`= f.input :tshirt_size, as: :select, collection: User::TSHIRT_SIZES.map { |k| [k, k] }, label: 'T-Shirt size', blank: false, required: false` I like the behaviour, I just wonder why it works this way...

Anyway, I think we can merge this PR, unless @carpodaster has a different opinion :)

carpodaster:
> […] unless @carpodaster has a different opinion :)

He doesn't ^^

LGTM!

PR #798, opened 2017-07-19
Change import format from .csv to .tsv

F3PiX:
Maria reported problems with uploading a .csv file in production., from a Google Spreadsheet export.
Google doesn't allow to override the column separator, other than choosing 'csv' or 'tsv'.
After changing the `col_sep` option to `','` , for Maria the date format with comma's inside ('Aug 7, 2017') did not cause problems, but on my computer, the comma inside was read as an extra separator. 

The safest solution is to use the tab separated values content type.

@carpodaster Do you agree?
(And look how easy the `content_type` changes were!  🙏  )

### DIFF
--- a/spec/fixtures/files/test.tsv
+++ b/spec/fixtures/files/test.tsv
@@ -1,7 +1,7 @@
-UID/tName/tStart date/tEnd date/tCity/tCountry/tRegion/tWebsite/tNotes
-2017001/tDeccan RubyConf 2017/tAug 12, 17/tAug 12, 17/tGent/tBelgium/tEurope/thttp://www.deccanrubyconf.org//t
-2017002/tJSFoo 2017/tSep 15, 17/tSep 16, 17/tBangalore/tIndia/tAsia Pacific/thttps://jsfoo.in/2017//t
-2017003/t/t/t/tDelhi/tIndia/tAsia Pacific/thttps://in.pycon.org//t1st week of Nov (dates are to be defined)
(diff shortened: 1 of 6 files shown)


spec/fixtures/files/test.tsv
```diff
@@ -0,0 +1,7 @@
+UID/tName/tStart date/tEnd date/tCity/tCountry/tRegion/tWebsite/tNotes
```
carpodaster: Is this what an actual Google export looks like? The  `/t` isn't a tab escape. `\t` would be. That's why I am curious.

F3PiX: 🙈

F3PiX:
I didn't find that we can tell Google to do that, but I am not very into Google Docs. On Maria's computer it does surround with quotes, on mine it doesn't. 
Either way, it would require more 'instructions' on what the imported file should look like. And that I don't like much.

F3PiX:
I searched about the Google settings. Google doesn't allow to set export settings, but it does allow import settings. So, that requires an extra import action into a new file, and than we get:
```
UID,Name,Start date,End date,City,Country,Region,Website,Notes
2017001,Deccan RubyConf 2017,"Aug 12, 2017","Aug 12, 2017",Gent,Belgium,Europe,http://www.deccanrubyconf.org/,
```
I think that is a bit unpractical and error prone. 
Are there objections against .tsv ?
@carpodaster

F3PiX:
Closing; see #800

PR #800, opened 2017-07-22
Change delimiter to comma

F3PiX:
Instead of #798 
Using the comma as a delimiter instead of semicolon.

### DIFF
--- a/spec/fixtures/files/test.csv
+++ b/spec/fixtures/files/test.csv
@@ -1,7 +1,7 @@
-UID;Name;Start date;End date;City;Country;Region;Website;Notes
-2017001;Deccan RubyConf 2017;Aug 12, 17;Aug 12, 17;Gent;Belgium;Europe;http://www.deccanrubyconf.org/;
-2017002;JSFoo 2017;Sep 15, 17;Sep 16, 17;Bangalore;India;Asia Pacific;https://jsfoo.in/2017/;
-2017003;;;;Delhi;India;Asia Pacific;https://in.pycon.org/;1st week of Nov (dates are to be defined)
-2017004;GHC India 2017;17/11/2017;15/11/2017;Bangalore;India;Asia Pacific;https://ghcindia.anitaborg.org/;They have student scholarship applications. Deadline: June 30, 2017
-2018005;PyCon Pune 2018;;;Pune;India;Asia Pacific;https://pune.pycon.org/;Dates are to be defined
-2017006;CheesyRuby;07/07/2017;15/07/2017;Amsterdam;NL;Europe;www.example.com;Duly Noted
+UID,Name,Start date,End date,City,Country,Region,Website,Notes
+2017001,Deccan RubyConf 2017,"Aug 12, 17","Aug 12, 17",Gent,Belgium,Europe,http://www.deccanrubyconf.org/,
+2017002,JSFoo 2017,"Sep 15, 17","Sep 16, 17",Bangalore,India,Asia Pacific,https://jsfoo.in/2017/,
+2017003,,,,Delhi,India,Asia Pacific,https://in.pycon.org/,1st week of Nov (dates are to be defined)
+2017004,GHC India 2017,"17/11/2017","15/11/2017",Bangalore,India,Asia Pacific,https://ghcindia.anitaborg.org/,They have student scholarship applications. Deadline: June 30, 2017
+2018005,PyCon Pune 2018,,,Pune,India,Asia Pacific,https://pune.pycon.org/,Dates are to be defined
+2017006,CheesyRuby,"07/07/2017","15/07/2017",Amsterdam,NL,Europe,www.example.com,Duly Noted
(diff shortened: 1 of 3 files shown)


spec/fixtures/files/test.csv
```diff
@@ -1,7 +1,7 @@
...
-2017002;JSFoo 2017;Sep 15, 17;Sep 16, 17;Bangalore;India;Asia Pacific;https://jsfoo.in/2017/;
-2017003;;;;Delhi;India;Asia Pacific;https://in.pycon.org/;1st week of Nov (dates are to be defined)
-2017004;GHC India 2017;17/11/2017;15/11/2017;Bangalore;India;Asia Pacific;https://ghcindia.anitaborg.org/;They have student scholarship applications. Deadline: June 30, 2017
-2018005;PyCon Pune 2018;;;Pune;India;Asia Pacific;https://pune.pycon.org/;Dates are to be defined
-2017006;CheesyRuby;07/07/2017;15/07/2017;Amsterdam;NL;Europe;www.example.com;Duly Noted
+UID,Name,Start date,End date,City,Country,Region,Website,Notes
+2017001,Deccan RubyConf 2017,"Aug 12, 17","Aug 12, 17",Gent,Belgium,Europe,http://www.deccanrubyconf.org/,
+2017002,JSFoo 2017,"Sep 15, 17","Sep 16, 17",Bangalore,India,Asia Pacific,https://jsfoo.in/2017/,
+2017003,,,,Delhi,India,Asia Pacific,https://in.pycon.org/,1st week of Nov (dates are to be defined)
+2017004,GHC India 2017,"17/11/2017","15/11/2017",Bangalore,India,Asia Pacific,https://ghcindia.anitaborg.org/,They have student scholarship applications. Deadline: June 30, 2017
```
carpodaster: Is this a file exported by Google? 'Cause it doesn't quote the commas in fields.

F3PiX: It is not exported by Google directly.   
Google does not quote comma's within fields; it only uses quotes around field values, depending on the settings of the file, the columns (and the computer) you are exporting.

carpodaster: Let us use a real file exported by Google please. As you stated before, CSV is a very loose format. We can't accomodate any CSV format under the sky. Instead we should aim for what Google exports right now (since one cannot customise the output format) and since Google is what Maria is using right now.

carpodaster:
@F3PiX @mkalininait I am beyond confused now. I interpreted https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/762#issuecomment-316683360 as "Google exports deterministically. With proper quotes so that commas in fields don't conflict with our separator. Is that assumption true? If yes: Maud, can you add a fixture file that reflects the output format Google Docs is using (see my other comment above)? Thanks!

F3PiX:
To the best of my knowledge, that assumption is not true.

mkalininait:
@carpodaster I still think that Google does export deterministically, with `,` as a separator (csv stands for 'comma-separated values' after all) and `"` wrapping of values which contain commas.

@F3PiX Could you please clarify how you see it? Do you mean that Google Sheets exports differently depending on some circumstances?

F3PiX:
@carpodaster @mkalininait 

I am sorry about the confusion. 🙈  I'll be more specific about what happens with export. And why we get different exports.  

1) For columns that are formatted `automatic` or `plain text`, it will surround the value with "" IF there is a comma in the value, and not if there is no comma in the value. 
 
2) For columns that are formatted as dates, it is the same: it will surround the value with "" if the chosen date format contains a comma. 

The format in the/Maria's Google conferences Spreadsheet has either a text format, or a different (non-Dutch) date format. With my Google Account settings (Dutch), that exact date format (`Aug 19, 2017`) is not available.     

So, the exported format is depending on:
* The format settings of a column in the Google Spreadsheet  
* A possible different format setting within that column - because 
* The Google Account language settings 
* and maybe other settings.

AFAIK (and could find) this is not very well documented, so it was trial and error. 
(Heads up: This was the point in my research that I concluded that the .tsv format delivers the more consistent output; either auto, text or date format, and even with different date formats within one column, the output is always and for everyone the same.)

After we went back to .csv, I updated the fixture file so that it reflects possible states of the dates values in the to-be-uploaded-conferences-csv file, depending on different settings. 
That is why I did not use a real export: first, I don't get the exact same export as Maria gets; second, I was thinking the test file should provide data to test processing the data, not to test the data format itself.
If nothing else, I am now less worried that the .csv format will cause troubles.   

But, now I am not sure if I should or shouldn't upload a new and real export file?
(Apart from that, I realise that I forgot to surround the notes field in the invalid `17004`; I will correct that and move the note to a valid input.)      

@carpodaster Can you please confirm once more if we go with a real export or a hand-made (and improved) 'mirror'? Because I am confused too :-)

carpodaster:
I would definitely go for the real thing as our fixture test file.

As for the comma values in dates: since everything in a CSV file is a string anyway, it should only be important that it's a format that Date.parse would understand.

F3PiX:
@carpodaster Back to you!

mkalininait:
Yay! Thanks @F3PiX @carpodaster!

> the url field sometimes contain more than one URL

This won't be the case in the real conf table :)

Issue #766, opened 2017-06-24
[Conferences] #3: Adding a new conference


mkalininait:
The idea: https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/751#issuecomment-308574743
Full list of conference user stories: https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/751#issuecomment-309138513

**User Story:**
As a student, I can add a cool conference which is missing in the longlist; I can choose it as my team's primary or secondary conference then.

**Things to do:**
On the Team's page, in the Conference preferences block (see Issue #767) add a new button: "I want to suggest another conference!", which opens the form "New conference" (this form existed at one point but I'm not sure we can re-use it). The form should contain the same fields which we import with the Importer (see Issue #762), except "id".

branquinhoaa:
@mkalininait we would like to do this, but I am almost sure that the previous issue has to be merged before I start this one. Am I right?

branquinhoaa:
@klappradla Could you take a look too? we have changed the things you asked for

mkalininait:
@branquinhoaa Hm.. Do you mean this PR? https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/789 This task looks independent to me. I *think* you can start working on it.

carpodaster:
@mkalininait is this solved by #802 ?

mkalininait:
@carpodaster jup!

Issue #790, opened 2017-07-13
"My team" button for users with multiple teams


mkalininait:
This task derives from this issue and comment: https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/621#issuecomment-276913772

There is the button "My team" on the Teams page. It was supposed to open user's current team's page. The problem is that some users have multiple current teams, e.g. coaches. 

Possible solutions:
1) "Can we just remove it [the button] or add a "My Team(s)" item to the nav bar that leads to a filtered teams#index with only those teams I am part of?"
2) "Maybe we can even add this filter to the teams page explicitly (instead of the button):
* [ ] My Team(s) "

branquinhoaa:
What if we change from a button to a dropdown button with all the teams that you have? when you choose the team in the list we redirect for that team page.

branquinhoaa:
@mkalininait What if we change from a button to a dropdown button with all the teams that you have? when a you choose the team in the list we redirect for that team page.

mkalininait:
@branquinhoaa This is an option, but I think it may be a bit weird to have two lists (and two ways of opening a team page) on one page...  I'm not a UX expert though.

troubalex:
I don't think I have access to the team app any longer, so I can't check what the current navigation looks like. A top-level navigation item sounds like a smart, pretty straight forward solution though. It's discoverable, gives people a feeling of location within the app, and probably yields the least surprises. So I'd go with that.

But then again, I am shooting from the hip here. A screenshot of the current navigation would be useful. 😇

alicetragedy:
A top-level navigation item makes the most sense to me as well. 
@troubalex I've pasted some screenshots of the current navigation and its sub-level items :)
<img width="414" alt="screen shot 2017-07-27 at 15 02 12" src="https://user-images.githubusercontent.com/3143348/28671344-9e86e7ce-72dc-11e7-9280-0909307f5d11.png">
<img width="384" alt="screen shot 2017-07-27 at 15 02 22" src="https://user-images.githubusercontent.com/3143348/28671346-a09c6124-72dc-11e7-9151-51e62721e198.png">

PR #802, opened 2017-07-25
Conferences/add new conference

branquinhoaa:
PR related with the issue #766 - Allow student to add a new conference in the conference long list

- Include a form for conference/new
- Include actions new and create in conference controller
- Remove the redirect (before action)
- Include authentication and confirm the allowed roles to include new conference (before action)
- Include controller tests

## New conference form:
conference_new

## Edit team page with a button to add new conference:

link_to_add_conf

### DIFF
--- a/app/controllers/conferences_controller.rb
+++ b/app/controllers/conferences_controller.rb
@@ -3,7 +3,6 @@ class ConferencesController < ApplicationController
   before_action :confirm_role, except: [:index, :show]
 
   def new
-    @team_id = params[:team_id]
     @conference = Conference.new
   end
 
@@ -16,7 +15,7 @@ class ConferencesController < ApplicationController
   end
 
   def create
-    team = Team.find(params[:team_id])
+    team = current_user.student_team
     @conference = Conference.new(conference_params)
     @conference.season_id = current_season.id
     @conference.gid = generate_gid(team)
--- a/app/controllers/conferences_controller.rb
+++ b/app/controllers/conferences_controller.rb
@@ -1,5 +1,6 @@
 class ConferencesController < ApplicationController
- before_action :redirect, except: [:index, :show, :new, :create]
+  before_action :authenticate_user!, except: [:index, :show]
+  before_action :confirm_role, except: [:index, :show]
 
   def new
     @conference = Conference.new
@@ -46,6 +47,10 @@ class ConferencesController < ApplicationController
     params.require(:conference).permit(
       :name, :twitter, :starts_on, :ends_on, :notes, :country, :region, :location, :city, :url, :season_id
     )
+  end
+
+  def confirm_role
+    redirect unless current_user.admin? || current_user.student?
   end
 
   def sort_params
--- a/app/controllers/conferences_controller.rb
+++ b/app/controllers/conferences_controller.rb
@@ -35,8 +35,8 @@ class ConferencesController < ApplicationController
 
   private
 
-  def generate_gid(conference)
-    conference.gid = Time.now.getutc
+  def generate_gid(team)
+    "2017-#{Time.now.getutc.to_i}-#{team.id}"
   end
  
   def conferences
--- a/app/controllers/conferences_controller.rb
+++ b/app/controllers/conferences_controller.rb
@@ -3,6 +3,7 @@ class ConferencesController < ApplicationController
   before_action :confirm_role, except: [:index, :show]
 
   def new
+    @team_id = params[:team]
     @conference = Conference.new
   end
 
@@ -17,11 +18,12 @@ class ConferencesController < ApplicationController
   def create
     @conference = Conference.new(conference_params)
     generate_gid(@conference)
+    @team = Team.find(params[:team_id])
 
     respond_to do |format|
       if @conference.save
-        format.html { redirect_to params[:redirect_to].presence || @conference, notice: 'Conference was successfully created.' }
-        format.json { render action: :show, status: :created, location: @conference }
+        format.html { redirect_to params[:redirect_to].presence || edit_team_path(@team), notice: 'Conference was successfully created.' }
+        format.json { render action: :edit, status: :created, location: @team }
       else
         format.html { render action: :new }
         format.json { render json: @conference.errors, status: :unprocessable_entity }
--- a/app/controllers/conferences_controller.rb
+++ b/app/controllers/conferences_controller.rb
@@ -18,6 +18,7 @@ class ConferencesController < ApplicationController
   def create
     team = Team.find(params[:team_id])
     @conference = Conference.new(conference_params)
+    @conference.season_id = current_season.id
     @conference.gid = generate_gid(team)
 
     if @conference.save
@@ -44,7 +45,7 @@ class ConferencesController < ApplicationController
 
   def conference_params
     params.require(:conference).permit(
-      :name, :twitter, :starts_on, :ends_on, :notes, :country, :region, :location, :city, :url, :season_id
+      :name, :twitter, :starts_on, :ends_on, :notes, :country, :region, :location, :city, :url
     )
   end
 
--- a/app/controllers/conferences_controller.rb
+++ b/app/controllers/conferences_controller.rb
@@ -23,7 +23,6 @@ class ConferencesController < ApplicationController
     if @conference.save
       redirect_to edit_team_path(team), notice: 'Conference was successfully created.'
     else
-      @team_id = team.id
       render action: :new
     end
   end
--- a/app/controllers/conferences_controller.rb
+++ b/app/controllers/conferences_controller.rb
@@ -34,7 +34,7 @@ class ConferencesController < ApplicationController
   private
 
   def generate_gid(team)
-    "2017-#{Time.now.getutc.to_i}-#{team.id}"
+    "#{Season.current.name}-#{Time.now.getutc.to_i}-#{team.id}"
   end
  
   def conferences
--- a/app/controllers/conferences_controller.rb
+++ b/app/controllers/conferences_controller.rb
@@ -16,13 +16,14 @@ class ConferencesController < ApplicationController
   end
 
   def create
+    team = Team.find(params[:team_id])
     @conference = Conference.new(conference_params)
-    generate_gid(@conference)
-    @team = Team.find(params[:team_id])
+    @conference.gid = generate_gid(team)
 
     if @conference.save
-      redirect_to edit_team_path(@team), notice: 'Conference was successfully created.'
+      redirect_to edit_team_path(team), notice: 'Conference was successfully created.'
     else
+      @team_id = team.id
       render action: :new
     end
   end
--- a/app/controllers/conferences_controller.rb
+++ b/app/controllers/conferences_controller.rb
@@ -1,5 +1,5 @@
 class ConferencesController < ApplicationController
- before_action :redirect, except: [:index, :show, :new]
+ before_action :redirect, except: [:index, :show, :new, :create]
 
   def new
     @conference = Conference.new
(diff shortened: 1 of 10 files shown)


mkalininait:
Hey @branquinhoaa Thank you for the PR! It works great for me!

One thing that was not very convenient: when you add a new conference, you get to its page and not to the team's edit page, where you can continue choosing your preferences. What do you think about it?

On another note, I've just noticed that we show the "Conference" column here: `/teams`. I'm not sure we should show conference *preferences* there.

branquinhoaa:
Yeah, about the column conference in /teams, I agree that we shouldn't show the teams preference...

branquinhoaa:
@mkalininait done! :dancer:

app/views/conferences/new.html.slim
```diff
@@ -0,0 +1,39 @@
...
+   h2 Conference location
+  fieldset
+    = f.input :country, required: true, as: :country, prompt: "Conference country", wrapper_html: { class: 'col-md-12' }
+    = f.input :region, required: true, collection: Conference::REGION_LIST, prompt: 'Conference region', wrapper_html: { class: 'col-md-12' }
+    = f.input :city, required: true, placeholder: 'Conference city', wrapper_html: { class: 'col-md-12' }
+   
+  .actions
+    ul.list-inline
+      li = f.submit 'Save', class: 'btn btn-success'
+      li = link_to 'Cancel', @conference, class: 'btn btn-default'
```
carpodaster: Does this cancel link work? I would assume that it cannot construct a route from an unpersisted Conference instance

app/views/teams/_form.html.slim
```diff
@@ -77,6 +77,7 @@
       = f.simple_fields_for :conference_preferences, f.object.conference_preferences.sort{ |a,b| a.option <=> b.option} do |c|
         = c.hidden_field :option, value: c.object.option
         = c.input :conference_id, collection: @conferences.group_by(&:region), as: :grouped_select, group_method: :last, label: c.object.option == 1 ? "Primary choice" : "Secondary Choice", prompt: "Conference", wrapper_html: { class: 'col-md-8' }
+    = link_to 'I want to suggest another conference!', new_conference_path(team: @team.id), class: 'btn btn-primary col-sm-offset-3'
```
carpodaster: For consistency, I would name the parameter `team_id` instead of just `team`

spec/controllers/conferences_controller_spec.rb
```diff
@@ -24,10 +24,31 @@
...
+      end
+
+      it 'are allowed to visit a new conference view' do
+        get :new
+        expect(response).to render_template("new")
+      end
+
+      it 'are allowed to create a new conference' do
+        expect{
+          post :create, params: { team_id: team.id ,conference: {name: 'name', country: 'country', region: 'region', location: 'location', city: 'city', season_id:'1'}}
```
carpodaster: For better readability, I recommend putting the conference params into a let variable

app/controllers/conferences_controller.rb
```diff
@@ -12,16 +15,46 @@ def show
...
 
+  def create
+    @conference = Conference.new(conference_params)
+    generate_gid(@conference)
+    @team = Team.find(params[:team_id])
+
+    respond_to do |format|
+      if @conference.save
+        format.html { redirect_to params[:redirect_to].presence || edit_team_path(@team), notice: 'Conference was successfully created.' }
+        format.json { render action: :edit, status: :created, location: @team }
```
carpodaster: I don't think we need to respond to json in this action. We don't have an API client (yet :wink: )

app/controllers/conferences_controller.rb
```diff
@@ -12,16 +15,46 @@ def show
...
+    conference.gid = Time.now.getutc
+  end
  
   def conferences
     Conference.ordered(sort_params).in_current_season
   end
 
+  def conference_params
+    params.require(:conference).permit(
+      :name, :twitter, :starts_on, :ends_on, :notes, :country, :region, :location, :city, :url, :season_id
```
carpodaster: If the assignment of the season is handled in the controller, the `season_id` can be removed from this list.

app/models/conference.rb
```diff
@@ -4,6 +4,14 @@ class Conference < ActiveRecord::Base
 
   has_many :conference_preferences, dependent: :destroy
   has_many :attendees, through: :conference_preferences, source: :team
+  REGION_LIST = [
```
carpodaster: Can you add another empty line before the constant please? Also, it's better to define constants at the very top of a class.

app/views/conferences/new.html.slim
```diff
@@ -0,0 +1,39 @@
+nav.page
+  .back-nav
+    ul
+      li = icon('chevron-left')
+      li = link_to 'Conferences', class: 'back'
```
carpodaster: This link does not have a href

app/views/teams/_form.html.slim
carpodaster: Ah, I just came to realise that the team parameter is probably only to eventually redirect back to the team's page. Can't we determine the student's team in the controller? This could be somewhat related to #790 (i.e. determining the currently relevant student team if a user is part of more than one team). It's not a blocker for this PR but it would be good if you gave it a try.

carpodaster: Furthermore, this link will discard whatever has been entered into the form. It should either open in a new tab or add a JS confirmation popup before it follows the link.  The latter would be preferable since a newly created conference would not show up in the open form. The page needs to be reloaded anyway to repopulate the select box.

app/controllers/conferences_controller.rb
```diff
@@ -12,16 +15,46 @@ def show
...
+  end
+
   def redirect
     redirect_to orga_conferences_path
   end
 
   private
+
+  def generate_gid(conference)
+    conference.gid = Time.now.getutc
```
branquinhoaa: Nice idea! 
Maria asked for some way to distinguish when a conference was suggested by a team or not. However, the idea is that the team suggests - not the student. So maybe we can make this GID in this format:
"2017-#{Time.now.getutc.to_i}-teamID"

What do you think?

app/controllers/conferences_controller.rb
branquinhoaa: The season_id comes from the conferences/new form (but it is a hidden param - always set to current season). Do you think it is better to handle it in the controller?

app/views/conferences/new.html.slim
branquinhoaa: :see_no_evil:

app/controllers/conferences_controller.rb
branquinhoaa: Or maybe:  "2017-ConferenceID-teamID"

branquinhoaa:
Hey @carpodaster, thank you for your suggestions!
I'm not sure about how the GID should be, but for now, I have implemented in this format:
"2017-#{Time.now.getutc.to_i}-#{teamID}"
I also had to change the column from integer to string

mkalininait:
@branquinhoaa Thank you! I think the feature can be used already. 

There is still a room for usability improvements though. Now, when I press the "Add another conference" button, the browser warns me "You are leaving the page, all your alterations will be lost, do you want to continue?" (meaning, I have a not very comfortable choice between loosing my changes and getting the function AND not getting the function at all). And then when I add a conference, I return back to the edit page, see the very top of it and need to scroll all the way down to the conference section.

In perfect case (as I see it), the "Add new conference" form should appear in a pop-up form (or something like that), so that you can add a new conference without interrupting your edit process. What do you think @carpodaster ?

But as I said before, we can merge the PR as is, too. I don't want to overwhelm anyone!

carpodaster:
@mkalininait a popup (or better yet: an in-place edit) would be best, that's for sure. However, we need to repopulate the select box where the student selects the conferences from. That adds a blast of javascript logic to this PR. My idea when suggesting the JS confirmation dialogue was that you can still save whatever edits you made and continue adding a new conf afterwards. It's definitely bumpy. Wdyt, @branquinhoaa and @mkalininait, should we go the extended javascript route?

app/controllers/conferences_controller.rb
```diff
@@ -16,18 +15,16 @@ def show
...
-        format.html { render action: :new }
-        format.json { render json: @conference.errors, status: :unprocessable_entity }
-      end
+    @conference.season_id = current_season.id
+    @conference.gid = generate_gid(team)
+
+    if @conference.save
+      redirect_to edit_team_path(team), notice: 'Conference was successfully created.'
+    else
+      @team_id = team.id
```
carpodaster: This is not needed anymore

app/controllers/conferences_controller.rb
```diff
@@ -16,18 +15,16 @@ def show
   end
 
   def create
+    team = current_user.student_team
```
carpodaster: Ah, very good. I wasn't aware this method existed! :+1:

app/controllers/conferences_controller.rb
```diff
@@ -37,8 +34,8 @@ def redirect
 
   private
 
-  def generate_gid(conference)
-    conference.gid = Time.now.getutc
+  def generate_gid(team)
+    "2017-#{Time.now.getutc.to_i}-#{team.id}"
```
carpodaster: Yes, much better to put the team in instead of the user id. However, the year should be hardcoded but retrieved dynamically (`Season.current.name`).

app/views/teams/_form.html.slim
```diff
@@ -77,7 +77,7 @@
       = f.simple_fields_for :conference_preferences, f.object.conference_preferences.sort{ |a,b| a.option <=> b.option} do |c|
         = c.hidden_field :option, value: c.object.option
         = c.input :conference_id, collection: @conferences.group_by(&:region), as: :grouped_select, group_method: :last, label: c.object.option == 1 ? "Primary choice" : "Secondary Choice", prompt: "Conference", wrapper_html: { class: 'col-md-8' }
-    = link_to 'I want to suggest another conference!', new_conference_path(team: @team.id), class: 'btn btn-primary col-sm-offset-3'
+    = link_to 'I want to suggest another conference!', new_conference_path, data: { confirm: 'You are leaving the page, all your alterations will be lost. Do you want to continue?' }, class: 'btn btn-primary col-sm-offset-3'
```
carpodaster: :+1:

db/migrate/20170727145559_change_data_type_for_gid.rb
```diff
@@ -0,0 +1,12 @@
+class ChangeDataTypeForGid < ActiveRecord::Migration[5.1]
+  def self.up
+    change_table :conferences do |t|
+      t.change :gid, :string
```
carpodaster: Very good catch!

spec/factories/conferences.rb
```diff
@@ -1,8 +1,12 @@
 FactoryGirl.define do
   factory :conference do
     name { [FFaker::CheesyLingo.title, 'Conf'].join ' ' }
+    url { FFaker::Internet.http_url }
     starts_on { Time.utc(Date.today.year, 7, 7) }
     ends_on { Time.utc(Date.today.year, 7, 15) }
+    city { FFaker::Address.city }
+    country { FFaker::Address.country }
+    region { 'Africa' }
```
carpodaster: (No need to change this, just a Factory Girl hint: you only need to pass a block for dynamic values. The value `'Africa'` is static and can be passed as a normal string argument)

spec/models/conference_spec.rb
```diff
@@ -6,6 +6,10 @@
   it { is_expected.to have_many(:conference_preferences).dependent(:destroy) }
   it { is_expected.to have_many(:attendees) }
   it { is_expected.to validate_presence_of(:name) }
+  it { is_expected.to validate_presence_of(:url) }
+  it { is_expected.to validate_presence_of(:city) }
+  it { is_expected.to validate_presence_of(:country) }
+  it { is_expected.to validate_presence_of(:region) }
```
carpodaster: Yes! :smiley_cat:

mkalininait:
@carpodaster @branquinhoaa I'd suggest to go with the easiest solution for now (probably as-is?), because we need to start the conference process asap (like, really asap). Maybe we create a separate issue to improve the usability of this feature?

app/controllers/conferences_controller.rb
branquinhoaa: Yeah, we created it on our last PR! :dancing_women:

branquinhoaa:
@mkalininait @carpodaster I have just made the last changes!
I think it is a nice idea to have another issue for usability  :+1:

Issue #769, opened 2017-06-24
[Conferences] #5: Terms and conditions


mkalininait:
The idea: https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/751#issuecomment-308574743
Full list of conference user stories: https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/751#issuecomment-309138513

**User Story:**
As an RGSoC orga, I want the students to be aware of the terms on which we hand out free conference tickets and confirm their (students') obedience.

**Things to do:**
On the Team's page, in the Conference preferences block (see Issue #767) add the following *mandatory* fields (must be checked by students):
 * (checkbox) We understand that RGSoC doesn't guarantee tickets to the conferences of our choice
 * (checkbox) We understand that all travel and accommodation costs will have to be covered by us; hereby we confirm that we are choosing conferences which we can afford

mkalininait:
Resolved with https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/810

Issue #768, opened 2017-06-24
[Conferences] #4: Additional info about about chosen conferences


mkalininait:
The idea: https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/751#issuecomment-308574743
Full list of conference user stories: https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/751#issuecomment-309138513

**User Story:**
As a student, I want to provide additional information to my team's conference choices: conditions / limitations / wishes etc., as well as our potential wish to give a lightning talk.

**Things to do:**
On the Team's page, in the Conference preferences block (see Issue #767) add the following fields:
 * (checkbox) We would like to give a lightning talk if possible
 * (text area) Comments 
   * hint below: E.g. "We would attend conference X only in they provide scholarships / have a LT session" or "We have already applied for the conference X ourselves" etc.

mkalininait:
Resolved with https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/810

PR #805, opened 2017-07-28
Conferences/aditional info

juuh42dias:
Related issues  #768 and  #769

Edit teams view before additional info

edit teams view old

Edit teams view after additional info 

edit teams view new

### DIFF
--- a/app/assets/javascripts/conferences_options.js
+++ b/app/assets/javascripts/conferences_options.js
@@ -1,5 +1,5 @@
 $(function() {
-  $('#team_conference_preferences_attributes_0_conference_id').on("change", function() {
+  $('#team_conference_preference_info_attributes_conference_preferences_attributes_0_conference_id').on("change", function() {
       $("select option").attr("disabled",false);
       disableOptions();
     });
--- a/app/controllers/teams_controller.rb
+++ b/app/controllers/teams_controller.rb
@@ -26,8 +26,7 @@ class TeamsController < ApplicationController
   def edit
     @team.sources.build(kind: 'blog') unless @team.sources.any?
     @conferences = conference_list
-    @team.conference_preferences.build(option: 1) unless @team.conference_preferences.find_by(option: 1)
-    @team.conference_preferences.build(option: 2) unless @team.conference_preferences.find_by(option: 2)
+    @team.with_all_built
   end
 
   def create
@@ -47,14 +46,10 @@ class TeamsController < ApplicationController
 
   def update
     @conferences = conference_list
-    respond_to do |format|
-      if @team.update_attributes(team_params)
-        format.html { redirect_to @team, notice: 'Team was successfully updated.' }
-        format.json { head :no_content }
-      else
-        format.html { render action: :edit }
-        format.json { render json: @team.errors, status: :unprocessable_entity }
-      end
+    if @team.update_attributes(team_params)
+      redirect_to @team, notice: 'Team was successfully updated.'
+    else
+      render action: :edit
     end
   end
 
@@ -84,7 +79,8 @@ class TeamsController < ApplicationController
         :'finishes_on(1i)', :'finishes_on(2i)', :'finishes_on(3i)', :invisible,
         :project_name,
         roles_attributes: role_attributes_list,
-        conference_preferences_attributes: [:id, :option, :conference_id, :_destroy],
+        conference_preference_info_attributes: [:id, :lightning_talk, :condition_term_ticket,
+          :condition_term_cost, :comment, :team_id, :_destroy, conference_preferences_attributes: [:id, :option, :conference_id, :_destroy]],
         sources_attributes: [:id, :kind, :url, :_destroy]
       )
     end
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -49,8 +49,8 @@ class Ability
       user.roles.organizer.any? || team.supervisors.include?(user)
     end
 
-    can :crud, ConferencePreference do |preference|
-      user.admin? || (preference.team.students.include? user)
+    can :crud, ConferencePreferenceInfo do |cp_info|
+      user.admin? || (cp_info.team.students.include? user)
     end
 
     can :read, Mailing do |mailing|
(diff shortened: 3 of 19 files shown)


mkalininait:
Hey @juuh42dias Thank you for the PR! I've tested it (but didn't check the code).
It works well for me!

A couple of UI notes:
1) Is there a reason why the checkboxes are aligned differently that all the other fields on the form? 
2) For the "terms and conditions" checkboxes I would probably remove the current titles and use the hints instead of the titles. The short titles sound a bit unclear to me.
3) I was playing with the team edit form and I'm not that sure anymore, that we should validate the "terms and conditions" when we save the form. It may be that a user wants to edit her team's blog url, and we're kinda forcing her to check the conference terms and conditions, too. It's probably not a good UX... What do you think?

klappradla:
Hey @juuh42dias thanx for the PR 💚 

I already looked into bits of the code and it made me wonder if the data model you chose may eventually make things harder for you as they needed to be. Therefore, can you upfront, before I dig deeper into this maybe quickly explain why you chose this data model?

I will briefly explain what I think would be a simpler approach - no need for you to take this as a source of truth and implement it like this of course! Please feel fee to always ping me on Slack if something is unclear or I need to explain things in more detail _(I reduced my active Slack teams to < 10 in the last week - should be able to respond fairly quick now)_

**How to add new attributes?**
I have two possible approaches, which depends on how the new `comment`, `lightning_talk` etc. attributes should be added to the `ConferencePreference` model. I would also love to have a word by @mkalininait on this, how it makes most sense for you. I see two approaches:
1. ONE `comment`, `lightning_talk` etc. for ALL preferences
1. ONE `comment`, `lightning_talk` etc. for PER preference _(different to how it is implemented in this PR)_

### 1. ONE `comment`, `lightning_talk` etc. for ALL preferences
_(this is the approach you chose - telling from e.g. the screenshot)_

If we take the `ConferencePreference` model as a starting point for orga users later on confirming actual attendences, it would be easiest to have the whole thing as one entity on the `Team`. So instead of having a `conference_id` and an `option` attribute on the model, it could also look like this:
```ruby
create_table "conference_preferences", id: :serial, force: :cascade do |t|
  t.bigint "team_id"
  t.bigint "conference1_id"
  t.bigint "conference2_id"
# etc.
```

If done like this, there's also no reason for an additional `ConferencePreferenceInfo` model - all the attributes, e.g. the `comment` can simply be added to the `ConferencePreference` object itself, so you end up with a simple 1 x 1 association between the `Team` and the `ConferencePreference` object:

untitled diagram 1


In terms of Rails, this afaik basically just means, the `ConferencePreference` model has two belongs-to associations to the `Conference` table:
```ruby
class ConferencePreference
  belongs_to :conference1, foreign_key: "conference1_id", class_name: "Conference"
  belongs_to :conference2, foreign_key: "conference2_id", class_name: "Conference"
```

For the two `terms_and_condition` types of check-boxes: these in my opinion can also be "virtual" attributes not persisted to the DB but only present on the object instance. Maybe take look at ActiveRecord's [acceptance type of validation](http://guides.rubyonrails.org/active_record_validations.html#acceptance).

### 2 ONE `comment`, `lightning_talk` etc. for PER preference
_(different to how you implemented it in this PR)_

For this approach, I'd also spare out the `ConferencePreferenceInfo` model and instead add the new attributes straight onto the `ConferencePreference` model. This approximately looks like this:

untitled diagram 2

Again, the `terms_and_condition` checkboxes can probably be added as non-persisted attributes.

### Pros / Cons / Preferences
Depending on @mkalininait 's initial idea of this, I'd personally prefer the first approach, since it's just the simplest way possible. There's not even the need to validate the number of `ConferencePrefereces` per `Team` (since it will always be just one), etc.

Of course, the requirements may potentially change within the next years, but I'm personally not a huge fan of building things more complex upfront for things that *"may*" eventually come 😉 

I'm putting this on "request changes" just for this question I have here _(it helps me to keep track of my Github notifications if I already see, that I did an initial review on this)_

juuh42dias:
Hi, @klappradla :) 

So, we created the cp_info model because we would have more autonomy in controlling the information related to the team who chose the conference_preferences. Even if there are not any changes soon, the table already allows modifications in a way that avoids much future rework. In other words, we thought this implementation would allow us to easily introduce changes that belong to one of the preferences or even to both. We have suggested this approach for @Maria, but if you want we can change it.

klappradla:
Ah alright @juuh42dias, thanx for coming back to my question so quick 🙏 

I tbh don't see the benefit it a lesser coupling between `ConferencePreference` and `ConferencePreferenceInfo`. The benefit of being able to then easier add new team-related attributes to it comes at the cost of having to maintain the way a way more complex association tree, a more complex form, etc., together with making sure, all of these objects depending on each other are available in the right moment and are deleted in the right moment.

So in my opinion, the maintenance and complexity overhead is too high a price for the benefits.

Aside that: I think the question whether there's **one info block per `Team`** vs. one info block per `ConferencePreference` should be answered before. It also does not make much sense to implement it it a way loose enough to be able to jump between these two things in my opinion. None foresees the future, so there's no point in introducing complexity and technical dept for requirements that may never come 🐌

mkalininait:
Hey @klappradla Thanks for the input!
The idea was to implement the "1. ONE comment, lightning_talk etc. for ALL preferences" strategy. I agree that it should be sufficient to have the `ConferencePreference` model and I don't see why we need the `ConferencePreferenceInfo` model.

We indeed discussed the models with @juuh42dias and Amanda some days ago; I looked to me that we agreed on exactly the same solution tbh. Maybe I misunderstood you @juuh42dias ? If so, I'm sorry! I thought we were on the same page.

klappradla:
Yay @mkalininait ✌️ - guess we're at least all on the same page _now_ 😉

juuh42dias:
1. ONE comment, lightning_talk etc. for ALL preferences.

@mkalininait  sounds good for me (too). I'm already implementing this new version of the relations.

juuh42dias:
@mkalininait @klappradla 
So, i think is better create a new PR with new changes, that's ok? :smile_cat:

mkalininait:
@juuh42dias This is ok with me :)

klappradla:
Alright @juuh42dias, so I'm closing this one, since you're working on a new one now ✌️

Issue #711, opened 2017-03-26
Locations in the list of teams is in duplicates


amrtanair:
For example, if the students are from the same place and have entered the location in a different fashion(say, one has entered it in capitals while the other has not). 

Image attached below: 
![whatsapp image 2017-03-26 at 9 34 55 am](https://cloud.githubusercontent.com/assets/16919437/24328451/9a90e772-1207-11e7-9126-b42742b9b576.jpeg)

amrtanair:
Can I try tackling this issue?

beanieboi:
@amrtanair that would be really cool! I'm happy to help if you have questions.

F3PiX:
Hi @amrtanair ! I take it you are not working on this, so I assigned myself. If you do want to take this on, please unassign me and assign yourself!

amrtanair:
Hi @F3PiX  ! I am sorry I took so long to reply. I am currently working on another project for RGSoC'17 and I would love to solve this issue. it kinda was on the back of my head, but I never got around to working on it :( If you want to, do go ahead, else just notify me and I will make it a point to resolve it :')

F3PiX:
@carpodaster I'd like to add the very cool [geocoder gem](https://github.com/alexreisner/geocoder) as a starting point for solving this. Is that okay?

carpodaster:
@F3PiX Why do you think we need exact geocoords? I'm not against geocoder, but I don't currently see why we would need it to de-duplicate locations. Your solution in #807 seems like the better approach to me.

F3PiX:
Apart from  all the cool things we can do with it... 
The solution on #807 works, given the user input is correct. If someone enters a city incorrect, or one student calls their city "Beijing" and the second student "Peking", we can't catch that now. With geocoding we get one unique name.
(We do not need the geocoder gem for that, Max already introduced geocoding for the application forms, but the gem offers a lot of possibilities. )

F3PiX:
Hmmm either way, it is probably a bad idea to do this mid-season. 
Closing this, and I'll check back in after summer.

PR #807, opened 2017-07-29
Solve duplicate locations in Teams Table

F3PiX:
Solves #711 
Solved by unifying the user location on :save.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -104,7 +104,7 @@ class User < ActiveRecord::Base
 
   accepts_nested_attributes_for :roles, allow_destroy: true
 
-  before_save :sanitize_location
+  before_save :format_location
   after_create :complete_from_github
 
   # This field is used to skip validations when creating
@@ -218,10 +218,9 @@ class User < ActiveRecord::Base
 
   private
 
-  # Ensures that the location column either contains non-whitespace text, or is NULL
-  # This ensures that sorting by location yields useful results
-  def sanitize_location
-    self.location = nil if self.location.blank?
+  def format_location
+    self.location = nil if self.location.blank? # for sorting
+    self.location = self.location.downcase.titlecase unless location.nil?
   end
 
   def complete_from_github
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -104,7 +104,7 @@ class User < ActiveRecord::Base
 
   accepts_nested_attributes_for :roles, allow_destroy: true
 
-  before_save :format_location
+  before_save :normalize_location
   after_create :complete_from_github
 
   # This field is used to skip validations when creating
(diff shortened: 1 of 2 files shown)


F3PiX:
This solves the duplicates, as long as the user input is correct. The 'city' field in the user profile is a text field, so the user input is not super reliable. 
See also https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/711#issuecomment-318818343.

app/models/user.rb
```diff
@@ -104,7 +104,7 @@ def student
 
   accepts_nested_attributes_for :roles, allow_destroy: true
 
-  before_save :sanitize_location
+  before_save :format_location
```
F3PiX: Or: `normalize_location` ? WDYT?

F3PiX:
screen shot 2017-07-29 at 16 01 12

app/models/user.rb
carpodaster: I like normalize a bit better

app/models/user.rb
```diff
@@ -218,10 +218,9 @@ def student_team
 
   private
 
-  # Ensures that the location column either contains non-whitespace text, or is NULL
-  # This ensures that sorting by location yields useful results
-  def sanitize_location
-    self.location = nil if self.location.blank?
+  def format_location
+    self.location = nil if self.location.blank? # for sorting
+    self.location = self.location.downcase.titlecase unless location.nil?
```
carpodaster: You don't need to `#downcase` before `#titlecase`

carpodaster: We could also add a `String#strip`. The entire method body could be condensed to:

```ruby
self.location = location.to_s.strip.titlecase.presence
```

app/models/user.rb
```diff
@@ -218,10 +218,9 @@ def student_team
 
   private
 
-  # Ensures that the location column either contains non-whitespace text, or is NULL
```
carpodaster: It would be better to leave some comment (normalisation to de-duplicate + NULL for sorting)

app/models/user.rb
F3PiX: 😍 `nil.to_s.presence` 
I prefer to add `downcase` to `titlecase` because of this:

` " AMsterdam ".strip.downcase.titlecase # => "Amsterdam" ` 

` " AMsterdam ".strip.titlecase #=> "A Msterdam" `

It was implicitly tested in the specs; I added a more explicit test.

carpodaster: Ah, very good! I only tested it with UPPER CASE in the console

spec/models/user_spec.rb
```diff
@@ -180,10 +180,15 @@
       end
 
       it 'unifies the location format' do
-        subject.location = "nEw yorK"
+        subject.location = "nEw york"
         expect { subject.save }.to change { subject.location }.to("New York")
       end
 
+      it 'does not separate words with foul caps' do
```
carpodaster: :ok_hand:

F3PiX:
Thanks!!

Issue #806, opened 2017-07-28
Conference preferences: NONE option


mkalininait:
I think we should add the 'None' option to the conference preference dropdowns, in case a team doesn't want to attend any conference. Now it's impossible to set this option after a conference has been already selected and saved.

juuh42dias:
@mkalininait  in my PR was included `include_blank: 'Conference' ` on Dropdown of Conferences.

klappradla:
Yes, should already be fixed with #810 , you're right @juuh42dias 🎉  _(closing this @mkalininait , feel free to reopen if it the workflow does not cover all your use cases)_

mkalininait:
@klappradla @juuh42dias Can we please rename it to "None"? Other than that it looks good to me.

klappradla:
Should be resolved now via #812

mkalininait:
@klappradla YAP!

PR #808, opened 2017-07-29
Conferences/create none option

branquinhoaa:
PR related with the issue #806 

- create none option for team

### DIFF
--- a/app/assets/javascripts/conferences_options.js
+++ b/app/assets/javascripts/conferences_options.js
@@ -1,7 +1,9 @@
 $(function() {
-  $('#team_conference_preferences_attributes_0_conference_id').on("change", function() {
+  $('.selected_conference').on("change", function() {
+    var selected = $(this);
       $("select option").attr("disabled",false);
       disableOptions();
+      verifyNoneSelected(selected);
     });
 
   function disableOptions(){
@@ -9,7 +11,7 @@ $(function() {
       var bSuccess=false;
       var selectedEl=$(this);
       $("select option:selected").each(function(){
-        if($(this).val()===selectedEl.val()){
+        if($(this).val()===selectedEl.val() && $(this).val() != ''){
           bSuccess=true;
           return false;
         }
@@ -17,4 +19,10 @@ $(function() {
         return bSuccess;
       }).css("display","none");
     }
+
+  function verifyNoneSelected($select){
+    var $optionSelected = $select.find('option:selected');
+    var $parent = $select.parent('.fields');
+    $parent.find('.destroy_option').val($optionSelected.val() === '');
+  }
 });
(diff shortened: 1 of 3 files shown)


branquinhoaa:
@mkalininait :+1: 
Just some things:
- not sure about the name 'none'. 
- there is a strange behavior in the form: the destroy only works if I explicitly pass the object id it in the view... no idea why!

mkalininait:
@branquinhoaa I didn't manage to make it run locally yet; I get some errors I don't understand yet ("undefined method `stringify_keys' for "/conferences/new":String", app/views/teams/edit.html.slim, line 14). Maybe I'll resolve it tomorrow.

>there is a strange behavior in the form: the destroy only works if I explicitly pass the object id it in the view... no idea why!

Could you please clarify what you mean by this?

branquinhoaa:
@mkalininait
weird, I'm not getting errors in my local environment... 
Did you discover what is wrong?

about my comment:
'there is a strange behavior in the form: the destroy only works if I explicitly pass the object id it in the view... no idea why!'

I mean that I had to put the conference_id in the form, otherwise the controller did not recognize it.

mkalininait:
@branquinhoaa I don't know, maybe I did something wrong with merges. I was missing the column `conference_preferences.team_id` after checkout/pull and tried to merge with `juuh42dias:conferences/aditional-info`. It helped but caused the error above.

mkalininait:
@branquinhoaa I managed to run the PR locally. It seems to work!
One minor thing: before User saves her preferences for the first time, she sees both 'none' and 'conference' options. Why do we need the 'conference' option? It disappears once User saves preferences. And in general it looks useless in presence of the 'none' option.

mkalininait:
@branquinhoaa 

> there is a strange behavior in the form: the destroy only works if I explicitly pass the object id it in the view... no idea why!
> I mean that I had to put the conference_id in the form, otherwise the controller did not recognize it.

¯\_(ツ)_/¯ 
Maybe @carpodaster knows.

mkalininait:
@branquinhoaa Works perfect! Good to merge from my side!

mkalininait:
Hey @klappradla This is a very little PR, do you think we can merge it? It's good to go from my side.
Not sure when Carsten will be able to look at it...

branquinhoaa:
Hey @klappradla, I think this PR is no longer useful after all the changes in ConferencePreference model (PR  #810). 
I was passing a _destroy: true in the cases when I needed to destroy the conference, but now we cannot do that (I guess)...
What do you think?

klappradla:
Sorry for coming back to this so late - my brain is kinda melted today 🙈 

You're right @mkalininait this could _"theoretically"_ be merged, I don't see a problem. However and there you're right @branquinhoaa it indeed does not make too much sense when combined with the changes we agreed upon for #810 .

Therefore I'd rather close this one in order to not introduce code that's not tested (the Javascript code) and that we'll with 100% certainty want to get rid off within implementing the next issues anyways. Are you both fine with that?

Aside that @branquinhoaa you're approach for passing the `_destroy` is totally valid 👍  I'd have probably solved it differently and also for the sake of easier testability without feature specs would have preferred a way of doing it _without_ Javascript, but that comes down to personal preference :v:

PR #810, opened 2017-08-01
Conferences/additional student team conference info

juuh42dias:
Edit teams view old:
edit teams view old

Edit teams view new 2.1
edit teams view new

Show team view new:
team 16 elastic rgsoc teams app

:


mkalininait:
Hey @juuh42dias , could you please move the button "I want to suggest another conference!" between the "Secondary choice" dropdown and the "Comment" input?

The rest I didn't test, because the migrations fail (as I reported to you on Slack).

mkalininait:
@juuh42dias I still can't run this PR locally, because the migrations fail. 
```
== 20170731173902 AddColumnsIntoConferencePreferences: migrating ==============
-- add_column(:conference_preferences, :comment, :text)
   -> 0.0039s
-- f()
rails aborted!
StandardError: An error has occurred, this and all later migrations canceled:

undefined method `f' for #<AddColumnsIntoConferencePreferences:0x007f9cad4306d8>
/Users/mronacher/Development/rgsoc-teams/db/migrate/20170731173902_add_columns_into_conference_preferences.rb:4:in `up'
bin/rails:4:in `require'
bin/rails:4:in `<main>'
NoMethodError: undefined method `f' for #<AddColumnsIntoConferencePreferences:0x007f9cad4306d8>
/Users/mronacher/Development/rgsoc-teams/db/migrate/20170731173902_add_columns_into_conference_preferences.rb:4:in `up'
bin/rails:4:in `require'
bin/rails:4:in `<main>'
Tasks: TOP => db:migrate
(See full trace by running task with --trace)
```

Can anyone @carpodaster @klappradla please look into it? Maybe I'm going something wrong? I don't want the PR to be stuck because of this :/

juuh42dias:
Migration error was just an incomplete word :see_no_evil:

app/controllers/orga/teams_controller.rb
```diff
@@ -48,7 +47,7 @@ def create
   def update
     @conferences = conference_list
     respond_to do |format|
-      if @team.update_attributes(team_params)
+      if @team.update_attributes(team_params) && terms_accepted?
```
klappradla: I would prefer to use an [acceptance](http://guides.rubyonrails.org/active_record_validations.html#acceptance) validation here instead. I personally find it way clearer, since it does not leak any validation code from the model into the controller.

It works like this:

1. You define a validation on the model:
```ruby
class ConferencePreference
  validates :terms_of_ticket, acceptance: true
  validates :terms_of_travel, acceptance: true
end
```
☝️  note that these two attributes **are not** backed by a DB column

2. Add the necessary field to the form:
```ruby
= f.simple_fields_for :conference_preference do |c|
  = c.input :first_conference_id, collection: @conferences.group_by(&:region)
  = c.input :second_conference_id, collection: @conferences.group_by(&:region)
  = c.input :comment
  = c.input :terms_of_ticket, as: :checkbox
  = c.input :terms_of_travel, as: :checkbox
  # etc.
```

3. And then simply permit it in the controller

app/controllers/orga/teams_controller.rb
```diff
@@ -89,6 +88,12 @@ def conference_list
     Conference.in_current_season
   end
 
+  def terms_accepted?
```
klappradla: As said in my comment above, I'd suggest to solve this via an acceptance validation - way cleaner in my eyes :v:

app/controllers/orga/teams_controller.rb
klappradla: You could even go one step further and only validate these field if the user is non-admin 😉 
But that's a bonus and not super necessary for now.

app/controllers/orga/teams_controller.rb
```diff
@@ -78,9 +77,9 @@ def team_params
       :name, :twitter_handle, :github_handle, :description, :post_info, :event_id,
       :checked, :'starts_on(1i)', :'starts_on(2i)', :'starts_on(3i)',
       :'finishes_on(1i)', :'finishes_on(2i)', :'finishes_on(3i)', :invisible,
-      :season_id, :kind, :project_name,
+      :project_name, :season_id, :kind,
+      conference_preference_attributes: [:id, :first_conference_id, :second_conference_id, :lightning_talk, :comment, :_destroy],
```
klappradla: When switching to using a validation for the two `terms` fields, they have to be permitted here as well of course.

app/controllers/teams_controller.rb
```diff
@@ -48,7 +47,7 @@ def create
   def update
     @conferences = conference_list
     respond_to do |format|
-      if @team.update_attributes(team_params)
+      if @team.update_attributes(team_params) && terms_accepted?
```
klappradla: Same as for the orga controller - I think this can be solved cleaner via an acceptance validation.

In addition to that, orga people should not really need to check the two terms-checkboxes anyways. But as said, that's an extra thing, not super necessary for now.

app/models/conference.rb
```diff
@@ -10,15 +10,14 @@ class Conference < ActiveRecord::Base
 
   include HasSeason
 
-  has_many :conference_preferences, dependent: :destroy
-  has_many :attendees, through: :conference_preferences, source: :team
+  has_many :preferences_first_conferences, :class_name => 'ConferencePreference', :foreign_key => 'first_conference_id'
```
klappradla: Comes down to personal preference, but I'd say a naming like:
```ruby
has_many :first_choice_conference_preferences
has_many :second_choice_conference_preferences
```
could be a bit more readable. What do you think?

app/models/conference_preference.rb
```diff
@@ -1,5 +1,6 @@
 class ConferencePreference < ActiveRecord::Base
 
-  belongs_to :team, inverse_of: :conference_preferences
-  belongs_to :conference, inverse_of: :conference_preferences
-end
+  belongs_to :team, inverse_of: :conference_preference
+  belongs_to :first_conference, class_name: 'Conference'
```
klappradla: Again personal preference, but I'd simply name this:
```ruby
belongs_to :conference1
belongs_to :conference2
```
Semi aligned with how it's done for the project choices for teams and with the idea in mind to not indicate a) both needing to be present and b) that teams to to two conferences _(first to the first one, then to the second one)_

app/views/orga/teams/_form.html.slim
```diff
@@ -80,9 +80,27 @@
...
+          .col-sm-9
+            .checkbox
+              label
+                = c.check_box :lightning_talk
+                |  I would like to give a lightning talk!
+          label.control-label.col-sm-3 Terms
+          .col-sm-9
+            .checkbox
+              label
+                = check_box_tag :terms_ticket, @team.conference_preference.present?, @team.conference_preference.present?
```
klappradla: As said above, when using an acceptance validation for the terms fields, this will simply be an input on the form object itself, same as e.g. the comment.

app/views/teams/show.html.slim
```diff
@@ -86,16 +86,25 @@ nav.page
...
+- if @team.conference_preference.present?
   h4 Conference preferences
   ul.conferences
-  - @team.conference_preferences.sort_by(&:option).each do |preference|
-    li
-      = links_to_conferences([preference.conference]).first
-      = preference_option(preference)
+    - if @team.conference_preference.first_conference.present?
+      li
+        = links_to_conferences([@team.conference_preference.try(:first_conference)]).first
```
klappradla: Given the conditional 2 lines above, I don't see a point in using `try` here - there'll always be a value different to `nil`. Aside that, I would suggest to just render the link "normally", the usage of the helper looks kind of overkill to me here, don't you think?

app/views/teams/show.html.slim
```diff
@@ -86,16 +86,25 @@ nav.page
...
+        = content_tag :span, 'First conference', class: "label label-primary"
```
klappradla: I'd align this with the other view and speak of "First choice" and "Secondary choice" here.

mkalininait: I'd say first & second OR primary & secondary :)

mkalininait:
Hey @juuh42dias The new features look good to me with the exception of the Terms validation. 
As we discussed before, we should only verify that the Terms are checked if at least one conference is selected. Now it's not the case, and I can't save the team edit form unless I accept the Terms (even if I didn't select a conference).

mkalininait:
@juuh42dias Great job!! Good to merge from my side. /cc @klappradla (didn't look in the code)

app/models/conference_preference.rb
```diff
@@ -1,5 +1,21 @@
 class ConferencePreference < ActiveRecord::Base
+  validates :terms_accepted?, acceptance: true
```
klappradla: Almost 💪 

You can spare the two `attr_accessor`s - it's just about the input that's coming in via the controller (essentially the form in the end). So you need two validations:
```ruby
validates : terms_of_ticket?, acceptance: true
validates : terms_of_travel?, acceptance: true
```

Plus: remove the `terms_accepted?` method, Rails does all this alone, including putting together the respective error message. The practical part about this: the validation only works if the attribute is give - so you'll still be able to create and update `ConferencePreferences` outside the context of user-interaction.

I put together a small [Rails app for you as an example](https://gist.github.com/klappradla/80eeda7538fd5a924ef83817156b8ffe):
```ruby
begin
  require "bundler/inline"
rescue LoadError => e
  $stderr.puts "Bundler version 1.10 or later is required. Please update your Bundler"
  raise e
end

gemfile(true) do
  source "https://rubygems.org"

  gem "rails"
  gem "pg"
  gem "pry"
end

require "active_record"

ActiveRecord::Base.establish_connection(adapter: "postgresql", database: "validation_example")

ActiveRecord::Schema.define do
  create_table :users, force: true do |t|
    t.string :name
    t.timestamps
  end
end

class User < ActiveRecord::Base
  validates_acceptance_of :terms_of_service
end

binding.pry
```

It's a Rails app in one file. In order to run it, you have to do:
```sh
createdb validation_example
```
once, and then
```sh
ruby validation_example.rb # or however the file is called
```
To run the app and get into an interactive pry session

**But you don't have to do any of this**, I ran the app for you and illustrated some examples of how this type of validation works:
```ruby
[1] pry(main)> User.new.valid?
=> true
[2] pry(main)> user = User.new(terms_of_service: "1")
=> #<User:0x007f951902e9e0 id: nil, name: nil, created_at: nil, updated_at: nil>
[3] pry(main)> user.valid?
=> true
[4] pry(main)> user = User.new(terms_of_service: "0")
=> #<User:0x007f951813ed90 id: nil, name: nil, created_at: nil, updated_at: nil>
[5] pry(main)> user.valid?
=> false
[6] pry(main)> user.errors
=> #<ActiveModel::Errors:0x007f9518d0d298
 @base=#<User:0x007f951813ed90 id: nil, name: nil, created_at: nil, updated_at: nil>,
 @details={:terms_of_service=>[{:error=>:accepted}]},
 @messages={:terms_of_service=>["must be accepted"]}>
[7] pry(main)> user = User.new(terms_of_service: "2")
=> #<User:0x007f9518458df0 id: nil, name: nil, created_at: nil, updated_at: nil>
[8] pry(main)> user.valid?
=> false
[9] pry(main)> user.errors
=> #<ActiveModel::Errors:0x007f951826e8c8
 @base=#<User:0x007f9518458df0 id: nil, name: nil, created_at: nil, updated_at: nil>,
 @details={:terms_of_service=>[{:error=>:accepted}]},
 @messages={:terms_of_service=>["must be accepted"]}>
[10] pry(main)>
```

So as you see, Rails does all the work already for you! I would prefer to have Rails do the work in this case, instead of having to implement it ourself and testing it (we don't have to test Rails' internals, we'll just assume they work 😉 )

So the boolean fields just have to be present in the form and allowed in the controller. For the rest, Rails will take care of things.

Everything clear? Just ask me if this is still confusing

juuh42dias: The problem I see with this validation is that it will always require the terms to be accepted with or without selected conferences. This forces the user to accept the terms every time he 'edit' his team.

klappradla: Right, but I don't see a problem with this. You can still add things like "only on create", etc. to the validation. It's still a normal Rails-validation.

juuh42dias: @mkalininait  what do you think? Sounds good for you?

mkalininait: @juuh42dias I'm not sure I understood Max's solution @klappradla  "only on create" means on create of a team? It won't be very convenient in the cases when we want to edit something else but the conference preferences. E.g. AFAIU we always show the conference block on the team view/edit pages, regardless of the current phase (applications, coding summer etc.). It means that we will have to accept the conference terms when we edit a team during the application phase, right? Even if we fix this and enable the conference functionality only for the accepted teams (which we should probably do in a separate issue), there will be time when we don't have a list of conferences yet, but already demand to accept the terms the moment a team wants to add some team info (e.g. blog url). 
I love the idea that we check the terms only if a conference is selected. On the other hand, I understand why Max @klappradla proposes what he proposes. After all, if we make the terms validation independent of the conference preferences, it won't be the end of the world, if we hide this for applying teams (at least). What do you think @klappradla ?

klappradla: "only on create" related to the `ConferencePreference` record, not the `Team`. So it would be checked once, when the team first choses conferences.

app/helpers/teams_helper.rb
```diff
@@ -1,6 +1,11 @@
 module TeamsHelper
 
-  def conference_exists_for(team)
-    team&.conference_preference&.persisted? && team&.conference_preference&.first_conference_id.present? || team&.conference_preference&.second_conference_id.present?
+  def conference_exists_for?(team)
+    conference_preference = team.conference_preference
+    return false unless conference_preference.present? && conference_preference.persisted?
```
klappradla: Way easier to understand now I'd say 💪 
Maybe shorten this line to:
```ruby
return false unless conference_preference&.persisted?
```
But that's just personal preference, you can also leave it as it is - there's no change in behavior.

spec/helpers/teams_helper_spec.rb
```diff
@@ -8,11 +8,11 @@
     let(:team_with) { conference_preference.team }
 
     it 'should return false if don\'t exists conference persisted' do
```
klappradla: Escaping within strings always looks kinda weird if there's no "real" reason for it 😉 
I'd suggest to either use double quotes here, or to just rename the example to something like
> it 'returns false if no conference_preference was persisted for the team'

Combining "it" and "should" for an RSpec example is also a bit weird, since they are actually both keywords for the same...

app/models/conference_preference.rb
```diff
@@ -1,6 +1,5 @@
 class ConferencePreference < ActiveRecord::Base
-  validates :terms_accepted?, acceptance: true
-
+  validates :terms, acceptance: true
```
klappradla: With the type of validation you're applying now, I think this is no longer an "acceptance: true" validation. It is instead now a so called [custom method validation](http://guides.rubyonrails.org/active_record_validations.html#custom-methods).

Therefore, I'd also suggest to rename this, so that it makes more sense for someone not familiar with the domain looking at it for the first time. Maybe just like you had it before
```ruby
validates :terms_accepted?
```

app/models/conference_preference.rb
```diff
@@ -9,13 +8,11 @@ class ConferencePreference < ActiveRecord::Base
 
   private
 
-    def terms_accepted?
+    def terms
       conference = first_conference_id.present? || second_conference_id.present?
       terms_checkbox = terms_of_ticket == '1' && terms_of_travel == '1'
```
klappradla: Technically speaking this is not a "checkbox". How about
```ruby
terms_checked = ...
```
? Expresses more what's inside the variable.

app/models/conference_preference.rb
```diff
@@ -9,13 +8,11 @@ class ConferencePreference < ActiveRecord::Base
...
       first_case = conference.present? && terms_checkbox.present?
```
klappradla: Using "present?" for something that is already a boolean looks a bit confusing for me...
```ruby
first case = conference_chosen && terms_checked
```
would do the same, but more expressive, what do you think?

app/models/conference_preference.rb
```diff
@@ -9,13 +8,11 @@ class ConferencePreference < ActiveRecord::Base
...
       second_case = conference.blank? && terms_checkbox.blank?
-      errors.add(:team, "If you choose any conference you must accept the terms") if conference && !terms_checkbox
-      errors.add(:team, "If you accept terms you must choose any conference") if !conference && terms_checkbox
       first_case || second_case
```
klappradla: Can you come up with better names for this?

Maybe the whole thing as something along the lines of
```ruby
def terms_checked?
  conference_chosen = first_conference_id || second_conference_id
  terms_checked = terms_of_ticket == '1' && terms_of_travel == '1'
  (conference_chosen && terms_checked) || (!conference_chosen && !terms_checked)
end
```

mkalininait:
@klappradla How far (you think) are we from merging this PR?

klappradla:
I think it's basically good to merge @mkalininait 
I'll have another look in a few minutes on the newest changes, but it looked already pretty good, so that I'll probably merge right away ✌️

mkalininait:
Yay, thanks @klappradla @juuh42dias ! We can start the conference process now! (I hope :D)

mkalininait:
Closes https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/768 and https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/769

mkalininait:
@juuh42dias @klappradla Oops, I've just tested the feature and noticed that validation of the terms doesn't work anymore. I can select a conference and save the form without checking the terms. BUT the app does it for me: next time I open the edit form, I see the terms checked even though I didn't check them.
Could you please fix it? We can't start the conference process with this bug I think.

juuh42dias:
@mkalininait I'll see this :)

PR #811, opened 2017-08-07
Conferences/assigning a conference

branquinhoaa:
PR related with issue #771
- include a new model 'ConferenceAttendance'
- allow orga user to assing a conference for a team
- allow students/supervisors/admin to see the offered conference for the team
### Teams show (info just for students, supervisors and orga members)

teams_show

### Team edit (form just for orga members)

teams_edit


app/models/ability.rb
```diff
@@ -29,6 +29,10 @@ def initialize(user)
       user.admin? or (user.student? && (team.students.include? user))
     end
 
+    can :see_offered_conferences, Team do |team|
+      user.admin? or (team.students.include? user) or team.supervisors.include?(user)
```
klappradla: Super minor thing, but it's considered bad practice to use `or` or `and` as they may introduce bugs and do only add very little readability. It would also be nice if you could align the usage of brackets here, would be easier to grasp I think:
```ruby
user.admin? || team.students.include?(user) || team.supervisors.include?(user)
```

app/models/team.rb
```diff
@@ -182,6 +184,10 @@ def without_preferences?(att)
     att[:conference_id].blank?
   end
 
+  def without_conferences?(att)
```
klappradla: 👍 
Maybe use `attributes` instead of `att` - would make it more obvious on first glance (same for the `without_preferences` method).

In this case, given the size of the class, I would actually even suggest to use the direct `proc` syntax, otherwise one has to scroll to the bottom of the large file to understand what's going on... What do you think?
```ruby
accepts_nested_attributes_for :conference_attendances, allow_destroy: true, reject_if: proc { |attributes| attributes[:conference_id].blank? }
```

But totally up to you, does not change anything in behavior.

app/views/orga/teams/_form.html.slim
```diff
@@ -77,14 +77,26 @@
     = f.input :season_id, as: :select, collection: Season.order('name DESC') { |e| [s.name, s.id] }
 
   .page-header
-    h2 Conference preferences
+    h2 Conferences
+    h3 Conference preferences
```
klappradla: Since this is a heading, I'd suggest to capitalize it:
> Conference Preferences

_(same for the offers)_

app/views/orga/teams/show.html.slim
```diff
@@ -82,6 +84,17 @@ nav.btn-group
       = links_to_conferences([preference.conference]).first
       = preference_option(preference)
 
+- if can? :see_offered_conferences, @team
+
+  - if @team.conference_attendances.present?
+    h4 Conference offers
```
klappradla: As said before, I'd suggest to capitalize these headlines.

app/views/teams/_form.html.slim
```diff
@@ -71,7 +71,8 @@
         = f.link_to_add 'Add another source', :sources, class: 'btn btn-primary col-sm-offset-3'
   -if can?(:update_conference_preferences, @team)
     .page-header
-      h2 Conference preferences
+      h2 Conferences
+      h3 Conference preferences
```
klappradla: As said before, I'd suggest to capitalize these headlines.

app/views/orga/teams/show.html.slim
```diff
@@ -82,6 +84,17 @@ nav.btn-group
...
+    dl
+      - @team.conference_attendances.each do |offer|
+        dt Offered conference:
+        dd = offer&.conference&.name
```
klappradla: I'm a bit torn about the lonely operator here 🤔  You already check whether or not there's a `ConferenceAttendance` record and afaik Rails 5 automatically enforces `belongs_to` associations, so there will always be a conference associated plus the conference has a validation on the presence of its name... For the sake of simplicity I would leave the lonely operators here.

app/views/orga/teams/show.html.slim
```diff
@@ -82,6 +84,17 @@ nav.btn-group
...
+        - if offer.orga_comment.present?
+          dt Orga comment:
+          dd = offer&.orga_comment
```
klappradla: No need for the lonely operator here, you're already checking for the presence 2 lines above.

spec/factories/conference_attendances.rb
```diff
@@ -0,0 +1,8 @@
+FactoryGirl.define do
+  factory :conference_attendance do
+    attendance false
+    orga_comment "MyText"
```
klappradla: Probably mostly personal preference, but I found it easier on the long run, to keep factories as simple as possible. Meaning, if the `orga_comment` is an optional attribute, I'd just fully skip it in the factory.

spec/models/ability_spec.rb
```diff
@@ -171,6 +171,35 @@
         end
       end
 
+      describe "just orga members, team\'s supervisor and team\'s students should be able to see offered conference for a team" do
```
klappradla: No need for escaping `'`s. They can be used within `".."`.

spec/models/ability_spec.rb
```diff
@@ -171,6 +171,35 @@
...
+        let(:user) { FactoryGirl.create(:student)}
```
klappradla: Is it really necessary to `create` the user here? Wouldn't `build` be sufficient?

spec/models/ability_spec.rb
```diff
@@ -171,6 +171,35 @@
...
+        end
+
+        context 'if it is a supervisor of another team' do
+           before do
+            allow(user).to receive(:supervisor?).and_return(true)
+          end
+          it { expect(ability).not_to be_able_to(:see_offered_conferences, Team.new()) }
+        end
+
+        context 'if it is a team\'s student' do
```
klappradla: Please use double quotes and don't escape the single quote then _(also in the other examples of course)_

spec/models/conference_attendance_spec.rb
```diff
@@ -0,0 +1,6 @@
+require 'spec_helper'
+
+RSpec.describe ConferenceAttendance do
```
klappradla: Only stylistic thing, but you can spare the `RSpec.` part.

spec/models/ability_spec.rb
```diff
@@ -171,6 +171,35 @@
         end
       end
 
+      describe "just orga members, team\'s supervisor and team\'s students should be able to see offered conference for a team" do
+        let(:user) { FactoryGirl.create(:student)}
+
+        context 'if it is an student of another team' do
```
klappradla: For the sake of readability, I would try to come up with something starting with either _"when"_ or "with"_ for the `context` hooks - it really helps understanding what's going on. Also: using "it" is a bit ambiguous here...

How about something like:
```ruby
context "when the user is a student on another team" do
  # do stuff
```

_(this also goes for the other contexts you define here)_

app/views/orga/teams/_form.html.slim
```diff
@@ -77,14 +77,26 @@
...
 
   fieldset.conferences_fieldset
     = f.simple_fields_for :conference_preferences, f.object.conference_preferences.sort{ |a,b| a.option <=> b.option} do |c|
       = c.hidden_field :option, value: c.object.option
       = c.input :conference_id, collection: @conferences.group_by(&:region), as: :grouped_select, group_method: :last, label: c.object.option == 1 ? "Primary choice" : "Secondary Choice", prompt: "Conference", wrapper_html: { class: 'col-md-8' }
 
+  .page-header
+    h3 Conference offers
```
klappradla: It would maybe make sense to extract this into a partial if it appears in exactly the same manner in two views. Would also take away the overhead of keeping them in sync... 😉

app/views/orga/teams/show.html.slim
branquinhoaa: yeah, that is true. I forgot to remove it. I just put it because before add a reject_if I had an offer without conference that was giving me some errors.

app/views/orga/teams/_form.html.slim
branquinhoaa: Are you referring to conference preferences? because the conference offer just appears in this specific view. But that is true, a lot better to share partials. I will do it.

branquinhoaa:
@klappradla What do you think now?

mkalininait:
@branquinhoaa Wow, it looks very good to me!
Two comments:
1) I'd rename the blank option to "None" to make it consistent with the Conference preferences.
2) I'd make the formatting of the offered conferences the same as of the preferred conferences: with links and (maybe) bullets.

screen shot 2017-08-10 at 20 18 49

What do you think?

app/views/teams/_conference_edit.html.slim
```diff
@@ -0,0 +1,28 @@
+.page-header
```
klappradla: 👍  Can you please rename the partial to `_conference_preference_fields`? It's otherwise confused with the teams form or edit view.

app/views/teams/show.html.slim
```diff
@@ -86,30 +86,7 @@ nav.page
     p = @team.supervisors.any? ? link_to_team_members(@team, :supervisor) : content_tag(:em, 'Will be added by RGSoC Organizers')
 
   h3 Conferences
```
klappradla: Nice! I'd suggest to put the h3 also into the partial, so that you just have:
```ruby
= render 'conference_info'
```

app/views/teams/show.html.slim
```diff
@@ -86,30 +86,7 @@ nav.page
...
-  - if @team.conference_attendances.present?
-    h4 Conference offers
-    dl
-      - @team.conference_attendances.each do |offer|
-        dt Offered conference:
-        dd = offer&.conference&.name
-        - if offer.orga_comment.present?
-          dt Orga comment:
-          dd = offer&.orga_comment
+  == render 'conference_info'
```
klappradla: This should be just one `=` (not sure if == actually outputs something.... 🤔 )

spec/factories/conference_attendances.rb
```diff
@@ -1,7 +1,6 @@
 FactoryGirl.define do
   factory :conference_attendance do
     attendance false
-    orga_comment "MyText"
```
klappradla: 👍

mkalininait:
@branquinhoaa The first screenshot looks good! <3 (`Orga comment: <comment>`)

branquinhoaa:
@mkalininait @klappradla fixed!

app/views/orga/teams/_form.html.slim
```diff
@@ -76,32 +76,19 @@
   - if admin?
     = f.input :season_id, as: :select, collection: Season.order('name DESC') { |e| [s.name, s.id] }
 
+    == render 'teams/conference_preference_fields', f: f
```
klappradla: Can you please also make this a single `=` _(even though it maybe accidentally works, it just does not look semantically correct)_

db/schema.rb
```diff
@@ -10,6 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
+
```
klappradla: Was this blank line automatically added?

klappradla:
🎉

mkalininait:
@branquinhoaa I'd add a space between the 'Orga comment:' and an actual comment ;)
But since the PR is already merged, maybe next time :)

**Screenshot**
screen shot 2017-08-14 at 23 42 42

PR #812, opened 2017-08-09
Fix validation scenario - Edit teams with/out Conferences

juuh42dias:
With 'none' option:

rgsoc teams app

### DIFF
--- a/app/models/conference_preference.rb
+++ b/app/models/conference_preference.rb
@@ -1,5 +1,5 @@
 class ConferencePreference < ActiveRecord::Base
-  validate :terms_checked?
+  validates :terms_checked?, acceptance: true
   attr_accessor :terms_of_ticket, :terms_of_travel
 
   belongs_to :team, inverse_of: :conference_preference
@@ -9,7 +9,7 @@ class ConferencePreference < ActiveRecord::Base
   private
 
   def terms_checked?
-    conference_chosen = first_conference_id || second_conference_id
+    conference_chosen = first_conference_id.present? || second_conference_id.present?
     terms_checked = terms_of_ticket == '1' && terms_of_travel == '1'
     (conference_chosen && terms_checked) || (!conference_chosen && !terms_checked)
   end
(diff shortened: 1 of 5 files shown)


klappradla:
This is supposed to fix #806. It looks ok codewise, but I'm not sure if the behavior is now **exactly** what's desired - maybe @mkalininait you have another look ✌️

mkalininait:
Hey @klappradla This PR is supposed to fix two things:
1) Rename the NONE option to NONE https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/806
2) Fix validation of the terms https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/810#issuecomment-321212774

@juuh42dias 
1) looks good to me!
2) works partially. Unfortunately the validation works only when we create a preference (which we can see from the code as well: `validates :terms_checked?, acceptance: true, on: :create`). If I uncheck a term checkbox after a preference has been created and save the form, the app still saves 'true'.

The whole term validation feature looks simple to me:
- If at least one preference is selected, we check if the terms are checked when we save the form:
  - If the terms are not checked, we show an error message.
  - If the terms are checked, we save the changes.
- If both preferences are not selected (set to 'None'), we don't check the terms at all.

Do you think we can make it work this way? @juuh42dias @klappradla

klappradla:
Thanx for clearifying again @mkalininait 💚 

This is all getting way too complex, I'm so sorry @juuh42dias 🙈 

Let's just step back and build it the most basic way possible: with real DB columns

1. Add two new columns to the `ConferencePreference` model:
    ```ruby
    terms_of_ticket: :boolean, default: false
    terms_of_travel: :boolean, default: false
    ```
1. Add the two checkboxes to the form and allow them in the controller
1. Add a simple validation of these boolean fields to the model
    ```ruby
    validates :terms_of_ticket, inclusion: { in: [true] }, unless: Proc.new { |a| a.first_conference_id.blank? && a.second_conference_id.blank? }
    validates :terms_of_travel, inclusion: { in: [true] }, unless: Proc.new { |a| a.first_conference_id.blank? && a.second_conference_id.blank? }
    ```
    _(you can also extract the proc into a method - my brain is just to melted to write this outta my head)_

I think this should finally give us the desired behavior and it's still understandable ✌️

klappradla:
Awesome 🥇

mkalininait:
@juuh42dias @klappradla Perfect! Thank you so much! <3

Issue #770, opened 2017-06-24
[Conferences] #6: Export of students' preferences


mkalininait:
The idea: https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/751#issuecomment-308574743
Full list of conference user stories: https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/751#issuecomment-309138513

**User Story:**
As an RGSoC orga, I want to export teams' conference preferences as a csv.

**Things to do:**
1. Permissions:
 - Orga
2. Format: csv
3. Available from:
 - `/orga/exports`
 - `/orga/conferences` (see the comment below)
4. Exported data (only selected teams of the current season):
 - Team name
 - Team location (optional)
 - Project name (optional)
 - Project technologies (optional) // skipping it for the moment, see https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/813#issuecomment-322189616
 - Conference primary choice
 - Conference secondary choice
 - We would like to give a LT (see #768)
 - Comments (see #768)
 - Terms accepted (YES if both checkboxes in #769 are checked)

If the optional fields are hard to retrieve, we can skip them.

branquinhoaa:
@mkalininait this export button can be in the conference/index, like this:
![export_btn](https://user-images.githubusercontent.com/15239141/28574103-3c5f8a2a-7123-11e7-9b93-65c8dc448c06.png)

mkalininait:
@branquinhoaa Good suggestion! And we should also add it here: `/orga/exports`

mkalininait:
@branquinhoaa I've updated the description of this task ☝️

branquinhoaa:
@mkalininait What about the additional info about chosen conferences #768? Will it be exported too?

mkalininait:
@branquinhoaa Goooood point, thank you! Yes, it will. I've updated the task description.

branquinhoaa:
Hi, @mkalininait! 
this one is done, but considering that the PR #810 change the conference preference model structure, we will need to wait for it to be merged (I get all the export data from Conference preference new model, so I need it).

There is just one field that I couldn't find: project technologies. Each team has a column named project_name, but no fields or relation with other models that have the project technology info.

Do you know how can I get this info?

mkalininait:
Hi @branquinhoaa Great news!

Sorry, I was not accurate enough. By project technologies I meant project tags (the "Tags" column in the "Projects" table).

Issue #19, opened 2013-06-21
* nice-to-have: List overview of all the involved projects with links to the participating teams


ellenkoenig:

amrtanair:
When you say "links to the participating teams", do you mean the ones that got selected for that project or the ones that have applied for it?  @carpodaster @ellenkoenig

alicetragedy:
@amrtanair the ones that got selected to participate.

F3PiX:
Surprise! This one is done in #791 .
There is also a follow-up issue #784. 
I think we can close this? (@carpodaster)

alicetragedy:
@F3PiX I'm not sure about this — where are the participating teams shown on the project list? I still think it would be great to see the teams + associated projects somehow, not sure if the project page is the best place for it though. wdyt?

F3PiX:
Oops, good catch! That was done in my dreams....  ;-) I'd really like to add a column with a link to the team working on the project, preferably their activity stream.
When we are showing the 'active projects' (= accepted and with a team working on it), I'd really like to see what team is associated - preferably with a link to their activity stream. I'd say that is the most useful team information from the 'projects page point of view'. 
 See #784 about the Project Page.

alicetragedy:
@F3PiX haha :D Yeah, I agree with you — adding that column with a link to the team working on the project makes a lot of sense. Would love to see this happen <3

alicetragedy:
@F3PiX haha :D Yeah, I agree with you — adding that column with a link to the team working on the project makes a lot of sense. Would love to see this happen <3

F3PiX:
Okay, let's do it!

mkalininait:
It would be also great to show the team's project on the team's page.
I wonder if it will require changing the way we assign projects to teams and the team model.

F3PiX:
I'll close this one, there is a follow-up issue, #784  where we can collect the discussions. Feel free to reopen if there is something that needs discussing.

PR #813, opened 2017-08-09
Conferences/export preferences

branquinhoaa:
This PR is related to the issue #770 - Export of students' preferences 

- include an export file for student's preference
- include a test file for this export file
- include a button in orga/conference for orga members download the csv

exportsconf
orgaconf

### DIFF
--- a/app/models/conference_preference.rb
+++ b/app/models/conference_preference.rb
@@ -7,8 +7,16 @@ class ConferencePreference < ActiveRecord::Base
   belongs_to :first_conference, class_name: 'Conference'
   belongs_to :second_conference, class_name: 'Conference'
 
+  scope :current_teams, -> {
+    joins(:team).where(teams: { season_id: Season.current.id })
+  }
+
   def has_preference?
     first_conference.present? || second_conference.present?
+  end
+
+  def terms_accepted?
+    self.terms_of_travel && self.terms_of_ticket
   end
 
   private
--- a/spec/exporters/conference_preferences_spec.rb
+++ b/spec/exporters/conference_preferences_spec.rb
@@ -1,0 +1,16 @@
+require 'spec_helper'
+
+RSpec.describe Exporters::ConferencePreferences do
+
+  describe '#current' do
+    let!(:old_team) { create :team, :with_preferences, name: "OLDTEAM" }
+    let!(:new_team) { create :team, :with_preferences, :in_current_season, name: "NEWTEAM" }
+
+    subject { described_class.current }
+
+    it 'exports conference preferences for teams of the current season' do
+      expect(subject).to match 'NEWTEAM'
+      expect(subject).not_to match 'OLDTEAM'
+    end
+  end
+end
--- a/app/exporters/conference_preferences.rb
+++ b/app/exporters/conference_preferences.rb
@@ -1,0 +1,12 @@
+module Exporters
+  class ConferencePreferences < Base
+
+    def current
+      preferences = ConferencePreference.current_teams
+
+      generate(preferences, 'Team name', 'Team location', 'Project name', 'Conference primary choice', 'Conference secondary choice', 'We would like to give a LT', 'Comments', 'Terms accepted') do |cp|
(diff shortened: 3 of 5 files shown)


branquinhoaa:
@mkalininait I think it is it! :+1:

mkalininait:
@branquinhoaa Couldn't test it because of an error.

screen shot 2017-08-10 at 20 07 34

mkalininait:
@branquinhoaa The export works now!
 - I think Project Tags are not being exported.
 - Please rename the button "Download team's preferences" to "Download teams' preferences".
 - Will the last exported column "Terms accepted" might have changed after this PR: https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/812 Could you please check this?

mkalininait:
@branquinhoaa UPD:
I've tested the column "Terms accepted". It looks like it always exports TRUE, even if the terms are not accepted.

branquinhoaa:
Yeah, that is because all the confirmation was being made without persisting the data, so we were assuming that if a team had a conference preference it had put true in the fields. Now that we are persisting those fields, we can verify this better. I will change it! :+1:

branquinhoaa:
@mkalininait Could you please test now? :+1:

mkalininait:
@branquinhoaa 
The button is fixed.
The tags are still not being exported.
The "Terms accepted" column I can't test, because I can't reset conference preferences anymore. When I try to set None in both preferences and save the form, the app doesn't save these changes and keeps old preferences. I guess it's a bug? How was it supposed to work?

branquinhoaa:
- The terms are fixed. Now you can set 'none' again.
- About the tags, I have to figure out what is happening.  
Locally, when I set the associations (teams with applications with projects with tags) using seeds it exports the tags normally.
Maybe, if you could provide me the production DB backup I could try to troubleshoot the issue locally. 
Alternatively, someone could confirm that the production teams have an application associated and projects(with tags) associated too. 
Like this: team.application.project.tags

mkalininait:
We decided not to export Project Tags for the moment. Teams are not associated with Projects atm; there is `project_name` (string) in the team model, but no `project_id`. Unfortunately, we can't get the info about a project associated with a team from it's application.
Maybe the Team-Project relation will be set to resolve this issue: https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/19

branquinhoaa:
Now I think it is working properly! :dancing_women:

branquinhoaa:
@mkalininait @klappradla I think we will need this one to be merged before the #772.

mkalininait:
@branquinhoaa Looks great! Green light from my side //cc @klappradla

app/models/conference_preference.rb
```diff
@@ -7,10 +7,18 @@ class ConferencePreference < ActiveRecord::Base
   belongs_to :first_conference, class_name: 'Conference'
   belongs_to :second_conference, class_name: 'Conference'
 
+  scope :current_teams, -> {
```
klappradla: It would be cool, if you could add a test for this scope in the `ConferencePreference` model specs.

spec/exporters/conference_preferences_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+
+RSpec.describe Exporters::ConferencePreferences do
+
+  describe '#current' do
+    let!(:old_team) { create :team, :with_preferences, name: "OLDTEAM" }
+    let!(:new_team) { create :team, :with_preferences, :in_current_season, name: "NEWTEAM" }
+
+    subject { described_class.current }
+
+    it 'exports conference preferences for teams of the current season' do
```
klappradla: Hm, I feel like this is under the hood just testing the `ConferencePreference`'s `current_team` scope, not actually the exporter in itself.

Can you come up with a test that better targets the exporter?

branquinhoaa: Nice! I will work on that.

klappradla: Cool! #823 is the dedicated issue. It is not super pressing, but since tests are equally part of the app as "normal" application code, they deserve the same amount of thoughts and love 🦄 

Feel free to ping me if anything is unclear about the issue.

mkalininait:
Yay!

Issue #502, opened 2016-07-15
Past Teams


inescoelho:
- [ ] It would be nice if we were able to browse past teams' activity. Some of this information is still available when people is connected with past teams (like me, since I was a student last year and now I'm a supervisor).
- [x] Moreover, the team profile presents the starting date and finishing date of the team. However, this information only displays day and month. Now that we are in the 4th edition of RGSoC, it get's confusing... It would be useful to add the year as well.

F3PiX:
I like the idea of browsing in the history. I would prefer to set up the design for the majority of users, for who the current year matters only. And thus, adding the year 2016 is a bit of noise. Maybe we can add something like a select box to filter activities by year and/or previous team names and what not?

inescoelho:
Hi @F3PiX! When you have someone involved in several teams like [this](https://teams.railsgirlssummerofcode.org/users/474), it's kinda hard to figure out where those teams come from. After opening each profile, you have to scroll down to their activity to figure out from each edition they are and it would be nice to have something more immediate.
The browsing on history feature could be useful to show to the world what is our program and what it has been done throughout the years. This could easily be done with a selection box on the Teams page, where you select the year and the table would display the teams of that year. ;)

F3PiX:
Haha, great profile page. 
I see what you mean about adding the year to the start and end date on the Team's profile page. I do think that is a good idea. 
And sure, browsing the history in one way or another can be useful. I am not sure if the best way is to add a filter to the teams index page. I think the community page is a better place for that? There is not much room for an extra column though. Apart from that, how do you think about that?

carpodaster:
Hey @inescoelho. I agree that it would be nice to make past teams more visible, but I also agree with @F3PiX that our main focus is always on the current season and it's therefore not super-high on our priority list. Would you maybe like to work on it yourself and send a PR? 😄 

You raise a valid point with the date format: that should be fixed. I took the liberty of editing your issue to create a checkable bullet list.

branquinhoaa:
Hello! what do you people think about this layout for this issue?
It is /teams index, as suggested by Ines.
![past_teams](https://user-images.githubusercontent.com/15239141/29176144-66920bba-7dc1-11e7-9bed-2f8465f3c2cf.png)

klappradla:
Closing this via #814

PR #814, opened 2017-08-11
Teams/add past teams

juuh42dias:
This PR is related to the issue  #502
- Include select dropdown to filter teams by year

## Select teams by year
screenshot from 2017-08-11 17-12-23

## Select teams by year and order by activities
screenshot from 2017-08-11 17-12-35


app/controllers/teams_controller.rb
```diff
@@ -106,4 +101,23 @@ def set_display_roles
       @display_roles = ['student']
       @display_roles.map!(&:pluralize)
     end
+
+    def show_teams_by(params)
```
carpodaster: `params` is available as a controller method, you don't need to pass it as an argument. With the argument gone, I would go for a more generic method name, something like `find_teams`

It is also a veeeery long method.

I'm wondering why we only include the activities (w/ sorting) when there is a `:sort` param. I would think this way of sorting / eager loading makes sense whether or not we sort stuff. Do you know, @klappradla or @alicetragedy ?

Provided we allow combining all these options, we can just build up the query instead of having a complex `case` or `if … else` statement:

```ruby
direction = params[:direction] == 'asc' ? 'ASC' : 'DESC'
# start with the "base scope"
@teams = Team.by_season_phase.includes(:activities).order("teams.kind, activities.created_at #{direction}").references(:activities)
# Re-assign @teams if conditions apply:
@teams = @teams.by_season if foo
@teams = @teams.by_something_else if bar
```

Does that make sense?

app/models/team.rb
```diff
@@ -63,10 +63,21 @@ def in_current_season
       where(season: Season.current)
     end
 
+    def by_season_year(year)
+      where(season: Season.where(name: year))
+    end
+
+    def select_teams_by_season_year(year)
```
carpodaster: This looks like a highly specialised method. Since the `by_season_year` returns an active relation, you could just add `.selected` after that. Let us go that route and remove this method.

Or if that doesn't work, try rephrasing the `selected` method as a `scope`.

app/models/team.rb
```diff
@@ -63,10 +63,21 @@ def in_current_season
...
+      by_season_year(year).selected
+    end
+
     def selected
       where(kind: %w(sponsored voluntary))
     end
 
+    def activities_ordered_by(direction)
```
carpodaster: This, too, seems like a highly specialised method and to me it doesn't feel like ordering the activities is part of the `Team` class' responsibility. I would move it back into controller.

app/models/team.rb
```diff
@@ -63,10 +63,21 @@ def in_current_season
       where(season: Season.current)
     end
 
+    def by_season_year(year)
```
carpodaster: Idea: making this `by_season(year_or_season)` that automatically detects whether the argument is a real `Season` object or a year (either as an integer or a string) and then constructs the `where` condition accordingly.

app/assets/javascripts/past_teams.js
```diff
@@ -0,0 +1,16 @@
+$(function() {
+  var $selectedYear = $('select#past_teams');
+
+  $selectedYear.on('change', function(){
+    var yearParam = "?year="+this.value;
+    var newUrl;
+    var removed = window.location.search;
+
+    if(removed.length>0){
```
carpodaster: I cannot wrap my head around how that would behave in the browser I'm afraid. Can someone else check that pls? 😁

klappradla:
Hey @juuh42dias looking at your screenshot I just realized you're using a native dropdown element _(looks a bit bootstrappy on Linux on first sight)_

However, in Chrome on Mac, this would end up looking like this:
screen shot 2017-08-13 at 17 19 53


The Bootstrap - and probably preferred way - would be this:
screen shot 2017-08-13 at 17 25 51

The changes for this would just be:
```ruby
= select_tag 'past_teams', options_for_select(Season.all_years), prompt: 'Past Teams', class: 'form-control'
```

However, since you'r not adding a form on this page, I think the element you're looking for is actually [a button with a dropdown](https://getbootstrap.com/docs/4.0/components/dropdowns/).

app/assets/javascripts/past_teams.js
klappradla: What it does: takes the current URL, removes the search params from it, and then appends new ones, e.g. `?year="2016"` to it.

It then changes the current URL which makes the browser do another get request with the new params.

Afaik @juuh42dias, you could shorten this to just:
```javascript
$(function() {
  var $selectedYear = $('select#past_teams');

  $selectedYear.on('change', function(){
    var yearParam = "?year="+this.value;
    window.location.search = yearParam;
  });
});
```
It should to exactly the same.

app/assets/javascripts/past_teams.js
```diff
@@ -0,0 +1,16 @@
+$(function() {
+  var $selectedYear = $('select#past_teams');
```
klappradla: This adds a lot of non-default behavior to this select element. My personal guideline for this is usually: if I use Javascript to add behavior to an element, use data attributes, in this case something like`data-behavior="switch-teams-year"` and leave classes and ids fully for anything related to CSS and styling.

I even added some JS code to the Teams App once where I use it. Have a look at [city.js](https://github.com/rails-girls-summer-of-code/rgsoc-teams/blob/master/app/assets/javascripts/city_fields.js) and the view files associated with it for inspiration.

Aside from that: I think it's sematically speaking slightly wrong to use a `select` tag here, since we're not in a form... What do you think?

app/controllers/teams_controller.rb
klappradla: 👍 sure, the eager loading should be independent of whether there's a `:sort` param or not - the `last-activity` timestamp is still displayed. Don't see the reason why it should be different @carpodaster.

Why we need to eager load all activities to then in the end just pluck a single timestamp per team is a different thing though 😉

app/assets/javascripts/past_teams.js
branquinhoaa: Really nice using data attributes for it!! 
I looked in your city.js but I couldn't find any view associated with it. I searched in the files and I put a breakpoint in your function to see the element. Is it used to auto-complete cities' names in forms? 
Anyway, I will replace the id by the data attribute.

klappradla: Yes, you're right, it's used by this type of input element: [here's the code for it](https://github.com/rails-girls-summer-of-code/rgsoc-teams/blob/master/app/inputs/city_input.rb)

It's not 100% the same, but the usage in views is similar in syntax.

juuh42dias:
@klappradla  @carpodaster  that's ok now?

app/controllers/teams_controller.rb
```diff
@@ -6,11 +6,19 @@ class TeamsController < ApplicationController
   load_and_authorize_resource except: [:index, :show]
 
   def index
```
klappradla: This looks a bit confusing to me 🙈 
What's the point in 3 times overriding the `teams` instance variable`? This is going to end up in multiple - redundant after all - calls to the database, which is usually not what you want. Can you think of a way to structure this nicer and get rid of redundant DB calls?

app/models/team.rb
```diff
@@ -43,6 +43,10 @@ class Team < ActiveRecord::Base
 
   scope :accepted, -> { where(kind: %w(sponsored voluntary)) }
 
+  scope :by_season_year, ->(year) { joins(:season).where('seasons.name': year) }
+
+  scope :with_last_activities_ordered_by, ->(direction) { includes(:activities).order("teams.kind, activities.created_at #{direction}").references(:activities) }
```
klappradla: I don't see this scope being used anywhere _(nor is it tested)_...

app/models/team.rb
```diff
@@ -70,6 +74,18 @@ def selected
       where(kind: %w(sponsored voluntary))
     end
 
+    def by_season(year_or_season)
```
klappradla: For what reason was this introduced? I don't see it being used anywhere 🤔

spec/controllers/teams_controller_spec.rb
```diff
@@ -41,6 +41,13 @@
       end
     end
 
+    context 'past teams' do
+      it 'passing year' do
```
klappradla: Hm, I'm afraid this does not say anything... Aside from the description not really telling the reader what's supposed to happen, the test only tests if the `index` action renders the `index.html` template. One can pass arbitrary params to an `index` action. The test would be the same with:
```ruby
get :index, params: { foo: 'bar'}
```

Given that the `index` action is rather complex, this test should probably better target the actual behavior there.

app/assets/javascripts/conferences_options.js
```diff
@@ -1,20 +0,0 @@
-$(function() {
```
juuh42dias: We don't use this `conferences_options.js` anymore.

app/controllers/teams_controller.rb
juuh42dias: I accept suggestions for refactoring this function.

My brain is blocked for now. :crying_cat_face:

app/assets/javascripts/conferences_options.js
klappradla: Ah ok, so it was just forgotten to be removed in another PR?

app/models/team.rb
juuh42dias: It was a Carsten suggest.

app/assets/javascripts/conferences_options.js
juuh42dias: Unfortunately :disappointed:

klappradla: Oh, alright, thanx for clearifying 🙏

app/models/team.rb
klappradla: Ok, I see. But what usecase did he intend it for?

Maybe I'm getting things wrong here, but I don't see the point in adding new _unused_ methods - that's like introducing legacy on purpose 😉

app/controllers/teams_controller.rb
klappradla: Oh, no worries 💚 

First of all I'd try to get rid of redundant queries. So a super simple thing here could for instance be to move the conditional:
```ruby
def index
  # stuff
  if year.present?
    @teams = query1
  else
    @teams = query2
  end
end
```

Then move the two queries from the second assignment into one:
```ruby
@teams = Team
  .by_season(year)
  .accepted
  .includes(:activities)
  .order("teams.kind, activities.created_at #{direction}")
  .references(:activities)
```

Can you use this as a starting point?

juuh42dias: @klappradla done :)

spec/controllers/teams_controller_spec.rb
```diff
@@ -41,6 +41,24 @@
...
+    context 'past teams' do
+      it 'return index by season year' do
+        get :index, params: { year: '2016'}
+        expect(response).to render_template 'index'
+      end
+
+      it 'return teams from current season' do
+        create_list :team, 2, season: Season.current
+        get :index, params: { year: Season.current.name }
+        expect(assigns(:teams).count).to eq 2
```
klappradla: How about making this even more strict?
```ruby
teams = create_list :team, 2, season: Season.current
expect(assigns(:teams)).to match_array(teams)
```

spec/controllers/teams_controller_spec.rb
```diff
@@ -41,6 +41,24 @@
...
+      end
+
+      it 'return empty list when the season is empty of teams' do
+        get :index, params: { year: '2000' }
+        expect(assigns(:teams)).to match_array []
```
klappradla: Hm, not sure if this test is really necessary. Since there aren't any `Team` records in the database, this would be positive for basically any query on the `team` table. I would suggest to either completely skip it, or add some records to really prove this one case.

spec/models/team_spec.rb
```diff
@@ -211,6 +211,21 @@
       end
     end
 
+    describe '.by_season' do
```
klappradla: I would suggest to make this a bit more explicit:
```ruby
let(:season) { create :season, name: '2016' }
let(:teams2016) { create_list :team, 2, season: season }

before { # create a team in different season than 2016 }

it "does something" do
  expect(Team.by_season('2016')).to match_array(teams2016))
end
```

What do you think?

spec/controllers/teams_controller_spec.rb
klappradla: Any news on what I wrote ☝️ @juuh42dias ?

One option could also be: move setting up the teams into a `let` macro
```ruby
let!(:teams) { create_list :team, 2, season: Season.current }
```

Doing so, the test around year 2000 is a bit more meaningful - what do you think?

klappradla:
🚢

PR #815, opened 2017-08-12
Update contribution guide with the ping method

klappradla:
As discussed on Slack, I added "the ping method" to the contribution guide and removed the `needs-review` label as it seems to be obsolete now that Github has the PR review request feature. I'd of course once this is agreed upon also remove the label from the repo ✌️ 
_(there only one old issue with this label at the moment)_


F3PiX:
👏  Nicely put, Max. +1 for removing the label.

carpodaster:
Top notch!

PR #816, opened 2017-08-14
Add team to projects page

F3PiX:
Solves #19 

- Restore the association between team and the project they are working on
- Add link to the team activities page in the Projects table 
- Create rake task for populating the `team.project_id`
  
Follow up: 
- [ ] After the records are updated with `project_id`,  the project_name can be removed from the table. 

Some background notes:
The team-project association has been removed in https://github.com/rails-girls-summer-of-code/rgsoc-teams/commit/7bcef84b42c13858a4cb106a8cab22f42dab6a59 as part of #354, the introduction of Project Proposals.

And even more background:
At that point an `application` had both a `project_id` and a `team_id`. 
Later on, the `project_id` in the application got replaced by the `project1_id` and `project2_id`, and stored in the [application_data hstore. ](https://github.com/rails-girls-summer-of-code/rgsoc-teams/blame/2e1936722aece933f03b7457a29f2d46f85c119a/app/models/application_data.rb)


app/views/projects/index.html.slim
```diff
@@ -28,6 +29,10 @@ table.table.table-striped.table-bordered.table-condensed.hidden-sm.hidden-xs
               = icon('minus')
       td = project_tags project
       td
+        - if project.teams.any?
```
carpodaster: You can remove the `if` – teams will always return a list. If it's empty, `#each` just won't do anything

lib/tasks/team_project.rake
```diff
@@ -0,0 +1,14 @@
+namespace :team_project do
```
carpodaster: Should we move this to `single_run.rake`? I would think we'd only run it once and can forget about it after that.

lib/tasks/team_project.rake
```diff
@@ -0,0 +1,14 @@
...
+  desc 'Change project_name to project_id'
+  task add_project_ids: :environment do
+    Team.all.each do |team|
+      next team unless team.project_name.present?
+      begin
+        team.project = Project.find_by(name: team.project_name)
```
carpodaster: We also need to take the season of both the team and the project into account

app/views/projects/index.html.slim
```diff
@@ -28,6 +29,10 @@ table.table.table-striped.table-bordered.table-condensed.hidden-sm.hidden-xs
...
+          - project.teams.each do |team|
+            = link_to team.name, activities_url(team_id: team.id)
```
carpodaster: I was about to write "let's add `Team#display_name` here" since that adds the season. But it also adds the project name. Do we need the season name?

app/models/project.rb
```diff
@@ -6,6 +6,7 @@ class Project < ActiveRecord::Base
 
   has_many :first_choice_application_drafts,  class_name: 'ApplicationDraft', foreign_key: 'project1_id'
   has_many :second_choice_application_drafts, class_name: 'ApplicationDraft', foreign_key: 'project2_id'
+  has_many :teams # working on the project; after application procedure
```
carpodaster: Should we name the association `accepted_teams` then to make it extra clear?

app/views/projects/index.html.slim
F3PiX: We only list [projects for the current season](https://github.com/rails-girls-summer-of-code/rgsoc-teams/blob/master/app/controllers/projects_controller.rb#L26)
So, adding the season name would seem redundant to me. Okay?
(Extra: I'd like to add the season name to the title though; that is part of [another issue](https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/784#issuecomment-316998327).)

app/models/project.rb
```diff
@@ -6,7 +6,7 @@ class Project < ActiveRecord::Base
 
   has_many :first_choice_application_drafts,  class_name: 'ApplicationDraft', foreign_key: 'project1_id'
   has_many :second_choice_application_drafts, class_name: 'ApplicationDraft', foreign_key: 'project2_id'
-  has_many :teams # working on the project; after application procedure
+  has_many :accepted_teams, class_name: 'Team' # working on the project; after application procedure
```
F3PiX: The good news: with the name `accepted_teams`, the comment is much more clear. But I feel it still needs the comment. For me, 'accepted_teams' is about the application procedure, less about 'working on the project'. 
I considered `:performing_teams`. ?

F3PiX:
Thank you @carpodaster. 🌷 Back to you. 

I realised there was a problem with the rake task, so I changed that too. 
The most important error seems to be if a team has an incorrect or outdated `project_name`. So I changed the rake task to found and log those errors.

lib/tasks/single_run.rake
```diff
@@ -65,4 +65,17 @@ namespace :single_run do
       end
     end
   end
+
+  desc '2017-08-15: Change project_name to project_id'
+  task add_project_id_to_team: :environment do
+    Team.all.each do |team|
```
carpodaster:
```ruby
Team.where(project_name: [nil, '']).each do |team|
  # no `next` needed
  # …
end
```

lib/tasks/single_run.rake
```diff
@@ -65,4 +65,17 @@ namespace :single_run do
...
+      next team unless team.project_name.present?
+      begin
+        team.project = Project.where(season_id: team.season_id).find_by!(name: team.project_name)
```
carpodaster: Find by lowercase names? Gives our spelling of project names a little more slack

app/models/project.rb
```diff
@@ -6,6 +6,7 @@ class Project < ActiveRecord::Base
 
   has_many :first_choice_application_drafts,  class_name: 'ApplicationDraft', foreign_key: 'project1_id'
   has_many :second_choice_application_drafts, class_name: 'ApplicationDraft', foreign_key: 'project2_id'
+  has_many :accepted_teams, class_name: 'Team' # working on the project; after application procedure
```
carpodaster: I agree, `accepted_teams` doesn't sound right. I'm not too happy with `performing_teams` (sounds a bit like a dance class :wink: ). What about `participating_teams` ?

carpodaster:
Back to you @F3PiX :)

lib/tasks/single_run.rake
F3PiX: 😍

app/models/project.rb
F3PiX: I'd vote for `accepted` over `participating`. 
New bid: `assigned_teams`. That doesn't cover the 'legal' relationship, but it communicates the bond between team and project. Okay?

carpodaster: 👍

lib/tasks/single_run.rake
F3PiX: Actually, I'd rather not. It would require to downcase both the `team.project_name` and each `project.name`.
I'd say the risk of this particular misspelling is very low, compared to all possible misspellings. Either way, problems will pop up in the console and will not go unnoticed.
Is it okay with you if I don't downcase?

F3PiX:
Back to you @carpodaster 

See also this comment on the previous review round: https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/816#discussion_r134454568

F3PiX:
Thanks! 🥇

F3PiX:
One more thing before we close this.
I don't have access to production. @carpodaster Can you run the rake task? Or who should I ask?

carpodaster:
@alicetragedy can you add Maud to the teams app on Heroku?

Issue #821, opened 2017-08-16
Conferences: show for accepted teams only


mkalininait:
I think we should show the 'Conference' block (incl. 'Conference preferences' and 'Conference offers') on the team view/edit pages only for accepted teams. Conference settings are absolutely useless for the teams who are applying for example.

I think such solution is better than showing the 'Conference' block depending on the current season/phase. We may want to see in which conferences the teams of the previous editions participated. Moreover, some teams have their conferences after the summer of code and they may want to see them on their team pages even after the summer is officially over.

Any thoughts?

klappradla:
Perfect @mkalininait 💚

PR #822, opened 2017-08-16
show conference preferences for accepted teams only

juuh42dias:
## When Team is valid/accepted
rgsoc teams app 5
team 1 gouda rgsoc teams app 1

## When Team is not valid/accepted
team 1 gouda rgsoc teams app
rgsoc teams app 4

PR issue related #821

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -26,7 +26,7 @@ class Ability
     end
 
     can :update_conference_preferences, Team do |team|
-      team.accepted? && (user.student? && team.students.include?(user))
+      team.accepted? && team.students.include?(user)
     end
 
     can :see_offered_conferences, Team do |team|
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -26,7 +26,7 @@ class Ability
     end
 
     can :update_conference_preferences, Team do |team|
-      user.admin? || (user.student? && team.students.include?(user))
+      team.accepted? && (user.student? && team.students.include?(user))
     end
 
     can :see_offered_conferences, Team do |team|
--- a/app/views/teams/_form.html.slim
+++ b/app/views/teams/_form.html.slim
@@ -69,8 +69,8 @@
               = s.link_to_remove 'Remove', class: 'btn btn-default'
       .form-group.form-btn-group
         = f.link_to_add 'Add another source', :sources, class: 'btn btn-primary col-sm-offset-3'
-  -if can?(:update_conference_preferences, @team)
-  = render 'conference_preference_fields', f: f
+  - if can?(:update_conference_preferences, @team) && @team.accepted?
+    = render 'conference_preference_fields', f: f
   .actions
     ul.list-inline
       li  = f.submit 'Save', class: 'btn btn-success'
--- a/app/views/teams/_form.html.slim
+++ b/app/views/teams/_form.html.slim
@@ -69,7 +69,7 @@
               = s.link_to_remove 'Remove', class: 'btn btn-default'
       .form-group.form-btn-group
         = f.link_to_add 'Add another source', :sources, class: 'btn btn-primary col-sm-offset-3'
-  - if can?(:update_conference_preferences, @team) && @team.accepted?
+  - if can?(:update_conference_preferences, @team)
     = render 'conference_preference_fields', f: f
   .actions
     ul.list-inline
--- a/app/views/orga/teams/_form.html.slim
+++ b/app/views/orga/teams/_form.html.slim
@@ -75,20 +75,20 @@
 
   - if admin?
     = f.input :season_id, as: :select, collection: Season.order('name DESC') { |e| [s.name, s.id] }
-
+  - if f.object.accepted?
     = render 'teams/conference_preference_fields', f: f
-  .page-header
-    h3 Conference Offers
+    .page-header
+      h3 Conference Offers
 
-  fieldset
-    = f.simple_fields_for :conference_attendances do |c|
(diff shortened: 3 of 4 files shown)


app/views/teams/_form.html.slim
```diff
@@ -69,8 +69,8 @@
               = s.link_to_remove 'Remove', class: 'btn btn-default'
       .form-group.form-btn-group
         = f.link_to_add 'Add another source', :sources, class: 'btn btn-primary col-sm-offset-3'
-  -if can?(:update_conference_preferences, @team)
-  = render 'conference_preference_fields', f: f
+  - if can?(:update_conference_preferences, @team) && @team.accepted?
```
klappradla: As far as I remember the ability `update_conference_preferences` was introduced specifically for this view / this context. So I would say if we change the restrictions, this should be done inside this specific ability itself, don't you think?
It would also have the benefit of having the view less cluttered with conditions 😉 

So what do you think about adapting the resembling ability and its tests instead?

juuh42dias: Something like this? 
`
    can :update_conference_preferences, Team do |team|
      (user.admin? || (user.student? && team.students.include?(user))) && team.accepted?
    end
`

klappradla: Yes, something along the lines of this. For admin users, it should not make a difference whether or not the team is accepted - I would in fact generally remove the admins here in this ability, as orga has their own team form (there you can simply enable all). And I would probably do the check for `accepted?` first.

juuh42dias: Done
`
can :update_conference_preferences, Team do |team|
  team.accepted? && (user.student? && team.students.include?(user))
end
`

mkalininait:
@klappradla @juuh42dias I think it makes sense to do the same for the orga view. Conferences are relevant only to the teams which participate in RGSoC (=have been accepted) anyway.

juuh42dias:
@mkalininait  only for conference preferences?

```
  - if f.object.accepted?
    = render 'teams/conference_preference_fields', f: f
  .page-header
```

mkalininait:
@juuh42dias It applies to both conference preferences and conference offers.

juuh42dias:
@mkalininait like this?

```
  - if f.object.accepted?
    = render 'teams/conference_preference_fields', f: f
    .page-header
      h3 Conference Offers

    fieldset
      = f.simple_fields_for :conference_attendances do |c|
        = c.input :conference_id, collection: @conferences&.group_by(&:region), as: :grouped_select, group_method: :last, label: "Offered conference", include_blank: "None"
        = c.input :orga_comment, label: "Comment"

      .form-group.form-btn-group
      = f.link_to_add 'Add another conference offer', :conference_attendances, class: 'btn btn-primary col-sm-offset-3'
```

mkalininait:
@juuh42dias Sorry, it's hard for me to judge a feature looking in the code only; I'm not a professional developer. I'll test.

mkalininait:
@juuh42dias I've tested the feature. It seems to be working!
One thing I noticed: as an admin, I don't see the conference preferences block on the non-orga team edit form. On production I do. Is it done intentionally?

juuh42dias:
@mkalininait it's was a change for user permissions, who can view this block or no, but if you want I'll return this method like was before.

klappradla:
Related to this: is it necessary that orga users use the "normal" edit form for teams @mkalininait? I think it would make the things less cluttered and less complex if this was just strictly divided and orga had their own edit form for this... What do you think?

mkalininait:
@klappradla @juuh42dias I don't mind showing offers only in the orga view. I just didn't expect this change in this PR. Let's keep it!

@klappradla Tbh I, as a user with the admin role, don't find this entire system with normal and orga views of the same things super convenient. But since we decided to use it at one point, let's try to stick to it 👍

klappradla:
Hm, I cannot really judge the convenience, don't use it enough for this. To me it seems to be "just" the problem that there's still a lot of interleaving. In an ideal Teams App world, an orga user would not even be able to access a "normal" Team for and submit it.
Any changes an orga person wants to do to data not directly owned by him or her (as in e.g. the personal profile) should be done via the orga area I guess - that's at least how I understand its purpose. Correct me if I'm wrong.

mkalininait:
@klappradla Yeaah... I guess I'm supposed to:
 - go to the orga dashboard
 - open the list of teams
 - find a team I want to edit there
 - edit it!

In fact I'm too lazy and use the `/teams` page, because I can open it much faster from the nav bar.  Then I find a team, press the 'Edit' button and manually add the `/orga` part in the url 🙈  This is my use case at the moment. Maybe I'm not representative though.

In any case, these are low-priority UX reflections which shouldn't affect this PR :)

carpodaster:
> In any case, these are low-priority UX reflections which shouldn't affect this PR :)

It's a valid point, though! And we should tackle this … in another issue/PR :wink:

app/models/ability.rb
```diff
@@ -26,7 +26,7 @@ def initialize(user)
     end
 
     can :update_conference_preferences, Team do |team|
-      user.admin? || (user.student? && team.students.include?(user))
+      team.accepted? && (user.student? && team.students.include?(user))
```
klappradla: 👍  could even be more simpler or am I wrong?
```ruby
team.accepted? && team.students.include?(user)
```
If a user does not have the role "student" he or she cannot be returned as the students of the team I guess.

juuh42dias: No, you're right :+1:

mkalininait:
@juuh42dias Thank you! 🎉

PR #827, opened 2017-08-21
Conferences/accepting rejecting

branquinhoaa:
PR related to the issue #772
- include an option for team students reject/accept a conference offer.
- include a tag in teams/view to show if a team accepted or rejected a conference offer.
- include descriptions into conference headers
- include new fields into exporters


mkalininait:
@branquinhoaa Thank you so much for the PR! It works great!
Two minor issues I noticed:
1) The conference exporter. Somehow it exports the header line with the `,` separator and all other lines with the `;` separator now.
2) I'd add more space between two offers (see below). But this is just a really minor thing.
01e8779783

branquinhoaa:
## new team/view with margin-bottom:

add_space

app/controllers/conference_attendances_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+      redirect_to @team, notice: 'Your attendance was successfully recorded.'
+    else
+      render json: @team.errors, status: :unprocessable_entity
+    end
+   
+  end
+
+  private
+    def attendance_params
+      params.permit(:attendance, :id, :team_id)
```
michaelem: Permitting `id`  in the `attendance_params` suggests that it is part of the data that need to be changed. The way to handle finding a record in an update action would probably be to use `params[:id]`. Also the rest of the params allow the setting of the `:team_id`, is that really intended? I think the controller name suggests that one should only be able to change the attendance attribute. One last thought: the params should probably be nested under `conference_attendance`?

app/helpers/application_helper.rb
```diff
@@ -212,4 +212,21 @@ def list_all_timezones
     TIMEZONES
   end
 
-end
+  #conference attendance helper
```
michaelem: If this should actually be part of a ConferenceAttendanceHelper I would probably move it to it's own helper class/file.

app/controllers/conference_attendances_controller.rb
```diff
@@ -0,0 +1,19 @@
+class ConferenceAttendancesController < ApplicationController
+
+  def update
+    @conference_attendance = ConferenceAttendance.find(attendance_params[:id])
+    @team = Team.find(attendance_params[:team_id])
```
michaelem: I think it would make more sense to not set @team separately dependent on the params but to use `@conference_attendance.team` for the redirects later on. This would communicate more clearly that the team that we are redirecting to is the team that is associated to this `@ conference_attendance` object.

app/controllers/conference_attendances_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+
+    if @conference_attendance.update_attributes(attendance_params)
+      redirect_to @team, notice: 'Your attendance was successfully recorded.'
+    else
+      render json: @team.errors, status: :unprocessable_entity
```
michaelem: I presume that using `@team.errors` is probably inteded to be `@conference_attendance`? `@team` is not saved/validated in this action so it can not have any errors.

app/views/teams/_conference_info.html.slim
```diff
@@ -15,10 +17,26 @@ h3 Conferences
...
\ No newline at end of file
+        .orgaComment
+          |Orga Comment:
+          p = offer.orga_comment
+
+        - if can? :accept_or_reject_conference_offer, @team
+          ul.attendance.list-inline
+            -if offer.attendance != true
+              li
+                = link_to "Yes, I can attend", conference_attendance_path(offer, attendance: :true, team_id: @team.id), method: :put, class: 'btn btn-default'
```
michaelem: I would probably try to nest the params here or use an inline form. If I'm not mistaken this currently generates a path with a bunch of get params and then submits them via put. I think it would be more elegant to have these be submitted as post params.

spec/controllers/conference_attendances_controller_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+    it "cannot update other team's attendance" do
+      other_attendance = create :conference_attendance
+      expect {
+        put :update, params: { id: other_attendance.id, attendance: true }
+      }.to raise_error ActiveRecord::RecordNotFound
+      expect(attendance.attendance).to be_falsy
+    end
+
+    it "can update her team's attendances" do
+      put :update, params: { id: attendance.id, attendance: true, team_id: team.id }
```
michaelem: I don't think this action should take team_id as a param - it's part of the attendance record already and if it is accepted as a param it could be set to anything here.

mkalininait:
Thank you @branquinhoaa ! The screenshot looks good to me. I'll test the entire PR one more time, once everything else mentioned by **michaelem** is fixed.

app/controllers/conference_attendances_controller.rb
```diff
@@ -1,19 +1,18 @@
 class ConferenceAttendancesController < ApplicationController
 
   def update
-    @conference_attendance = ConferenceAttendance.find(attendance_params[:id])
-    @team = Team.find(attendance_params[:team_id])
+    @conference_attendance = ConferenceAttendance.find(params[:id])
+    @team = @conference_attendance.team
 
     if @conference_attendance.update_attributes(attendance_params)
       redirect_to @team, notice: 'Your attendance was successfully recorded.'
```
michaelem: You could still skip line 5 and just use `@conference_attendance.team` here (if I'm not missing any other uses).
That way the only assigned @-variable would be `@ conference_attendance`, signaling that this is the object that we are dealing with in this action. (But this is really a small code style thing.)

michaelem:
@mkalininait all yours ;)

branquinhoaa:
@mkalininait I can fix it in this PR :+1:

branquinhoaa:
@mkalininait Fixed! 
btn_style

mkalininait:
Looks awesome, thank you @branquinhoaa 
Merge! Merge! Merge!

Issue #826, opened 2017-08-21
Can't add a new conference in UI


mkalininait:
When I try to add a new conference I get Error 500 on production. 
This is how the error looks locally:

![screen shot 2017-08-21 at 14 31 21](https://user-images.githubusercontent.com/1745735/29519404-a8e606f0-867d-11e7-8e39-3111aa1e23e6.png)

mkalininait:
Hey @branquinhoaa I assigned you to this bug, but if it's not your area of responsibility, feel free to "unassign" yourself.

branquinhoaa:
When you tried to add the conference, are you logged as an organizer? that error is because we use team id to generate the gid, if you are logged as an orga you have no team.

mkalininait:
@branquinhoaa Ah, I see :) Yep, I am an orga without a team. Would be great to have a possibility to add a conference in this case, too.

If it works for the students, the bug becomes of a lower priority, I guess :)

branquinhoaa:
Yeah, I submitted a PR fixing it. In cases where the user does not have a team, I'm using her user id to generate the GID.

mkalininait:
@branquinhoaa 
Still getting an error when trying to create a new conf as an orga :/

UPD: my test conf (https://teams.railsgirlssummerofcode.org/conferences/110) has been created with a correct gid, so maybe this error is related to the redirect.

UPD 2: I created this conf from `/conferences`, not `/orga/conferences`, which is a valid use case IMO.

mkalininait:
@branquinhoaa Am I getting this right that some confs will have a team id in their gids, and some a user id?

carpodaster:
@mkalininait I got the error on Sentry. It's indeed the redirect. As an orga person, what's your expected/desired redirection target? Students get redirected to their team's edit form which doesn't work for organisers.

mkalininait:
@carpodaster In all cases I'd prefer to get redirected to the same page, where I pressed the "New conference" button (regardless of my roles):
 - team edit page
 - conferences
 - orga/conferences

branquinhoaa:
I will fix the redirect, we had a lot of changes related to this but now I think I can work properly on this issue. Just a minute.

branquinhoaa:
@mkalininait PR #860 I think this will work!

Issue #825, opened 2017-08-21
Conferences: adding a new conf from the list of confs


mkalininait:
It would be convenient to have a possibility to add a new conference from the list of conferences here: `/conferences`.

Add a button "+ New Conference" on the conference page (maybe the same way as it's done for `/teams`?). The button should open the "Create a new conference" form. 

Who should see the button: 
 - orga
 - students of accepted teams

klappradla:
Any obligations in putting this into the `conferences` milestone @mkalininait ? I think it would make sense, now that @juuh42dias and @branquinhoaa are already touching this area of the app, not to leave any "holes".

mkalininait:
@klappradla I'm fine with anything you decide :)

klappradla:
Closing this via #829

PR #828, opened 2017-08-21
refactor action and replace team id by user id

branquinhoaa:
PR related to the issue #826

- Use user id instead teams id into conference gid generator.
- Redirect to conferences after orga create new conference ( but to edit/teams when is a student that create a new conference )
- Refactor action


app/controllers/conferences_controller.rb
```diff
@@ -33,8 +32,15 @@ def redirect
 
   private
 
-  def generate_gid(team)
-    "#{Season.current.name}-#{Time.now.getutc.to_i}-#{team.id}"
+  def build_conference
+    conference = Conference.new(conference_params)
+    conference.season_id = current_season.id
```
F3PiX: You can set the season_id at the same time as the conference_params like this:
`conference = Conference.new(conference_params.merge(season: current_season))`
The same goes (I think) for the generated `gid`, but I don't mind if you leave that separately.

app/controllers/conferences_controller.rb
```diff
@@ -33,8 +32,15 @@ def redirect
...
+    conference.gid = generate_gid
+    conference
+  end
+
+  def generate_gid
+    "#{Season.current.name}-#{Time.now.getutc.to_i}-#{current_user.id}"
```
F3PiX: My main question: If I understand this correctly, this means that _every_ logged_in user can add conferences to the main list now? Because this controller is about the conferences that are imported via the importer, right?

app/controllers/conferences_controller.rb
```diff
@@ -16,12 +16,11 @@ def show
 
   def create
     team = current_user.student_team
```
F3PiX: What happens here if the current_user is not a student?

app/controllers/conferences_controller.rb
F3PiX: Can you add tests please for the non-student users? The current specs cover only logged in _students_ , no other users.
You can wait to do them untill we finished the Q&A ;-)

app/controllers/conferences_controller.rb
branquinhoaa: if a current user is not a student, the team will be empty. 
When the conference was saved, the 'team.present?' will return false and then the user will be redirected to a conferences_path after creating a new conference.
However, this button just appears for users that have an ability to 'update_conference_preferences' -  just orga members and students of that specific team.

app/controllers/conferences_controller.rb
branquinhoaa: Good idea! :dancer:

app/controllers/conferences_controller.rb
branquinhoaa: The button to add new conference are supposed to appear just for a student of a specific team and orga members. However, orga members do not have a team, so I thought that use the current user id in the GID could help to track who added the conference, instead what team added the conference. But as far as I know, just the students (and orga members) are allowed to add conferences. You are right, I have to include tests for it! :D thank you

app/controllers/conferences_controller.rb
```diff
@@ -33,10 +33,7 @@ def redirect
   private
 
   def build_conference
-    conference = Conference.new(conference_params)
-    conference.season_id = current_season.id
-    conference.gid = generate_gid
-    conference
+    Conference.new(conference_params.merge(season: current_season, gid: generate_gid))
```
F3PiX: 👍 Nice!

app/controllers/conferences_controller.rb
F3PiX: Aha, it is only for orga members and students. Then I still don't get the use case for the change to user_id. Sorry ;-) This controllers#create is were the students enter their 'suggestions', right?
So, why should an orga member use _this_ create action?  Why not either in the namespaced orga/conferences? OR even better, import them via the google database. 
I have the feeling that we are trying to press two very different use cases into one. WDYT?

branquinhoaa:
This one is related to the issue #826 (fix a bug in button to a new conference in teams/edit ). The other issue (#825) is an enhancement  - just to add an extra button. 
But if you want I can join in one PR.

app/controllers/conferences_controller.rb
branquinhoaa: Yeah, makes sense just team's student create new conferences. However, in the description of this issue, it was asked to let orga members do it too. Even in the new issue #825 related to add new button, was asked to allow both orga members and students to add new conference.

F3PiX: Okay, I see what you mean! Do you mind if we set this on hold, then? And leave it in this separate PR. 
I am sorry for the complications :-(
But let's first _implement_ the enhancement before we solve bugs in that enhancement 😂

F3PiX:
Hey @branquinhoaa 👋  
Now that #829 has been merged, would you mind rebasing?

klappradla:
Could you please bring this up 2 date with `master` @branquinhoaa ? I think some changes in the `ConferencesController` are already obsolete.... 🙈 

Are all your changes in @F3PiX ?

F3PiX:
@klappradla @branquinhoaa Looks like we are cross-typing ;-)
I'll update/sync my review when the branch is updated. 👍

branquinhoaa:
@klappradla @F3PiX  I updated the PR.

klappradla:
Yay, looks all good from my side. Can you check back @F3PiX ? I'd like to merge this soon.

F3PiX:
Looks good to me too!! Thanks @branquinhoaa 🥇

mkalininait:
Still getting an error when trying to create a new conf as an orga :/

Issue #809, opened 2017-07-31
Conference preferences: adding new conference UX


mkalininait:
Team edit view, conference preferences

This issue derives from here: https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/802#issuecomment-318490848

It would be great to add a new conference in an in-place edit or a popup form instead of opening a new page.

LuisaAPF:
Hi! I'd like to work on this.

klappradla:
Yay @LuisaAPF 🎉 

I'll go ahead an assign it to you. Do you need any more information upfront?

LuisaAPF:
Hi... I'm confused.  You can or cannot assign me this issue?

klappradla:
I cannot for technical reasons (Github doesn't allow me to).

Sorry for causing confusion, I could have skipped the whole assign-topic 😉

LuisaAPF:
That's odd.. Do you think this might be some configuration on my site? Can I work on it anyway?

klappradla:
No, it's simply the way Github works. **But yeah, of course you can work on it** 💚

mkalininait:
I've just tested this issue and noticed the following: while the "We want to suggest another conference!" button works absolutely great for students on their team's page here `/teams/NNN`, it doesn't work at all for the admins in the orga space: `/orga/teams/NNN`. I think we should either make it work or delete it from there.

klappradla:
Closed via #867

PR #829, opened 2017-08-21
Add a button for create a new conference

juuh42dias:
PR related issue #825 

## General Conference View

rgsoc teams app 7

## Orga Conference View

rgsoc teams app 6


klappradla:
Thanx @juuh42dias 🙏 

Before I dive into the code, some suggestions for the UI:
* I'd suggest to put the interaction **below** the table. Background: people should see all existing conferences first before suggesting a new one _(there's no mechanism to prevent duplicates)_
* Same as with the `ConferencePreferences`, we need some text / some explanation here. I feel like we need to stop just putting buttons without explaining the user what's possible there 🙈   
The text could be something like _"The tech conference you're looking for is not yet in our list? Just suggest it!"_

What do you think?

juuh42dias:
@klappradla 

## Orga Conference view 
rgsoc teams app 9

## General Conference view
rgsoc teams app 8

mkalininait:
Hi @juuh42dias Thank you for the PR. I have two issues with it:

1) **Security.** It's great that the students of the unaccepted teams don't see the "New Conference" button on the list of conferences, BUT they can access the form directly by the url `/conferences/new` and successfully add a conference this way.
I think this issue was not introduced in this PR, it probably existed from the beginning. Do you think we can (and should) fix it in this PR @klappradla @juuh42dias ?

2) **Redirect.** When I add a conference from the conf list, I get redirected to the team edit form. I understand why it works this way, it makes sense when you add a conference from the team edit form. But when you add a conference from the conf list, it makes sense to return back to this page (normal or orga view correspondingly).

klappradla:
Thanx for pointing out @mkalininait 🙏 

Yes, I think we should definitely fix the access rights to `conferences/new` - by using the appropriate `cancancan`check in the controller @juuh42dias ✌️ 
You could do this in a separate PR though to not change the topic here too much.

I'll quickly create a ticket for this @juuh42dias ☝️

klappradla:
Here's the respective ticket for ☝️ #840 -> I did not yet assign it to you @juuh42dias , maybe @branquinhoaa wants to take care of it. Anyways, feel free to assign it to whoever does it.

app/controllers/conferences_controller.rb
```diff
@@ -15,26 +15,33 @@ def show
...
+      redirect
     else
       render action: :new
     end
   end
 
   def redirect
-    redirect_to orga_conferences_path
+    message = 'Conference was successfully created.'
+    case true
```
klappradla: You can skip the `true` here ✌️

app/controllers/conferences_controller.rb
```diff
@@ -15,26 +15,33 @@ def show
...
+      redirect_to @conference, notice: message
+    end
   end
 
   private
 
-  def generate_gid(team)
-    "#{Season.current.name}-#{Time.now.getutc.to_i}-#{team.id}"
+  def generate_gid(user)
+    "#{Season.current.name}-#{Time.now.getutc.to_i}-#{user.id}"
```
klappradla: Hm, just wondering, does this have any impact on the conferences imported up to now?

app/models/ability.rb
```diff
@@ -57,6 +57,8 @@ def initialize(user)
       user.admin? || (preference.team.students.include? user)
     end
 
+    can :crud, Conference if user.admin? || user.current_student?
```
klappradla: According to the comments on this PR: shouldn't this only be allowed for students from accepted teams?

app/controllers/conferences_controller.rb
```diff
@@ -48,7 +55,7 @@ def conference_params
   end
 
   def confirm_role
-    redirect unless current_user.admin? || current_user.student?
+    redirect_to root_path, alert: 'You are not authorized to access this page.' unless current_user.admin? || current_user.current_student?
```
klappradla: This is somehow duplicating what you define as abilities to `crud` conferences - could you also reuse the ability here in the controller?

app/controllers/conferences_controller.rb
```diff
@@ -15,26 +15,33 @@ def show
...
 
     if @conference.save
-      redirect_to edit_team_path(team), notice: 'Conference was successfully created.'
+      redirect
     else
       render action: :new
     end
   end
 
   def redirect
```
klappradla: Having these edge cases based on the current user's role makes me wonder if we should not just have a dedicated controller for the orga namespace: `controllers/orga/conferences`... 🤔 

On the other hand the amout of logic here is still handleable... What do you think @juuh42dias ?

spec/models/ability_spec.rb
```diff
@@ -403,4 +403,18 @@
       expect(ability).to be_able_to(:join, helpdesk_team)
     end
   end
+
+  context 'Crud Conferences' do
```
klappradla: Yay, looks already really cool 👍 

Do these specs also work if you use `build_stubbed` instead of `create` (would speed them up 😉 )
Plus: it would be convenient to also add a negative example, where a user cannot crud conferences.

app/controllers/conferences_controller.rb
F3PiX: We already have that controller. 💃  https://github.com/rails-girls-summer-of-code/rgsoc-teams/blob/master/app/controllers/orga/conferences_controller.rb

klappradla: Oh, should have looked at the repo 🙈  thanx a million @F3PiX 🙏 

So yeah, I guess we should opt for this controller for users coming from the orga views @juuh42dias ✌️

F3PiX: I wonder if it makes sense to add a separate form in the orga namespace? So we don't have all the cancancan checks in the views. AFAIK one of the goals of the namespaces is getting independent of the cancancan checks. WDYallT?

juuh42dias: @klappradla I don't understand what you mean.

app/controllers/conferences_controller.rb
juuh42dias: @branquinhoaa  says now we use 'user' instead 'team'.

app/models/ability.rb
juuh42dias: `user.current_student?` returns students from accepted teams.

```
  def current_student?
    roles.joins(:team).
      where("teams.season_id" => Season.current.id, "teams.kind" => %w(sponsored voluntary)).
      student.any?
  end
```

app/controllers/conferences_controller.rb
F3PiX: 🤔  The change to :user instead of :team is implemented in another PR #828. So, it is better to not change it here. 👍

app/controllers/conferences_controller.rb
klappradla: Sorry @juuh42dias 🙈 
I just meant: the `ConferencesController` should not have any specific logic to distinguish between `admin` and normal users. `admin` users should instead use the controller in the `orga` namespace (plus the respective forms and views of course).

app/models/ability.rb
klappradla: Oh ok, looks fine, thanx @juuh42dias 🙏

mkalininait:
@juuh42dias It looks like the redirect issue is still in place. What do you think about it? Are you planning to fix it?

cc @klappradla

klappradla:
Any news on that @juuh42dias? Do you need help somewhere? Or are you already done?

juuh42dias:
@klappradla I believe that's ok. I created new/create a page for orga conferences to solve so many redirects into conferences of non orga.

LuisaAPF:
Hi everyone! I would like to point out that the redirect to the teams/edit page won't be necessary anymore, as the conference will be created with an ajax request. I will send my pull request on this issue today.

F3PiX:
@LuisaAPF Sounds interesting :-) What issue number are you working on?

LuisaAPF:
Sorry, I forgot to mention it,  @F3PiX. It's issue #809 adding new conference ux.

klappradla:
Ok, thanx for the info 🙏 

So good to merge from my side. Any obligations @mkalininait ?

klappradla:
Ok, just tried it out locally. Orga users cannot use the normal form for creating conferences (500) - but I guess that's fine in the end and for now at least @mkalininait

klappradla:
🎉  thanx for you efforts @juuh42dias 🙏

mkalininait:
@klappradla Sorry, was super busy these days. Thanks for merging the PR! The redirect issue will be resolved with PR #850 I guess, and the Error 500 for admins with PR #828. 

@juuh42dias Thank you for your work!

LuisaAPF:
@mkalininait, I actually didn't change the redirect on my PR #850. I think there is no need for a respond to html on the conferences_controller/create anymore, as this is already being dealt with on the respective method under the orga folder. If you agree, I can do the clean up.

mkalininait:
@LuisaAPF Ah, ok. Sorry, I misunderstood it then.

I'm not sure we can completely delete the respond to html on the conferences_controller/create, because we still can create a conf from here: `/conferences` (not orga). Please tell me if I'm getting this wrong.

LuisaAPF:
Hey, @mkalininait, no problem :) I'm still a bit confused myself. So, the idea is to forbid the access through the path conferences/new, right? For non-organizers, it will only be possible to add a new conference from the modal inside the teams/edit form. So I guess the conferences_controller/create should respond only to an ajax request.

mkalininait:
Hey @LuisaAPF I'm terribly sorry that I'm replying to you only now. The end of Summer of Code was a busy time for us, and I couldn't find time to go through GH updates until this week.

>  For non-organizers, it will only be possible to add a new conference from the modal inside the teams/edit form.

It's not entirely correct. We want to let students of *accepted* teams (= teams which participate in the summer of code) to add new conferences; it doesn't matter how they do it, from their team's page or from the list of conferences. 
We do want to restrict access to the users who registered in the Teams app, but haven't been accepted to RGSoC, see [this comment](https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/829#issuecomment-324588854). Here is an issue to fix this: #840 

I listed all possible scenarios on how a new conferences can be added from UI here: https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/860#issuecomment-334148668 (by students we mean *accepted* students there). I think it looks good in general, we just need to fix the last redirect and access for the users other than accepted students and orga.

LuisaAPF:
I get it now. Thanks!

PR #830, opened 2017-08-21
Fix error handling in csv upload

### LINKED ISSUE #819: Conference import: error when a file is not chosen
`/orga/conferences`

If I don't choose a csv file and press the "Import csv file" button, I get this error:
screen shot 2017-08-16 at 12 16 58

It would be great to show a user friendly error message instead.

carpodaster:
Fixes #819 but should also prevent this error from happening at all by disabling the submit field until a file has been selected.

### DIFF
--- a/app/views/orga/conferences/_form.html.erb
+++ b/app/views/orga/conferences/_form.html.erb
@@ -1,9 +1,13 @@
 <h1>Import conferences (.csv only)</h1>
 <p>Importing will update existing conferences with the data from the csv file</p>
 
-<%= form_tag import_orga_conferences_path, multipart: true, class: "form-inline" do %>
-  <%= file_field_tag :file, accept: '.csv' %>
-  <%= submit_tag "Import csv file", class: "btn btn-default btn-small"  %>
+<%= form_tag import_orga_conferences_path, multipart: true do %>
+  <div class="form-group">
+    <%= file_field_tag :file, accept: '.csv' %>
+  </div>
+  <div class="form-group">
+    <%= submit_tag "Import CSV file", class: "btn btn-default btn-small"  %>
+  </div>
 <% end %>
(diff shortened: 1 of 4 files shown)


app/views/orga/conferences/_form.html.erb
```diff
@@ -1,14 +1,18 @@
 <h1>Import conferences (.csv only)</h1>
 <p>Importing will update existing conferences with the data from the csv file</p>
 
-<%= form_tag import_orga_conferences_path, multipart: true, class: "form-inline" do %>
-  <%= file_field_tag :file, accept: '.csv' %>
-  <%= submit_tag "Import csv file", class: "btn btn-default btn-small"  %>
+<%= form_tag import_orga_conferences_path, multipart: true do %>
+  <div class="form-group">
+    <%= file_field_tag :file, accept: '.csv', id: 'input-csv-file-upload' %>
```
F3PiX: Either way, how about setting `autofocus: true` on the file field? That makes is hard to miss.

carpodaster: Even with autofocus, you could still fire the submit button. It doesn't matter now that the error is caught in the controller as well, but theoretically, one should never even get there anymore: the JS disables the submit button when there is no file selected.

carpodaster:
@F3PiX i've removed the whitespace

F3PiX:
@carpodaster I got an TypeError in console.  

screen shot 2017-08-23 at 09 55 15

carpodaster:
Ah shoot. The script is loaded everywhere – not just in orga/conferences_controller. My bad, that's a bug. I won't have time today to fix it. Does it break other JS?

F3PiX:
My first impression is : yes. The download button on the same page acts weirdly also.

I can do a temp patch or something right now. Does that help?

PR #831, opened 2017-08-21
Prefer Vanilla JS and ERb templates

carpodaster:
Opionated PR. `/discuss` :smile:  Happy to close if this feels too constraining.

Ping @klappradla @F3PiX @branquinhoaa @juuh42dias

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -96,6 +96,8 @@ If you've cloned the app a while ago, you want to make sure that your cloned `ma
 
 We aim to follow the [Ruby Style Guide](https://github.com/bbatsov/ruby-style-guide) but we don't enforce it (i.e. there is no [RuboCop](https://github.com/bbatsov/rubocop)). The most important thing is probably just proper indentation: two whitespaces.
 
+We prefer JavaScript (`*.js`) over CoffeeScript (`*.coffee`). jQuery is perfectly fine, but if you can, use [Vanilla JS instead of jQuery](https://gist.github.com/liamcurry/2597326). Indentation for JavaScript files: four whitespaces.
+
 If you enjoy polishing up other people's code: be aware that **purely** *code-cosmetic* changes are likely to not be accepted. We aim to be welcoming to beginners. So while e.g. pointing out accidental N+1 queries is a useful contribution, just rewriting things to look more conventional can be discouraging for newcomers.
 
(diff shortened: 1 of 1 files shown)


F3PiX:
I'd love to get rid of slim 👍 
For me, using jQuery is easier than being forced to use plain js. Talking about beginner friendly: as jQuery used to be part of Rails, its use in Rails is better documented than plain js. 
So, I'd vote against vanilla js. 🍨

carpodaster:
@F3PiX 🤔  You might be right about the better documentation. But this is certainly to no small amount thanks to:

> Q: How do I upcase a String in JS?
> A: Here, take my 400KB jQuery plugin that does this for you! `[Accepted answer]`
> – Virtually _every_ Stackoverflow about JavaScript 

😉 

What's your take on CoffeeScript?

F3PiX:
I agree with not using CoffeeScript!

branquinhoaa:
I used to prefer .erb, but after the 'teams app' I think I really like slim. The .erb files look so polluted!
about the CoffeScript, I agree with not using it.

juuh42dias:
I preferer .erb but I vote for use .slim (because is a Teams App pattern) and Vanilla JS(JQuery) :smile:

My votes:
Slim Template Engine :white_check_mark: 
Vanilla JS/JQuery :white_check_mark:

klappradla:
Great topic @carpodaster 😄 

**JS vs. Coffeescript**
**JS** -> 👍 for not using Coffeescript anymore (it's basically dead and I don't see a point for anyone diving into it - even the Atom editor gets now fully rewritten in Vanilla JS)

**slim vs. erb**
❔ -> I don't care, `slim` is pretty close to `haml` which almost as widely used as `erb`, don't see it as that huge of a weirdness of the Teams App

**jQuery vs. Vanilla JS**
**Vanilla JS** -> 👍 for solving things with pure JS 🎉  I think it really increases code quality and understanding. Plus: JS is not that bad anymore _(and does even feel better if you skip the jQuery parts)_

carpodaster:
Thank you all for participating in this discussion! :green_heart: 

I've removed the paragraph about Slim vs. ERb. I changed the JS one so that it :ok: s jQuery but encourages Vanilla JS.

What do you think?

klappradla:
🎉 all good to 🚢 in my opinion
(although writing code with 4 spaces indentation kinda hurts my eyes 😹 )

michaelem:
Funny I did not see this but started thinking about doing a PR that removes coffeescript since it's only 3 files left I think. Anybody already on that? (Otherwise I'd do it this weekend.)

carpodaster:
Last call for objections :grinning:

PR #838, opened 2017-08-24
Prevent overload of errors on rake task to set project_id

F3PiX:
Follow up on #816

Populating the new `project_id` on teams via the `team.project_name` will generate 'RecordNotFound' noise. Adding a scope to make the task running on accepted teams only should reduce some of that noise.


F3PiX:
To further reduce the amount of errors would be to run the task on teams `:accepted.in_current_season`. Replace the `project_name` input field on the Team form with a project dropdown. 
Because currently, we only use the project_id  in the Projects index view for the current season. 
🤔

Issue #823, opened 2017-08-16
Test ConferencePreference scopes and ConferencePreferences exporter


klappradla:
As said in #813, it would be nice to have a dedicated test for the `current_teams` scope in the `ConferencePreference` model specs. If that is properly tested, we can rely on this contract with the conference_preference exporter. This means: we stub out this method and make the test there less heavy on the database. It is enough for us to know **that `current_teams` get's called**, the internals of this method are tested elsewhere (in the model specs).

So, try to define what the actual behavior of the exporter is, and then test this, potentially stubbing out the internals of other parts of the system that are already tested elsewhere.

carpodaster:
Done in #839

PR #839, opened 2017-08-25
Tests/conference preference

branquinhoaa:
Related issue #823
- Change the test for conference preference exporter (test if receives current teams)
- Include file to test model of conference preference


branquinhoaa:
Thank you @carpodaster! I searched a little bit about contexts and you are right, it is really better to use it. I refactored the test file with your suggestions, but if you think there is another thing to improve just let me know.

spec/exporters/conference_preferences_spec.rb
```diff
@@ -3,14 +3,9 @@
...
-    let!(:old_team) { create :team, :with_preferences, name: "OLDTEAM" }
-    let!(:new_team) { create :team, :with_preferences, :in_current_season, name: "NEWTEAM" }
-
-    subject { described_class.current }
-
-    it 'exports conference preferences for teams of the current season' do
-      expect(subject).to match 'NEWTEAM'
-      expect(subject).not_to match 'OLDTEAM'
+    it 'calls the current_teams scope in conference preference' do
+      expect(ConferencePreference).to receive(:current_teams).and_return([])
```
carpodaster: Oh yeah, smart! :tulip:

spec/factories/teams.rb
```diff
@@ -36,6 +36,12 @@
       end
     end
 
+    trait :without_terms do
```
carpodaster: Is `without_terms` the right name for this? I feel it should be `:with_conference_preference`

spec/models/conference_preference_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+  it { is_expected.to belong_to(:first_conference) }
+  it { is_expected.to belong_to(:second_conference) }
+
+  describe 'validates terms of ticket and terms of travel' do
+    let(:preference_without_terms) { build(:conference_preference) }
+    let(:preference_with_terms) { build(:conference_preference, :with_terms_checked)}
+
+    context 'when both terms are not checked' do
+      it 'do not save conference preferences' do
+        expect(preference_without_terms).not_to be_valid
```
carpodaster: Checks for overall object validity may fail because a different error is causing the object to be in an invalid state. Personally, I prefer something in the form of this:

```ruby
expect(subject.errors[:my_attribute_name]).not_to be_empty
```

(You may need a call to `subject#valid?` first to trigger the validations)

spec/models/conference_preference_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+
+    context 'when conference preference terms are accepted' do
+      it 'returns true' do
+        expect(conference_preference_with_terms.terms_accepted?).to eql true
+      end
+    end
+
+    context 'when conference preference terms are not accepted' do
+      it 'returns false' do
+        expect(conference_preference_without_terms.terms_accepted?).to eql false
```
carpodaster: RSpec adds automatic matchers for predicate-style methods (= methods that end with a `?`).

```
expect(conference_preference_without_terms).not_to be_terms_accepted
```

spec/models/conference_preference_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+      end
+    end
+  end
+
+
+  describe '#has_preference?' do
+    let(:conference_preference) { FactoryGirl.build(:conference_preference, :with_terms_checked)}
+
+    context 'when conference preference has neither first or second choice set' do
+      before { conference_preference.first_conference_id = nil, conference_preference.second_conference_id = nil }
```
carpodaster: I think the comma should be a `;`.

Better to use a multi-line `do … end` before block here

spec/models/conference_preference_spec.rb
branquinhoaa: I used it in the other method too. But sounds a little bit strange: 'not_to be_has_preference' :joy: 
That is ok?

PR #841, opened 2017-08-26
Add tests for Role#github_handle=

carpodaster:
Related to PR #837 and issue #794

 - Minor refactoring of `Role` and its spec file
 - Remove `GithubHandle` module and puts everything in `Role` directly


app/models/role.rb
```diff
@@ -40,6 +41,13 @@ def includes?(role_name)
     end
   end
 
+  def github_handle=(github_handle)
+    return unless github_handle.present?
+    self.user = User.where('github_handle ILIKE ?', github_handle)
+                    .first_or_initialize(github_handle: github_handle)
+    user.github_import = true
```
carpodaster: Something for another PR, but I wondered if this should only be set when a user with that DB couldn't be found

app/models/concerns/github_handle.rb
```diff
@@ -1,15 +0,0 @@
...
-
-  def github_handle
-    user.try(:github_handle)
-  end
-
-  def github_handle=(github_handle)
-    return unless github_handle.present?
-    self.user = User.where('github_handle ILIKE ?', github_handle)
-                    .first_or_initialize(github_handle: github_handle)
-    user.github_handle = github_handle
```
carpodaster: I removed this line. At this point, every user instance should have the `github_handle` set properly already. Or am I overlooking something?

app/models/role.rb
```diff
@@ -27,7 +28,7 @@ class Role < ActiveRecord::Base
 
   class << self
     def includes?(role_name)
-      !where(name: role_name).empty?
+      where(name: role_name).any?
```
klappradla: 👍

app/models/role.rb
klappradla: Yes, would make sense. Otherwise it should already be set to `true` anyways if I understand correctly... 🤔

app/models/concerns/github_handle.rb
klappradla: Yes, `first_or_initialize` should have done it already 👍

Issue #763, opened 2017-06-16
DRY up conferences controllers


F3PiX:
There are currently two conferences controller, controllers/orga/conferences_controller.rb and controllers/conferences_controller.rb.
They share a few methods, like the methods for sorting the columns. 

- [x] Propose a solution to make those methods shared by the two controllers.
- [x] Make it happen 🚀

juuh42dias:
Has any possibility for work together on this issue? Me and @branquinhoaa

F3PiX:
The conferences controllers will change dramatically after #775 
I'd suggest to wait for that one, or be sure to check with your coaches/mentors.

carpodaster:
I think the conference controller has undergone its heavy-weight changes now. It would be ready for refactoring.

m-gb:
Is this project open to pull requests?

klappradla:
Sure it is @m-gb 🤗 
Please check our [Readme](https://github.com/rails-girls-summer-of-code/rgsoc-teams) for all general information and especially our [contribution guide](https://github.com/rails-girls-summer-of-code/rgsoc-teams/blob/master/CONTRIBUTING.md) on how to get involved 📚 

**Before doing anything**, please be sure you have read and understood our [code of conduct](https://github.com/rails-girls-summer-of-code/rgsoc-teams/blob/master/CODE_OF_CONDUCT.md) 🦄

m-gb:
Hi @klappradla, I've created PR #842 to address the issue, let e know what you think!

klappradla:
Closed via https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/842

F3PiX:
@m-gb 👏  🎉  🥇 
Great to see this solved, and chapeau for the way you solved it. 🙇

F3PiX:
@m-gb Just to confirm:  I noticed that you picked one of the few issues that are labeled beginner friendly, and not rgsoc17. You are very welcome to pick issues with the rgsoc17 label as well. Just check if someone is assigned, or ask in the Issue if it is available.

PR #842, opened 2017-08-27
Added a module in the concerns folder to reduce duplicate code in conferences controllers

m-gb:
Related issue #763 

I've moved a few duplicate methods into a separate module inside the concerns folder.
Let me know if this helps :)


app/controllers/concerns/conferencable.rb
```diff
@@ -0,0 +1,18 @@
+module Conferencable
```
klappradla: We aren't super opinionated on syntax, but there's **this one thing™** we care about (see the [contribution guide](https://github.com/rails-girls-summer-of-code/rgsoc-teams/blob/master/CONTRIBUTING.md#write-code)):
>  indentation: two whitespaces

So, for the sake of consistency, would be super cool if you could update this to be with 2 instead of 4 spaces indentation 💚

app/controllers/conferences_controller.rb
```diff
@@ -2,14 +2,12 @@ class ConferencesController < ApplicationController
   before_action :authenticate_user!, except: [:index, :show]
   before_action :confirm_role, except: [:index, :show]
 
+  include Conferencable
```
klappradla: Hm, this name sounds super reasonable if this would be about models... But in controller-context, this I'm afraid does not really name the behavior or "thing" the module is about... 🤔 

Do you think you can come up with a name better describing / "naming" the logic or behavior encapsulated in the new module? Would be ace ✌️ and make it way more understandable for someone looking at these controllers for the first time.

app/controllers/concerns/conferencable.rb
```diff
@@ -0,0 +1,18 @@
...
+    extend ActiveSupport::Concern
```
carpodaster: I don't think we need this line. We're just mixing in instance-level methods and there are no nested includes. It should work with extending the module.

app/controllers/concerns/conferencable.rb
```diff
@@ -0,0 +1,18 @@
...
+    
+    def index
+      @conferences = conferences
+    end
+
+    def conferences
```
carpodaster: This method and sort_params should be private

app/controllers/concerns/conferencable.rb
```diff
@@ -0,0 +1,18 @@
...
+        Conference.ordered(sort_params).in_current_season
```
carpodaster: I know this PR just moves existing code into a different file, but while we're at it: what do you think about removing this method entirely and call it in `index` directly? It's not complex logic and I would argue that it improves the narrative of the code to remove that level of indirection.

F3PiX:
Cool @m-gb ! 
Re:  naming How about naming the module `ListConferences`

m-gb:
Hi all, thanks for commenting!
I've tried to incorporate the changes you requested.
I renamed the module to "OrderedConferences" because that's the end result of the module as I understand it. 
For testing, I tried to create a dummy class which includes the module so it's an isolated spec. I'm just not sure how to proceed, it's currently throwing an error about params being undefined. :(

spec/controllers/concerns/ordered_conferences_spec.rb
```diff
@@ -0,0 +1,18 @@
+require 'spec_helper'
+
+RSpec.describe OrderedConferences do
+  render_views
+
+  describe '#index' do
+    let!(:dummy_class) { Class.new { extend OrderedConferences } }
```
carpodaster: The anonymous class you're defining doesn't have a `params` method, that's why you're getting the error. Further down you're checking for `@conferences` but that would reference an instance variable of the _test class_.

Does this code work? (only written here on GH, haven't run it anywhere)

```ruby
subject { dummy_class.new.index }

let!(:dummy_class) do
  Class.new do
    include OrderedConferences
    def params; {} end
    attr_reader :conferences # exposes the @conferences ivar
  end
end

# other let variables

it 'returns an ordered list of conferences' do
  expect(subject).to eq  [conference_africa, conference_europe, conference_na]
end
```

m-gb:
Hey @carpodaster thanks for your suggestion.
That worked perfectly! I've adapted your example into the code and all tests are passing once more.

klappradla:
🚢  🎉

Issue #642, opened 2017-02-05
Issue with user confirming workflow: clicking on confirmation link in email throws an error


alicetragedy:
This bug is probably related to https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/635; we've had a couple of reports of people being unable to:
- receive a confirmation email and
- confirm their account once the confirmation email was triggered manually

When clicking on the token in the confirmation link, users are directed to a page that throws an internal server error (500). It seems that all users affected ended up with a confirmed account after clicking the link, so the issue might really just be in the page users get redirected to?

mkalininait:
May be this error: https://sentry.io/rgsoc/rgsoc-teams/issues/247596626/
(based on what user 2458 wrote to our support).

LuisaAPF:
I get an error if I click on the confirmation link while logged out. I could solve this by redirecting to the root path, instead of the sign in path. The difference is that now the user is not automatically signed in. I don't think this might be a problem, as this is the default behavior of devise and is the recommended procedure in terms of security https://stackoverflow.com/questions/18655334/avoid-sign-in-after-confirmation-link-click-using-devise-gem

carpodaster:
Would someone find the time to check if this error has been solved by #843? Merci beaucoup!

mkalininait:
@carpodaster @LuisaAPF 

>I get an error if I click on the confirmation link while logged out. 

I can't reproduce this anymore. If this is the only use case when we get the error described in the issue, then we can probably close it 👍

carpodaster:
I have never been able to reproduce the error deterministically

klappradla:
I at the time forgot to comment it on #843, but I could reproduce the problem locally and then have the new code fix it @carpodaster. So I guess it's fine 🙈

PR #843, opened 2017-08-27
Stops trying to sign in user after email confirmation

LuisaAPF:
Related issue #642 

 - Fixes the workflow of email confirmation when user is not logged in. Instead of redirecting the user to the sign in path, it redirects now to the root. This might be a better approach, as it complies with devise's default behavior and security, as discussed in this stackoverflow issue: https://stackoverflow.com/questions/18655334/avoid-sign-in-after-confirmation-link-click-using-devise-gem


LuisaAPF:
Thanks : ) Someone asked me if I could have solved this issue by simply removing the new_session_path method. Well, I tried this and got a no method error. In [this](https://github.com/plataformatec/devise/wiki/OmniAuth:-Overview) documentation, it is stated that if you are using only omniauth authentication, you have to define this method on your application controller.

klappradla:
Ah thanx @LuisaAPF 🙏 
I think I asked it somewhere in a comment. But I already tried it out myself so did not come back to it 😉

Issue #84, opened 2014-03-28
Add a CommunityController


carpodaster:
Currently `UsersController` behaves differently for the RGSoC ramp up phase than it will once the summer has begun. There's lots of commented code.

I'm thinking bringing `users#index` to a CommunityController that displays what we have now. Launching the summer is then a matter of editing the routes file instead of commenting out / in everything (which we'll have to do for 2015 all over again).

Are there other items where the teams app behaves differently depending on what phase we're in?

Thoughts?

carpodaster:
#108 is related.

klappradla:
It seems this can be solved on the way to the new community page 👉  moving this into the [community page milestone](https://github.com/rails-girls-summer-of-code/rgsoc-teams/milestone/13)

branquinhoaa:
@carpodaster we are discussing this issue and we have some questions:
Since /community will be responsible for listing and searching users, we are wondering if it is useful to have an action /users. Can we simply remove it? 
Besides the community link in the header, the path to /users is used when someone calls the 'impersonate' method. The person is redirected to /users. We can change this behaviour, redirecting the person for /user/:id (if you agree to remove the users index).

carpodaster:
@branquinhoaa you are right, I also think we don't need the `UsersController#index` action anymore (the rest of the actions are). Redirecting to `UsersController#show` after impersonation a user makes sense, too! :ok_hand:

carpodaster:
@klappradla @mkalininait we don't really have a Definition of Done for this item, do we? I'm inclined to close this (very old) issue now that #844 has been merged. Subsequent changes (e.g. filters) should be covered by other issues. What do you think?

klappradla:
Yep, would have suggested the same 💚

mkalininait:
Sounds good!

PR #844, opened 2017-08-30
Community/add community controller

branquinhoaa:
Related issue #84
- include a controller for community (listing and searching all users)
- remove the action index for user controller.


config/routes.rb
```diff
@@ -27,6 +27,7 @@
   resources :comments, only: :create
   resources :conferences
   resources :contributors, only: :index
+  resources :community, only: :index
```
carpodaster: Since the index action has been moved, can you remove index from the call to `resources :users` in line 23 please?

app/controllers/users_controller.rb
```diff
@@ -76,12 +59,12 @@ def destroy
 
   def impersonate
     impersonate_user(@user)
-    redirect_to users_path, notice: "Now impersonating #{@user.name}."
+    redirect_to user_path(@user), notice: "Now impersonating #{@user.name}."
```
F3PiX: I am curious about this change in behaviour. The impersonating buttons are still in the index list, right? What is the reasoning for redirecting to the user profile now?

branquinhoaa: We could use the community_index_path as well. I thought that would be more intuitive the user be redirected to the page of the user he chooses to impersonate, but I can change it.

F3PiX: Yes please! 👍  I'd appreciate that. 💃

branquinhoaa: @carpodaster about the redirect I asked you. Do you prefer to keep the old behavior?

carpodaster: @branquinhoaa I don't mind, really. I think it makes sense to redirect to the person whose identity you've just assumed but it's also changing a behaviour whereas everything else in this PR basically maintains the old behaviour, just moves it between controllers.

@F3PiX seems to be in favour of keeping the existing behaviour, so let's roll with that.

branquinhoaa:
Changes were done! :+1:

F3PiX:
🙇 Thanks for the redirect fix! 😃

PR #846, opened 2017-09-03
remove coffeescript

michaelem:
This removes all remaining coffeescript from the application and replaces it with equivalent javascript.

I've done sone poking around and everything seems to precompile and work as expected, although the app has no tests covering the javascript as far as I can tell. (Currently plotting on adding some system tests.)

Caveats: 
* I'm really not good at this js thing, these might not be up to current coding standards.
* I think we should probably try to unify our javascript somewhat (although this might be in opposition to our contribution guidlines).


F3PiX:
> I think we should probably try to unify our javascript somewhat (although this might be in opposition to our contribution guidlines).

It looks like we already does that, by not allowing ES6 🤣

michaelem:
@F3PiX yes, about that: The last commit to jshint master was more than 2 years ago (quasi 2 millennia in javascript time), I think it might be time to drop it from the build process? Of course I can also edit this PR to not use ES6 syntax if that is the consensus.

F3PiX:
I don't know enough about js set ups to have an opinion on this. :-( 
But I love (using) ES6.

carpodaster:
I don't do JS but I'm all for ES6 from what I've heard. I don't know the new language constructs, so my question would be (before I lock myself out of JS on this project): is it possible/advisable to mix "old" style JS with ES6? (sorry if that is a noob question, but I _am_ a JS noob)

F3PiX:
Yes, you can use them together. (BTW ES6 will be compiled to 'old' style JS.)

Gemfile
```diff
@@ -26,7 +26,6 @@ gem 'font-awesome-sass'
 gem 'pretender'
 
 gem 'sass-rails'
-gem 'coffee-rails'
```
klappradla: 💚 for also straight removing it from the `Gemfile` 👍

app/assets/javascripts/rating.js
```diff
@@ -0,0 +1,32 @@
+class RatingSidebar {
+  constructor() {
```
klappradla: Hm, 🤔 I've never seen this in JS yet, are you sure this works? I know `constructor` is a thing in `.coffee`, but for JS I thought one needs to explicitly call it...?

michaelem: re: "I'm really not good at this js thing" ;)
I'll take another look at the rating js 👍

LuisaAPF:
My pull request #850 is throwing an error because I'm trying to use ES6.
Should we allow this syntax on the whole project? Meanwhile I'll go back to
the old syntax.

2017-09-08 11:15 GMT+02:00 michaelem <notifications@github.com>:

michaelem:
@klappradla had to refactor the rating thing anyway in order to move away from the es6 syntax, seems to do the popover and scrolling as I expect it now.

Generally I would much prefer to use the ES6 syntax, but let's see if that works with babel then.

Anyway, from my side I think this is done, pending another review.

F3PiX:
@michaelem 👏

michaelem:
@LuisaAPF re: ES6: I think we should allow it, but I think if we switch to webpacker we can use babel to transpile before it hits jshint. Then again: not sure if that is a good idea?

F3PiX:
I'd vote for collecting the es6/webpacker discussions in that issue, if that is okay? Easier to keep track of the arguments. 

For instance I thought we already decided on es6. 🙃

klappradla:
Yay, thanx a ton @michaelem 🙏 

I don't know anything about the whole JS party, but am still ➕ for doing it ✌️ 

Anyways, for now this works, lets have the ES6 stuff in a separate PR.

Issue #594, opened 2017-01-07
Community page: “seasoning” and filters


mkalininait:
Users reported that it was hard to use the community functionality in the Teams app, because it doesn't allow you to filter users.

It would be useful to provide filters by:
 - availability in the upcoming season
 - location
 - technologies (see [Issue 593](https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/593))

**One possible way to implement "seasoning":**
 - Add a checkbox "Active in the upcoming season" - should be reset to 0 for all profiles at the end of each season

carpodaster:
It is true that with the Teams App keeping users of past seasons, it's hard to know whether or not someone who once put them on the roster as willing to help would still be up for it in the upcoming season.

I don't know if a filter solves that, since it would require all users on the Teams App to proactively check that option. How would we do that? Emailing all users?

mkalininait:
@carpodaster I think that we can't consider a user active, unless he/she explicitly marks him/herself active in the Teams app. Emailing and tweeting would be the most obvious ways to remind them to set a check.
I guess that this check would be most useful for coaches and students looking for coaches. The "Active in the upcoming season" check wouldn't guarantee availability of a coach of course, but it would increase the probability of it. If a student can't find an active coach, she can always try to contact a coach without the "Active in the upcoming season" check.

F3PiX:
If that is the use case: students finding coaches, how about this: add a filter for NON-active coaches, let coaches be able to adjust their availability in their profile.

Easier for coaches, bigger chance for students to persuade a coach to join their team.

mkalininait:
@F3PiX Do you mean that the coaches who won't be available should set a check "I'm not available in the upcoming season"? I'm afraid that people who can't or lost interest in coaching may not bother reporting this. Or did I misunderstand you?

F3PiX:
Exactly! And they will set that option in their profile as soon as it is important to THEM, ie, the moment a student asks them (or with a ping from you to the coaches in Slack).

alicetragedy:
I agree with Maria: if someone has joined the Teams App once and showed interest in coaching (e.g. in 2014) they might not go back to the Teams app if they've lost interest to say that they will not be coaching anymore and check a box. 
On the other hand, someone interested in coaching will be logging into the teams app again, and would be able to check said box.

[...]

PR #849, opened 2017-09-05
Community/seasoning and filters

branquinhoaa:
Related issue #594 

## Implements "seasoning": 
 - When an orga member chooses, he can go to '/orga/users/info' and click in reset availability. Then, all the coaches availability will be reset (set to false)

reset

 - In the user profile edit, the coaches will have a checkbox to signalize her "availability" for the upcoming season.

users_edit

 - The coaches' availability will be shown in coaches' profile

user_show

## Enhance the filter/search possibilities
 - filter by availability by location and by the availability 
* We add a column for availability and we want to know what do you think about it

rgsoc teams app 12


app/controllers/community_controller.rb
```diff
@@ -9,12 +9,17 @@ def index
...
     }
+
+    @countries = User.pluck(:country)
     @users = User.ordered(params[:sort],params[:direction])
         .group('users.id').with_all_associations_joined
     @users = @users.with_assigned_roles if Time.now.utc > (current_season.starts_at || Date.new)
     @users = @users.with_role(params[:role]) if params[:role].present? && params[:role] != 'all'
     @users = @users.with_interest(params[:interest]) if params[:interest].present? && params[:interest] != 'all'
-    @users = Kaminari.paginate_array(@users.search(params[:search])).page(params[:page])
+    @users = @users.as_coach_availability if params[:availability].present?
```
carpodaster: This @users assignment is getting very long. Could you move this to a private method?

config/routes.rb
```diff
@@ -73,8 +73,10 @@
 
   get 'orga/users/info', to: 'orga/users_info#index', as: :orga_users_info
   patch 'orga/seasons/switch_phase', to: 'orga/seasons#switch_phase', as: :switch_phase
+  post 'orga/community/reset_user_availability', to: 'orga/community#reset_user_availability'
```
carpodaster: From a RESTful perspective, this would be more of a `PUT` request … I think 🤔

app/models/user.rb
```diff
@@ -87,6 +87,10 @@ def supervisor
       where(name: 'supervisor')
     end
 
+    def coach
```
carpodaster: This could be a good time to refactor all these similar methods (except the `admin` one) with some meta programming. Ping me if you need any pointers.

app/models/user/availability_switcher.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
```
carpodaster: I would remove this empty line

app/views/users/_roles.html.slim
```diff
@@ -7,5 +7,9 @@
       label.radio
         = radio_button_tag :role, role == 'all' ? '' : role, params[:role] == role
         = role.capitalize
-
+    h4 Coach availability
+    label.radio
+      = radio_button_tag :availability, true
```
carpodaster: I'm not sure about the radio button. 1st, you cannot uncheck it. 2nd, I would imagine that a call-to-action button is better: a button "I am available for the upcoming season" with a direct `PUT` action being triggered (ideally via AJAX so that we don't get a page reload)

app/views/users/_roles.html.slim
```diff
@@ -7,5 +7,9 @@
...
+      | Availability in the upcoming season
```
carpodaster: As for the term "upcoming": shall we replace it with the actual year? (`Season#name`)

app/controllers/orga/community_controller.rb
```diff
@@ -0,0 +1,10 @@
+class Orga::CommunityController < Orga::BaseController
+ 
+  def reset_user_availability
+    if User::AvailabilitySwitcher.reset
+      redirect_to orga_users_info_path, flash: { notice: 'Coaches were reset with success.' }
```
carpodaster: I always try to reduce multiple exits to a single statement at the end of a method/function (unless it's a [guard clause](https://devblast.com/b/what-are-guard-clauses) in its first line). The `redirect_to` acts as a method exit in a controller. You could use the conditional branches to set the `flash[key]` messages and then redirect to `orga_users_info_path` without giving a flash kwarg _after_ the conditional block.

app/views/users/_roles.html.slim
branquinhoaa: I think you are talking about the checkbox btn that refers to coach availability update. 
This view that you commented above belongs to the filter page.
I removed the checkbox for availability and did an ajax call. I am not sure if that is what you mean...

availability_btn


availability_label

app/models/user.rb
branquinhoaa: ok, this one we have to study a little bit but we will add the refactor soon. :dancer:

branquinhoaa:
@carpodaster the corrections were made! :+1: Now, @mkalininait  just need to confirm that user#interest will be used instead coach role. Then we can update the code!! :dancer:

app/views/users/_roles.html.slim
carpodaster: Yes, this is exactly what I meant <3

app/models/user.rb
carpodaster: 👍

app/models/user/availability_switcher.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+class User::AvailabilitySwitcher
+  def self.reset
```
carpodaster: Two thoughts on this class: 

1. I suggest naming a method like this with a bang (`!`) at the end: `reset!` It's a Ruby convention to name "dangerous" method like that (oftentimes methods have a non-bang and a bang version; a "less-dangerous" non-bang version doesn't really apply here but have a look at the stdlib `String` documentation for a couple of good examples on that).
2. Since the name of the class ends in _Switcher_, it is suggested that there is a reverse action (that is: switching back and forth between two states). The sole action in this class only sets all values to `false`, though. It may sound like nit-picking, but naming is very important (if not amongst the most important things): I suggest naming  this class `User::ResetsAvailability` or something along the lines of the. In order to not re-iterate the name of the class in the one public API method (which would look like `User::ResetsAvailability.reset!`) I suggest the common interface for lamba-/proc style callables: `.call` (this time without a bang – confusing, I know :wink:)

app/views/community/index.html.slim
```diff
@@ -23,6 +23,7 @@ table.users.table.table-striped.table-bordered.table-condensed.table-hover
       th = sortable :country
       th = sortable :location, User.human_attribute_name(:location)
       th Time now
+      th Availiability
```
carpodaster: Typo

spec/models/user/availability_switcher_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+  describe '.reset' do
+    let!(:coach_available) { create :coach, :available }
+    let!(:user_available) { create :user, :available }
+    before do 
+      User::AvailabilitySwitcher.reset
+    end
+
+    context 'when an user is a coach' do
+      it 'reset the availability to false' do
+        expect(coach_available.reload.availability).to be false
```
carpodaster: A better way to test a *state change* is to use the `expect { … }.to change { … }` RSpec macro.

Example:

```ruby
expect { User::AvailabilitySwitcher.reset }
  .to change { coach_available.reload }
  .from(true).to(false)
end
```

You don't need the `before` block that way.

spec/models/user_spec.rb
```diff
@@ -63,6 +63,9 @@
 
   describe 'scopes' do
     context 'role scopes' do
+      roles_except_organizer = User::ROLES
```
carpodaster: I'm undecided whether it's good to have the test run use a MP pattern as well. I feel that it would be better for this test to catch regressions by defining the list of roles as a static list, not re-using `User::ROLES`

app/assets/javascripts/user_form.js
```diff
@@ -7,6 +7,29 @@ $(function () {
      $('.form-group.user_company_info').toggle(this.checked);
   }).change(); //ensure visible state matches initially
 
+
+   //update coach availability
+
+   $(function () {
+      var $update_availability = $('[data-behavior="update_availability"]');
+      $update_availability.on('click',function(){
+        var url = "/users/current/update_availability";
```
michaelem: I think hardcoding routes in javascript is somewhat of an issue since it can lead to javascript breaking when those routes changes. This is especially troublesome if these things are not covered by system tests since they then will fail in production. I think the right way to pass the url in this case would probably be to have to url in an additional data field in the element that has the data-behaviour field.

mkalininait:
@carpodaster 
> I'm more and more convinced that just looking at coaches (= users who have been assigned a coach role before) is not the the way to go. This would ignore new users to the teams app that are willing to coach but have never been a coach before. Or RGSoC Alumna. I feel like we need to consider interested_in as well.

I agree! But I'm actually also not sure that we need to know availability of people who have the 'coach' role, but are not currently interested in coaching.

@branquinhoaa I didn't test the PR yet, but I will do after you make this change covering ppl who are *interested in coaching*.

branquinhoaa:
@carpodaster  @mkalininait  About the availability field: it is now available to all user interested_in coaching (not just coaches). 
@carpodaster I removed the AvalabilitySwitcher class because it was useless (now I call a user class method availability_reset!  when I want to reset the user's availability).

klappradla:
Hey @branquinhoaa what's the state of this? Did you already address everything @carpodaster and @mkalininait mentioned and want me to have another look at it 👀  ???

branquinhoaa:
Hey @klappradla! Yeah, I have already finished with the @carpodaster suggestions. 
It would be nice if you can take a look.

klappradla:
Alright, gonna have a look when I'm home later today ✌️

app/assets/javascripts/user_form.js
```diff
@@ -7,6 +7,29 @@ $(function () {
...
+      var $update_availability = $('[data-behavior="update_availability"]');
+      $update_availability.on('click',function(){
+        var url =$(this).data("url");
+        updateAvailability(url);
+      });
+    });
+
+   function updateAvailability(url) {
+      $.ajax({
+        type:     'PUT',
```
klappradla: Also: I think it's the wrong key. It should be 
```js
method: "PUT"
```
since this is about type of the request.

Maybe you were referring to `dataType`??? It's optional here, jQuery will do an intelligent guess, but you could also deliberately set it to whatever you need. Would be
```js
dataType: "script"
```
here I suppose.

app/models/community_users.rb
```diff
@@ -0,0 +1,32 @@
...
+    @users         = users
+    @role          = params[:role].presence
+    @interest      = params[:interest].presence
+    @availability  = params[:availability].presence
+    @search        = params[:search].presence
+    @location      = params[:location].presence
+    @page          = params[:page].presence
+  end
+
+  def all
```
klappradla: Can you come up with tests for this? You can stub out all the class methods on `User`, I think it would just be helpful to ensure the contract, so that if maybe in the future someone removes one of these class methods, this is caught by our tests.

app/models/user.rb
```diff
@@ -159,9 +155,21 @@ def with_interest(interest)
...
+
+    def as_coach_availability
+      where(availability: true)
+    end
+
     def immutable_attributes
       [:github_handle]
     end
+
+    def reset_availability!
```
klappradla: Can you add tests for this? It's a method with rather large impact... Would be good to know if it's really (and only) doing what it should 😉

app/controllers/orga/community_controller.rb
```diff
@@ -0,0 +1,7 @@
+class Orga::CommunityController < Orga::BaseController
+
+  def reset_user_availability
```
klappradla: Can you add tests for this?

app/views/users/_roles.html.slim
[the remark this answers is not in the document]
branquinhoaa: In fact it is always unchecked. When you check it, the search starts. I will remove this useless true.

app/views/users/edit.html.slim
```diff
@@ -50,6 +50,16 @@
   = f.input :company_name, label: 'Company name'
   = f.input :company_info, label: 'Company info', hint: 'Where is the office, how much space can you provide, etc.'
   = f.input :bio, label: 'About you', hint: 'You can use basic <a href="http://en.wikipedia.org/wiki/Markdown" target="_blank">Markdown</a> here.'.html_safe
+
+  - if @user.interested_in_coaching?
+    h3.page-header Availability for the next season:
```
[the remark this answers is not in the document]
branquinhoaa: That is so true!! :joy_cat: I did not think about it...

app/controllers/users_controller.rb
```diff
@@ -49,6 +49,10 @@ def update
     end
   end
 
+  def update_availability
+    current_user.update(availability: true)
```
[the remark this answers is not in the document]
branquinhoaa: Yeah, that is exactly what I understand by @carpodaster comment. The user can set his availability, but he cannot give up.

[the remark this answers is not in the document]
klappradla: Alright, perfect! Just wanted to make sure.

mkalininait:
I've tested the PR and have several thoughts.

**User's page (show and edit):**
1) It would be cool if the dependency between "Interested in helping as a coach" and "Availability" on User's page was clearer. E.g. when I set a check that I'm interested to be a coach on the edit page, I see the "I'm available in the next season" checkbox in real time somewhere below. Now this checkbox appears only after I reopen the edit form and it's called "Availability for the next season" which doesn't indicate any relation with the coach status. Same with the title "Availability" on the show page. It's a bit confusing. I think we should at least rename "availability" to "coach availability" everywhere.
2) Apparently I (as a person interested in coaching) can't set my availability status to "Not available" myself. I think it's a necessary feature.
3) Also, once I press the button "I'm available in the next season", I see another button "Successful updated" which I can't press. Why is it a button at all? I'd expect to see a button "I'm NOT available in the next season" or something like that. And if we want to notify that the status is successfully updated, I'd show the message "Successfully updated", not "Successful updated".
4) More about naming: on the edit form the title is "Availability **for** the next season:", but the button and the label are "I am available **in** the next season". At the same time on the show page I see "I am available **for** this season!".  I think we should use "in" (not "for") and "next" (not "this") consistently.

**Resetting :**
1) When I reset coach statuses on the admin panel I see this notification:  "Coaches were reset with success." It's better to say: "Coach availability was reset with success." or something like this.

**Community page:**
1) "Availability in the 2017 season": Why is it a radiobutton? How to reset this filter? What triggers changing the year (say, 2017 to 2018)? I think it should be a checkbox called "Coach availability in the next season". 
2) The country filter doesn't work correctly for me: it shows repeating values.
screen shot 2017-10-03 at 15 20 57

cc @branquinhoaa The summer of Code is kinda over already, so I'm not sure if I should mention you here or not :)

branquinhoaasimbiose:
@mkalininait Of course you should mentioning me here!! I will keep working on Teams app issues :dancer:

mkalininait:
@branquinhoaa Do you want to add more commits or is the PR ready for testing?

branquinhoaa:
@mkalininait Hi! @juuh42dias will fix some things and then I ask you to review ok?

klappradla:
Hey @branquinhoaa 👋 

I know your Summer of Code is over 😿 

But I just wanted to check what's the state of this PR 🕵️ 
Do you plan to back back to it, or should I take it over to add the last bits to it?

branquinhoaa:
@klappradla Hey!! I would like to finish this pr. I will do the last alterations now and then I ask your review, ok? thank you!!!

klappradla:
Awesome 💚 

And yeah, just ping me again.

branquinhoaa:
@klappradla It is ready to the new review!! Thank you

app/assets/stylesheets/global.sass
```diff
@@ -67,6 +67,12 @@ html, body
       &:last-child:after
         content: none
 
+#availability
```
klappradla: This is never used (or at least I cannot find it in this PR 🙈 )

So can you please remove this?

app/controllers/community_controller.rb
```diff
@@ -9,12 +9,12 @@ def index
       mentoring:  'Helping as a Mentor',
       organizing: 'Helping as an Organizer'
     }
+
+    @countries = User.pluck(:country).uniq
     @users = User.ordered(params[:sort], params[:direction])
```
klappradla: I think it would be a bit more readable if only those things later to be used in the view would be instantiated as instance variables. So how about having:
```ruby
users = User.order...
communtiy = CommunitUsers.new(..., users)
@users = community.all
```

What do you think?

app/controllers/users_controller.rb
```diff
@@ -1,7 +1,7 @@
 class UsersController < ApplicationController
   before_action :set_user, only: [:show, :edit, :update, :destroy, :impersonate]
 
-  load_and_authorize_resource except: [:index, :show, :impersonate, :stop_impersonating]
+  load_and_authorize_resource except: [:index, :show, :impersonate, :stop_impersonating, :set_available]
```
klappradla: Since this list is getting quite long now, why not "reverse" it and use `only` instead?
```ruby
load_and_authorize_resource only: %i[new create edit update destroy]
```

config/routes.rb
```diff
@@ -73,6 +73,8 @@
 
   get 'orga/users/info', to: 'orga/users_info#index', as: :orga_users_info
   patch 'orga/seasons/switch_phase', to: 'orga/seasons#switch_phase', as: :switch_phase
+  put 'orga/community/reset_user_availability', to: 'orga/community#reset_user_availability'
```
klappradla: Can you please move this into the `orga` namespaces`?  It will have the `/orga` prefix automatically.

config/routes.rb
```diff
@@ -73,6 +73,8 @@
...
+  put 'users/current/update_availability', to: 'users#set_available'
```
klappradla: Can you please move this to the `users` resource and make it a member action? It's really hard to find otherwise, given the size of the `routes` file 🙈

db/migrate/20170901163452_add_availability_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddAvailabilityToUsers < ActiveRecord::Migration[5.1]
+  def change
+    add_column :users, :availability, :boolean
```
klappradla: Hm, personal preference, but I would suggest to **always** put a `null: false` and a default value (`false` in this case) to a boolean value. You'll otherwise end up with a 3 state boolean: `null, true, false` - which can be a potential source of problems later 😉

db/migrate/20170905222833_add_default_value_to_availability.rb
```diff
@@ -0,0 +1,5 @@
+class AddDefaultValueToAvailability < ActiveRecord::Migration[5.1]
+  def change
+    change_column_default :users, :availability, false
```
klappradla: Why are you using a separate migration for this? I'd suggest to just do it in the migration where the column is added.

spec/controllers/orga/community_controller_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+RSpec.describe Orga::CommunityController do
+  render_views
+
+  context 'with admin logged in' do
+    include_context 'with admin logged in'
+    describe 'PUT reset_user_availability' do
+
+      it 'reset availability for users interested in coaching' do
+        put :reset_user_availability
+        expect(response).to redirect_to (orga_users_info_path)
```
klappradla: I'm missing an expectation on the side effects of this request here 👮  My suggestion would be to have something like:
```ruby
expect { put :reset_user_availablity }.to change { user.reload.availablity }.to(false)
```

spec/models/community_users.rb
```diff
@@ -0,0 +1,84 @@
+require 'spec_helper'
+
+describe CommunityUsers do
+  let(:student) { create(:student) }
+  let(:student_looking_for_coach) { create(:student, :interested_in_coaching) }
+  let(:coach) { create(:coach, :available) }
+  describe '#all' do
```
klappradla: Can you please add a blank line before the `describe` block? Just easier to read.

spec/models/community_users.rb
```diff
@@ -0,0 +1,84 @@
...
+    before do
+      @params = {
```
klappradla: I don't really see the point of switching from `lets` to instance vars here 🙈 

Can you please convert them to `let()` statements?

This also does not need to be in a before block from what I see.

spec/controllers/orga/community_controller_spec.rb
```diff
@@ -0,0 +1,17 @@
+require 'spec_helper'
+
+RSpec.describe Orga::CommunityController do
+  render_views
+
+  context 'with admin logged in' do
+    include_context 'with admin logged in'
+    describe 'PUT reset_user_availability' do
+
+      it 'reset availability for users interested in coaching' do
```
klappradla: Typo
> it 'resets the availablity

spec/models/community_users.rb
```diff
@@ -0,0 +1,84 @@
...
+        }
+
+      @users = User.ordered(@params[:sort], @params[:direction])
+                   .group('users.id')
+                   .with_all_associations_joined
+
+      @community = CommunityUsers.new(@params, @users)
+    end
+
+    context 'return filtered users' do
```
klappradla: Since there's only one `context` I don't see the point in having it 😉  it just increases the indentation level in this case and does not really give more information.

spec/models/community_users.rb
```diff
@@ -0,0 +1,84 @@
...
+      describe 'with params presence' do
```
klappradla: This should be a `context` instead. On a technical level, this does of course make no difference, but things beginning with "with" and "when" are usually `context` blocks and make reading easier.

spec/models/community_users.rb
```diff
@@ -0,0 +1,84 @@
...
+        it 'return user with role specified' do
+          @params[:role] = student.roles.first.name
+
+          community = @community
```
klappradla: unnecessary assignment

spec/models/community_users.rb
```diff
@@ -0,0 +1,84 @@
...
+      @users = User.ordered(@params[:sort], @params[:direction])
+                   .group('users.id')
+                   .with_all_associations_joined
+
+      @community = CommunityUsers.new(@params, @users)
+    end
+
+    context 'return filtered users' do
+      describe 'with params presence' do
+        it 'return user with role specified' do
```
klappradla: I'd reword to
> it 'returns only users with the specified role'

Same for the following examples.

spec/models/community_users.rb
```diff
@@ -0,0 +1,84 @@
...
+        it 'return user by location' do
+          @params[:location] = student.country
+
+          community = @community
+          expect(community.all).to match_array student
+        end
+      end
+    end
+
+    context 'private methods' do
```
klappradla: It's best practice to not test private methods. Unit tests would just cover the public interface and through this also indirectly test the internals.

Could you convert these tests so that you instead test the outcome of these scenarios on the `all` method?

spec/models/user_spec.rb
```diff
@@ -63,6 +63,14 @@
 
   describe 'scopes' do
     context 'role scopes' do
+      roles_except_organizer = [
```
klappradla: I would suggest to couple this closer to the actual implementation. Right now, changing the `ROLES` constant in the users model would probably not even break this test - but it should 😉 

So how about
```ruby
roles-exept_organizer = User::ROLES.except('organizer)
```

spec/models/user_spec.rb
```diff
@@ -126,6 +126,20 @@
         end
       end
 
+      describe '.with_location' do
+        it 'returns users matching one out of many location' do
```
klappradla: Typo: "locations"

spec/models/user_spec.rb
```diff
@@ -391,6 +405,27 @@
     end
   end
 
+  describe '.reset_availability!' do
+    let!(:user) { create(:user, :available) }
+    let!(:user_interested_in_coaching) { create(:user, :available, :interested_in_coaching) }
+
+    before do
+      User.reset_availability!
```
klappradla: I feel like doing this in a before block reduces the possibilities to test the actual behavior of the method.

I would suggest to rather test the "change" and its impact:
```ruby
expect { User.reset_availablity! }.to change { user.reload.availablity }.from(true).to(false)
```

F3PiX:
Very cool feature, however, it doesn't look like this PR is easy to cherrypick-and-live-on. I am going to close this; please reopen if you think it should stay here. 
Maybe we can break down the original issue in smaller steps, and make them available as beginner friendly issues.

PR #850, opened 2017-09-08
Modal for conference creation in team edit view

LuisaAPF:
Related issue #809 

 - Improves ux for adding a new conference on the teams/edit page. The form is now shown as a modal, instead of a new page and it is submitted via ajax. After submission, the added conference appears selected on the primary choice element on the team edit form.
- Screenshot showing validations after trying to submit empty form.
image


app/assets/javascripts/conference_form.js
```diff
@@ -0,0 +1,23 @@
+$(function () {
+  $('#new_conference').on('ajax:success',
```
klappradla: Personal preference, but what do you think about using a data attribute instead of the ID for attaching the behavior to this element? It could be something like `data-behavior="new-conference-form"` 🤔 
I personally prefer to leave classes and ids for styling elements and use the more powerful (you can pass down data :v:) data attributes for adding certain behavior.
What do you think?

app/assets/javascripts/conference_form.js
```diff
@@ -0,0 +1,23 @@
...
+      selected.prop('selected', 'selected');
+      selected.val(conference.id);
+      selected.text(conference.name);
+    }
+  );
+  $('#new_conference').on('ajax:error',
+    function(e, xhr) {
+      if(xhr.status == 422) {
+        var errors = JSON.parse(xhr.responseText);
+        $('#new_conference_modal_errors').removeClass('hidden').html('<ul></ul>');
```
klappradla: 👍

app/assets/stylesheets/global.sass
```diff
@@ -329,3 +329,7 @@ h1.header, h2.header
   h5
     font-style: italic
     margin-bottom: 0
+
+.modal
+  h2
+    display: inline
```
klappradla: I may be missing something, but can you explain why this is necessary?

app/views/conferences/_form.html.slim
```diff
@@ -0,0 +1,32 @@
+.modal.fade id='new_conference_modal' tabindex='-1' role='dialog'
+  .modal-dialog role='document'
+    .modal-content
+      = simple_nested_form_for @conference, remote: true do |f|
```
klappradla: How about ditching `@conference` and using `Conference.new` here instead?
I may be overlooking something, but I for now don't see a benefit in assigning another view variable in the controller for this modal form if it's just a `Conference.new` directly passed into the form method.

app/controllers/teams_controller.rb
```diff
@@ -27,6 +27,7 @@ def new
   def edit
     @team.sources.build(kind: 'blog') unless @team.sources.any?
     @conferences = conference_list
+    @conference = Conference.new
```
klappradla: I'd prefer to now assign another view variable, see my comment in the form partial 👇

app/controllers/teams_controller.rb
```diff
@@ -47,6 +48,7 @@ def create
 
   def update
     @conferences = conference_list
+    @conference = Conference.new
```
klappradla: I'd prefer to now assign another view variable, see my comment in the form partial 👇

app/views/teams/_form.html.slim
```diff
@@ -74,4 +74,4 @@
   .actions
     ul.list-inline
       li  = f.submit 'Save', class: 'btn btn-success'
-      li = link_to 'Cancel', @team, class: 'btn btn-default'
\ No newline at end of file
+      li = link_to 'Cancel', @team, class: 'btn btn-default'
```
klappradla: 👍

spec/controllers/conferences_controller_spec.rb
```diff
@@ -54,4 +54,16 @@
       end
     end
   end
+
+  describe 'POST create' do
+    context 'unauthenticated users' do
+      let(:params) { { name: 'name', country: 'country', region: 'region', location: 'location', city: 'city', season_id:'1', url: 'www.conference.com' } }
+      it 'cannot create new conference' do
```
klappradla: Could you please leave a blank line before an example block? It's kinda best practise in the Ruby community and in my opinion really increases the readablity.

spec/controllers/conferences_controller_spec.rb
```diff
@@ -54,4 +54,16 @@
       end
     end
   end
+
+  describe 'POST create' do
+    context 'unauthenticated users' do
+      let(:params) { { name: 'name', country: 'country', region: 'region', location: 'location', city: 'city', season_id:'1', url: 'www.conference.com' } }
```
klappradla: What's the point in assigning this huge params hash if you in the end only want to test whether or not a user is signed in? I'd suggest to shorten it to:
```ruby
it "does not create a new conference record" do
  expect { post :create, params: conference: {} }.not_to change { Conference.count }
end
```

spec/controllers/conferences_controller_spec.rb
```diff
@@ -54,4 +54,16 @@
       end
     end
   end
+
+  describe 'POST create' do
+    context 'unauthenticated users' do
```
klappradla: Only a small note on "naming things"™: in order to end up with an almost human readable RSpec example, it makes sense to formulate contexts starting with "with" or "when", e.g. "with an unauthenticated user" or "without current_user".
Would be cool if you could come up with a naming like this, it in my opinion also increases readability :v:

spec/controllers/conferences_controller_spec.rb
```diff
@@ -54,4 +54,16 @@
       end
     end
   end
+
+  describe 'POST create' do
```
klappradla: Do you think you can come up with a positive example as well? At best testing both, the xhr and the http request.

app/controllers/conferences_controller.rb
```diff
@@ -18,10 +18,14 @@ def create
...
     @conference.gid = generate_gid(team)
 
-    if @conference.save
-      redirect_to edit_team_path(team), notice: 'Conference was successfully created.'
-    else
-      render action: :new
+    respond_to do |format|
+      if @conference.save
+        format.html { redirect_to edit_team_path(team), notice: 'Conference was successfully created.' }
+        format.js { render json: [@conference] }
```
klappradla: Why do you return an array instead of just one element? This being a `conferences/create` endpoint, I'd expect only a single element being returned:
```ruby
render json: @conference
```

app/assets/javascripts/conference_form.js
```diff
@@ -0,0 +1,23 @@
+$(function () {
+  $('#new_conference').on('ajax:success',
+    function(e, data) {
+      var conference = JSON.parse(data)[0];
```
klappradla: See my comment about the return value from the controller in the `ConferencesController`. In my opinion, it's semantically confusing to return an array, plus I think
```js
JSON.parse(data)
```
also makes it more obvious that we're dealing with a single record.

app/assets/javascripts/conference_form.js
LuisaAPF: I can do this, @klappradla. No problem!

app/controllers/conferences_controller.rb
LuisaAPF: I know this is odd, but if I don't return an array, the ajax success function is not even called.

app/assets/javascripts/conference_form.js
LuisaAPF: See my comment in the end

app/assets/stylesheets/global.sass
LuisaAPF: This is necessary because the default display of a h2 element is a block. In this case, the modal title and the close button get misaligned.

app/views/conferences/_form.html.slim
LuisaAPF: Agreed!

spec/controllers/conferences_controller_spec.rb
LuisaAPF: I thought this would isolate the behavior I'm looking at, so that the test won't fail because the parameters are invalid, but because the user is not logged in.

spec/controllers/conferences_controller_spec.rb
LuisaAPF: I tried using xhr post, but I kept getting a no method error... Maybe this should be a separate issue?

app/controllers/conferences_controller.rb
klappradla: Hm, looks more like a general format / parsing confusion here... 🤔 

Have you tried
```ruby
format.js { render json: @conference.to_json }
```

Plus: I'm not 100% sure if `remote: true` will automatically make the request a `format: script`one. It could also be a `JSON` request. Have you checked this?

spec/controllers/conferences_controller_spec.rb
klappradla: Can you post a snipped of the code you tried? Background: the format changed from Rails 4 to 5 -> xhr is now simply a parameter of the `post` method:
```ruby
post :create, params: params, xhr: true
```

spec/controllers/conferences_controller_spec.rb
klappradla: Valid point!

I don't know the `conference` factory from the top of my head, but eventually `attributes_for :conference` would be shorter and saver 🤔 

Maybe the test would be also be more meaningful if you'd also look at the response status? As in: if it's a `401`, we know the request was no authorized.

app/controllers/conferences_controller.rb
LuisaAPF: I tried .to_json but it didn't work. Also, remote:true looks for a js format. It doesn't respond to json.

klappradla: Sorry, didn't make myself clear in the previous comment 🙈 

You can set the content type yourself in the form `data: { type: :json }` and this may eventually fix your problem. For jQuery, `script` and `json` are two different things and it parses it differently. From what I remember for `script` / `js` it expects plain text, for `json` something different.

However, if you can fix it - yay - if not, we can also do it later ✌️

spec/controllers/conferences_controller_spec.rb
LuisaAPF: Thanks! It worked!

app/controllers/conferences_controller.rb
LuisaAPF: You're right. The data-type=json fixed it.

klappradla: Ace 🎉  Just ping me again when you're done with all the changes.

LuisaAPF: I'm done, @klappradla

spec/controllers/conferences_controller_spec.rb
```diff
@@ -54,4 +54,58 @@
...
+        expect(response).to have_http_status(401)
+      end
+
+    end
+
+    context 'with user logged in' do
+      let(:student) { FactoryGirl.create(:student) }
+      let(:params) { attributes_for :conference }
+
+      before :each do
```
klappradla: Minor thing, but can you please remove the `:each` part? It's the default behavior and considered bad practise to add it 😉 
```ruby
before { sign_in student }
```
does the job

spec/controllers/conferences_controller_spec.rb
```diff
@@ -54,4 +54,58 @@
...
+
+      it 'cannot create new conference' do
+        expect {
+          post :create,
+          params: { conference: params },
+          xhr: true
+        }.not_to change { Conference.count }
+      end
+
+      it 'should return a 401 response status' do
```
klappradla: 👍

spec/controllers/conferences_controller_spec.rb
```diff
@@ -54,4 +54,58 @@
       end
     end
   end
+
+  describe 'POST create' do
+    context 'with an unauthenticated user' do
+      let(:params) { attributes_for :conference }
```
klappradla: Minor naming thing, but technically speaking these aren't the "params"... How about `conference_attrs`?

spec/controllers/conferences_controller_spec.rb
```diff
@@ -54,4 +54,58 @@
...
+
+      it 'cannot create new conference' do
+        expect {
+          post :create,
+          params: { conference: params }
+        }.not_to change { Conference.count }
+      end
+
+      it 'cannot create new conference' do
```
klappradla: Having two examples named identically and internally only diffing by two words makes reading a bit hard here 🙈 
How about adding some more information to the example description or wrapping it into a "context" that says it's about xhr requests now?

spec/controllers/conferences_controller_spec.rb
```diff
@@ -54,4 +54,58 @@
...
+      it 'should return a 401 response status' do
+        post :create, params: { conference: params }, xhr: true
+        expect(response).to have_http_status(401)
+      end
+
+    end
+
+    context 'with user logged in' do
+      let(:student) { FactoryGirl.create(:student) }
+      let(:params) { attributes_for :conference }
```
klappradla: Same here for naming: these aren't the "params". Maybe `conference_attrs` would be a better choice, or moving it into something like:
```ruby
let(:params) { { conference: attributes_for(:conference) } }
```

config/initializers/simple_form.rb
```diff
@@ -118,7 +118,7 @@
   # in this configuration, which is recommended due to some quirks from different browsers.
   # To stop SimpleForm from generating the novalidate option, enabling the HTML5 validations,
   # change this configuration to true.
-  config.browser_validations = false
+  config.browser_validations = true
```
klappradla: Yay 🎉

app/views/conferences/_form_modal.html.slim
```diff
@@ -0,0 +1,32 @@
...
+          p
+           h3 Conference location
+          fieldset
+            = f.input :country, required: true, as: :country, prompt: "Conference country", wrapper_html: { class: 'col-md-12' }
+            = f.input :region, required: true, collection: Conference::REGION_LIST, prompt: 'Conference region', wrapper_html: { class: 'col-md-12' }
+            = f.input :city, required: true, placeholder: 'Conference city', wrapper_html: { class: 'col-md-12' }
+
+        .modal-footer
+          button.btn.btn-default type='button' data-dismiss='modal' Cancel
+          = f.submit 'Save', class: 'btn btn-success'
```
klappradla: I did not yet try this in the browser, but is it possible to "DOS-click" the button and submit the form a million times? If so, it would be cool to disable the button after submit ✌️

spec/controllers/conferences_controller_spec.rb
F3PiX: We also have[ shared contexts](https://github.com/rails-girls-summer-of-code/rgsoc-teams/tree/master/spec/support/shared_contexts), for different kind of users. I think you can use them instead of the `before` block

spec/controllers/conferences_controller_spec.rb
LuisaAPF: Hmm.. I don't understand the difference. If I print params, I get a hash with name, url, etc, just like the post parameters.

spec/controllers/conferences_controller_spec.rb
LuisaAPF: You mean it's a bad practice to use :each because I just have one context after it, or it should not be used at all?

klappradla: Sorry for causing confusion 🙈  - did not make myself clear. It's only an implicit vs. explicit thing, some linters will tell you to skip the `(:each)` part, because it's the default behavior. That's all.
It's uncommon for me to see, so I suggested to make the `before(:each)` a `before`, but it does not have any impact on the code itself.

Plus: it's of course also totally fine if you leave it like this :v:

For the shared contexts: yes, @F3PiX is right, you could use them here. But since you're setup is super minimal, I don't see a real benefit in doing so. Would probably just make things harder to read.

spec/controllers/conferences_controller_spec.rb
klappradla: Sorry, it's just a naming thing 😉 
The "params" look like this: `{ conference: { name: "bla", url: "yay.de" ... } }`
What you assign as params here is only the `{ name: "bla", etc }` part.

But again, this is only a question of naming things, has nothing to do with the test itself. The code work fine.

spec/controllers/conferences_controller_spec.rb
LuisaAPF: I'm thinking if it makes sense at all to keep the tests for the http submission in this file. For what I understand, now all files regarding the creation of conferences outside the scope of teams/edit are in the orga folder, meaning that only organizers should have access to them. Am I correct, @klappradla?

klappradla: You're right 💪  makes sense!

I could have seen it earlier, but this constant switching between programming languages and projects today make my brain kinda fuzzy 👻

mkalininait:
Hey @LuisaAPF Thank you for the PR! I've tested it. It looks good to me. There are two flaws that I noticed:
1) After I added a conference, it's automatically set as my primary choice. I think it's not very convenient, because it can override my previous primary choice. And not transparent, it's easy to not notice this and save changes. Besides, this year's experience shows that students add new conferences and don't necessarily choose them in their own preferences.
2) After I added a conference, I see it in the dropdowns "Primary choice" and "Secondary choice" as not categorized: it doesn't belong to any region and hangs there in the list as the first item.

LuisaAPF:
Hey, @mkalininait, I designed it like this because I thought the students would only suggest a new conference if they didn''t find any in the list that suited them. But, as you said, this doesn't seem to be the case. I'll change it.

mkalininait:
@LuisaAPF I was surprised too :) Thank you!

LuisaAPF:
I'm done with the changes, @mkalininait, @klappradla

app/views/conferences/_form_modal.html.slim
```diff
@@ -0,0 +1,32 @@
...
+      = simple_nested_form_for Conference.new, data: { form: 'new-conference-form', type: "json" }, remote: true do |f|
+        .modal-header
+          h2.modal-title Create a new conference
+          button.close type='button' data-dismiss='modal'
+            span &times;
+
+        .modal-body
+          .alert.alert-danger.hidden data-errors='new-conference-modal-errors'
+
+          p
```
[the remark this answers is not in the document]
LuisaAPF: You're right :grimacing:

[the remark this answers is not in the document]
LuisaAPF: You're right, @klappradla . After I change it, can I merge it myself?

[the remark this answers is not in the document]
klappradla: Yes 💚

F3PiX:
🎉

PR #854, opened 2017-09-13
Refactor Team.by_season and TeamsController#index

carpodaster:
Some refactorings post #814.

Well, strictly not entirely a refactoring PR as I removed the `raise` from `Team.by_season` to just return an empty relation.

### DIFF
--- a/app/models/team.rb
+++ b/app/models/team.rb
@@ -46,9 +46,7 @@ class Team < ActiveRecord::Base
 
   scope :by_season, ->(year_or_season) do
     case year_or_season
-      when Integer
-        joins(:season).where('seasons.name': year_or_season)
-      when String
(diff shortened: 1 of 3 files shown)


app/models/team.rb
```diff
@@ -46,16 +46,14 @@ class Team < ActiveRecord::Base
...
-        where(season: year_or_season)
-      else
-        raise
-      end
+    when Integer, String
+      joins(:season).where('seasons.name': year_or_season)
+    when Season
+      where(season: year_or_season)
+    else
+      none
```
klappradla: 🎉  haven't seen this before, but it makes HUGE sense here 👍  (`[]` would not allow to chain scopes...)

Issue #853, opened 2017-09-12
Blog posts uploaded as daily logs in supervisor dashboard for team_id 338


linberg:
In the supervisor dashboard (id: 1306) for team_id 338, blog posts from other teams have been uploaded as daily logs on 7 occasions (11.08., 15.08., 04.09., 06.09., 08.09., 09.09. and 11.09.) always at 02:00 CEST. Please see screenshot.
![screen shot 2017-09-12 at 17 31 27](https://user-images.githubusercontent.com/16653138/30337384-6aa8cebe-97e8-11e7-81d4-a484c44da2b4.png)

EDIT (Maud) In the screenshot, only the posts with the photo avatar are really by team id 338, the other posts are feed entries from other teams.

F3PiX:
Thanks for reporting this @linberg. I'll look into it.

F3PiX:
The supervisor dashboard redirects to the activity page to show the status updates of the team. 
The problem occurs with the activity page. 

It looks like there are two problems here.  
1) The feed entries are trying to load pictures on AWS S3 and don't get access.
![feed entry aws](https://user-images.githubusercontent.com/6314687/30393728-4d761dd6-98c1-11e7-9154-a9306e399a15.png)


 2) Team id 338 is the _only_ team where the activities try to get access to other teams' feed entries 🤔 
![activities team 338](https://user-images.githubusercontent.com/6314687/30393905-cca0d448-98c1-11e7-8734-6630a95ead63.png)



I did not figure out yet what the source of the problem(s) are. Tips that gives me a direction are welcome!
( @carpodaster , @klappradla  )

F3PiX:
@linberg Do you recall if this problem existed before last week? Or did the post from Sept 8 and Sept 9 appear at once this week?

linberg:
First time this issue occurred was with the first blog post, so August 11th. There was another bug with this team as well during the application process, their application had showed up under their last year's team name (as they had also applied then). Also at the beginning of the program their daily logs were also going under their last year's team name, I unfortunately don't know the ID, but I think @mkalininait had looked into that issue.

klappradla:
I don't have time to look into this right now @F3PiX, but the S3 issue may have something to do with me updating the gem...? It's the only touchpoint recent commits had with S3 as far as I remember.
I can just revert the commit and see if we're back at normal with this.

F3PiX:
[...]

PR #856, opened 2017-09-14
Changed the gemfile instead of reverting commit

F3PiX:
Related issue #853 part 1: Quickfix: reverting the upgrade to avoid problems with the AWS gem v3.

Quickfix: downgrade the AWS v3 gem back to v2 to avoid problems with fetching images.


F3PiX:
@klappradla I did it the 🐔 way. :-)

Issue #845, opened 2017-08-30
Discussion: static code analysis via Rubocop / Codeclimate


carpodaster:
Even though we currently state in our [CONTRIBUTION.md](https://github.com/rails-girls-summer-of-code/rgsoc-teams/blob/master/CONTRIBUTING.md) that we don't automatically lint code, it has been suggested by @F3PiX that we consider adding a couple of simple rules.

This issue exists to discuss how everyone is feeling about this and what kind of static code analysis would make sense.

carpodaster:
I always thought that rubocop rules aren't very inviting for people new to the Ruby programming language. I may be wrong on this one, but I would still like to keep the linting rules to a minimum.

Things that come to mind (opinionated):

* Indentation rules: 2 whitespaces, no extra indentation for private methods
* New hash syntax (`{ foo: :bar }` instead of `{ :foo => :bar }`)
* Aligning `case` and `when` on the same level
* Lines not longer than ~100 characters
* No trailing whitespace
* First-level `describe` should be `RSpec.describe`
* No parentheses for method definitions _without_ args, mandatory parentheses for method definitions _with_ args
* Omit parenthesis on method calls without arguments
* Omit `FactoryGirl` and just use `create :foo`
* No explicit `self`
* No explicit `return` at the end of a method

I was looking at the first half of https://github.com/bbatsov/ruby-style-guide

F3PiX:
FYI : Why I suggested Rubocop:
As discussed on Slack, I OTOH think that rubocop rules can be very accommodating for RubyNubies.
It helps me to avoid easily avoidable mistakes like indentations. It helps me delivering the best PR I can offer. And it helps with consistency between code reviews.

F3PiX:
Great list!!! 🎯

klappradla:
👍 for adding a `rubocop` step to our CI pipeline with some small and basic rules like the ones you described @carpodaster 💚
From my experience, it helps to reduce some efforts on keeping code consistent and I **hugely** agree with @F3PiX, it's an awesome resource for learning some best practises _(the docs are really great)_ - I really liked it when I started out with Ruby.

I would volunteer to set everything up ☝️ and update the guides and readme accordingly one we came to a decision (I'm on a conference next week, so I may have some free time).

klappradla:
Small note on 
> Omit FactoryGirl and just use create :foo

[...]

PR #857, opened 2017-09-16
Static code analysis via Rubocop

klappradla:
Related to the discussion in https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/845, this adds static code analysis via Rubocop as an additional step in the CI pipeline *(a super fast step though, no worries)*.

Please see the commit history and the `.rubocop.yml` file in this PR for the details on which cops are enabled and which I already ran.
Rubocop has a [large list of per-default enabled cops](https://github.com/bbatsov/rubocop/blob/master/config/enabled.yml). Since the number of cops we want to have enabled is rather small, I instead of manually disabling the ones we don't want, disabled all and just added some specific ones. The file is rather large, since I also included the description and link to the styleguide for each cop. I thought it could be more helpful for newcomers or people coming to the list from the contribution guide.

From @carpodaster's initial list of cops, I for now skipped everything related to specs since it would require another gem (I can do this once we agreed on the setup for the app code) and did not enable the one for line length and the frozen string literal comment, since this would basically touch every file in the repo 🙈  (I'd then also do this in a separate PR).

Aside that, I also enabled some very basic cops related to things like trailing whitespaces, indentation, etc. (see the file and the commit history for details).

I also update the Readme and Contribution guide to include all this ☝️


F3PiX:
Wow Max, that is an amazing job! Thank you!! 🎁  
My future self is already immensely grateful. 😛 

I feel you added some great cops (re: Ruby/Rails conventions), and also some that are probably up for debate; for instance one that I don't see as an important thing, and even one that I don't like at all 🙊  :trollface:  
I don't name them, because my concern is: how are we going to prevent lengthy discussions about the extra cops? Especially because that was one of the concerns when we talked about introducing 
Rubocop in the first place. 
So, is it okay if I just review the code, and refrain from adding opinions on the cops itself? Or did you actually _hope_ to get second opinions? Because then we should think of something to address that.

Either way, I am really stoked about getting Rubocop. Dev Happiness Score: 💯

klappradla:
Well, I neither have any expectations, nor do I share your concerns @F3PiX. So feel free to comment as you wish ✌️

.rubocop.yml
```diff
@@ -0,0 +1,188 @@
...
+  StyleGuide: '#indent-public-private-protected'
+  Enabled: true
+
+Layout/IndentationWidth:
+  Description: "Use 2 spaces for indentation."
+  StyleGuide: "#spaces-indentation"
+  Enabled: true
+
+Layout/CaseIndentation:
+  Description: "Indentation of when in a case/when/[else/]end."
```
F3PiX: Maybe include in the description, something like: align when/else/end/ with case. ?
To make it easier to understand what we are suppose to do about it.

.rubocop.yml
```diff
@@ -0,0 +1,188 @@
...
+# TODO
+Style/FrozenStringLiteralComment:
+  Description: "Add frozen_string_literal comment to the top of files"
+  Include:
+  - "app/**/*.rb"
+  - "lib/**/*.rb"
+  Enabled: false
+
+Style/HashSyntax:
+  Description: "Prefer Ruby 1.9 hash syntax { a: 1, b: 2 } over 1.8 syntax"
```
F3PiX: Maybe add a beginner friendly ` { :a => 1, :b => 2 }` after `1.8 syntax` in the description?

.rubocop.yml
```diff
@@ -0,0 +1,188 @@
...
+
+
+#################### Bundler ###############################
+
+Bundler/DuplicatedGem:
+  Description: "Checks for duplicate gem entries in Gemfile."
+  Enabled: true
+  Include:
+    - "**/Gemfile"
+    - "**/gems.rb"
```
F3PiX: 👍  Love that we cover this already 👯

.rubocop.yml
```diff
@@ -0,0 +1,188 @@
...
+  Description: "Limit lines to 80 characters."
+  StyleGuide: "#80-character-limits"
+  Enabled: false
+
+
+#################### Bundler ###############################
+
+Bundler/DuplicatedGem:
+  Description: "Checks for duplicate gem entries in Gemfile."
+  Enabled: true
```
F3PiX: 😍

.rubocop.yml
```diff
@@ -0,0 +1,188 @@
+AllCops:
+  TargetRubyVersion: 2.4
+  DisabledByDefault: true
+
```
F3PiX: Are there any files or directories that we should exclude? I know the `db/schema.rb` is sometimes excluded, but I don't know why TBH.

.rubocop.yml
```diff
@@ -0,0 +1,188 @@
...
+Style/RedundantSelf:
+  Description: "Don't use self where it's not needed."
+  Enabled: true
+
+Style/TrailingCommaInArguments:
+  Description: "Checks for trailing comma in argument lists."
+  StyleGuide: "#no-trailing-params-comma"
+  Enabled: true
+
```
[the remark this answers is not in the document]
klappradla: Sorry, I don't understand your concern about this cop. What's the benefit of
```ruby
def foo(bar,)
  # do stuff
end
```
over
```ruby
def foo(bar)
  # do stuff
end
```
???

You can run `rubocop -a` and it will automatically remove these commas for you.

app/controllers/roles_controller.rb
```diff
@@ -57,7 +57,7 @@ def set_team
 
     def set_role
       @role = if params[:id]
-        @team.roles.find(params[:id])
+                @team.roles.find(params[:id])
       else
```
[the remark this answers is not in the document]
klappradla: Oh that's just a mistake on my side - I ran a different cop for aligning if and else and then only half-reverted it... I'll undo that, thanx for pointing out 🙏

.rubocop.yml
klappradla: You're right, normally one would exclude e.g. the `db` and `bin` directories, just because the code there is not really regarded as "app code" and linting it often times does not make too much sense.

In our case, since we disable all cops, this is not really a problem for now. We don't have any cops that could be problematic for the schema file, migrations or the bin directory.

.rubocop.yml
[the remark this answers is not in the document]
F3PiX: Maybe I misunderstood. I thought it were about the kind of comma's at the end of this line https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/857/files#diff-ea406fc966c043b0342994808d837c01L65 
Of course I don't weep about the example you give above 😂

.rubocop.yml
F3PiX: Aha 💡 .  Thanks 🙇

F3PiX:
How about we set up [CodeClimate](https://codeclimate.com/product) ? That was my initial idea for Rubocop. It is free for Open Source.

.rubocop.yml
carpodaster: I had to google what `gems.rb` is all about. Good stuff, looking forward to it!

CONTRIBUTING.md
```diff
@@ -94,7 +92,8 @@ If you've cloned the app a while ago, you want to make sure that your cloned `ma
 
 ### Write Code
 
-We aim to follow the [Ruby Style Guide](https://github.com/bbatsov/ruby-style-guide) but we don't enforce it (i.e. there is no [RuboCop](https://github.com/bbatsov/rubocop)). The most important thing is probably just proper indentation: two whitespaces.
+We aim to follow the [Ruby Style Guide](https://github.com/bbatsov/ruby-style-guide) and have  [RuboCop](https://github.com/bbatsov/rubocop) set up to make sure we stay consistent with some basic rules (see our [rubocop.yml](https://github.com/rails-girls-summer-of-code/rgsoc-teams/blob/master/.rubocop.yml) file for details).  
+But don't be afraid, the most important thing is probably just proper indentation: two whitespaces ✌️
```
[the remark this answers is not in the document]
carpodaster: :+1: to @F3PiX's suggestion

app/controllers/conference_attendances_controller.rb
```diff
@@ -11,6 +11,7 @@ def update
   end
 
   private
+
     def attendance_params
```
carpodaster: Shouldn't this extra level of indentation be caught by RC?

.rubocop.yml
```diff
@@ -0,0 +1,188 @@
...
+#################### Metrics ###############################
+
+# TODO
+Metrics/LineLength:
+  Description: "Limit lines to 80 characters."
```
carpodaster: I know it's flagged as TODO and disabled, but should we up the limit to 100-something?

.rubocop.yml
carpodaster: Oh, that's a good one, yes

.travis.yml
```diff
@@ -17,6 +17,7 @@ before_script:
 - npm install -g jshint
 script:
 - bundle exec rake spec
+- bundle exec rubocop
```
carpodaster: I wonder if we should use CodeClimate to run the cops; IMHO it gives better feedback on Github PRs: you get two CI results. Travis may run the specs just fine, it may "just" be coding style that needs improvement.

I have never set up CodeClimate with custom rules, though. If we cannot reuse our rubocop.yml, it's not worth pursuing this.

app/models/student.rb
```diff
@@ -43,7 +43,7 @@ def current_team
 
   def current_drafts
     @current_drafts ||= if current_team
-      current_team.application_drafts.in_current_season
+                          current_team.application_drafts.in_current_season
```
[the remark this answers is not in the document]
carpodaster: I think it is, but I cannot name it. I really like this way of indentation. Vim's Ruby syntax auto-formats blocks like this in exactly the same way.

.travis.yml
F3PiX: I think it works like this: we add a codeclimate.yml where we enable rubocop  (and a js linter), and then cc uses our rubocop.yml for the rules.

carpodaster: If it is that easy, I'd prefer CC over running it on Travis

.rubocop.yml
klappradla: I in 99% of cases just directly copied over the description from Rubocop's internal default rule set. Not sure if it makes sense to tweak them and do a lot of changes to them. For most cases the thing that helps explaining / understanding these rules best are examples - and they can be found in the Rubocop docs or the styleguide (also referenced for each rule).

So my idea was to keep the task of explaining things to the part the knows it best - naming the docs / styleguide - and not try to explain it in detail here.

But yeah, I still can do it if you think there's a huge benefit in having it.

app/controllers/conference_attendances_controller.rb
klappradla: Yes, but it's a different cop we don't yet have. But if we already have the one for a blank line before access modifiers and its indentation, it just makes sense to also add the one for proper (in my opinion 😉 ) indentation after it as well.

I'll add it ✌️

app/models/student.rb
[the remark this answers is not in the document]
klappradla: It's [this cop](http://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Layout/ElseAlignment) and I'm gonna add it, since it otherwise conflicts with strict 2-spaces indentation anyways (like here...)

.rubocop.yml
klappradla: Maybe 120? Not sure what really makes sense... Currently we have `215` that's why it's a bit of an effort and I did not want to run it right away.

.travis.yml
klappradla: Just for the sake of completeness: I just came across [build stages](https://docs.travis-ci.com/user/build-stages) for Travis. Would also solve the problem I think.

.rubocop.yml
F3PiX: Sure, let's leave it at the documentation! 👍

.travis.yml
F3PiX: I'll make a separate issue to discuss/decide on the CodeClimate way. Get it out of the way of merging this. 🎉

F3PiX:
🎉  🎉  🎉

PR #859, opened 2017-09-18
Fix flaky controller spec

klappradla:
This attempts to fix (and while that not slow down the test suite...) a test that at least for me frequently failed on CI. It was basically hardcoding an ID which did not always work out right.

I switched it to use RSpec's `assigns` to get the assigned object from within the controller. Doing something like
```ruby
comment = Comment.last
```
would have been another option, but I thought we could spare another hit to the DB for this example.

### DIFF
--- a/spec/controllers/comments_controller_spec.rb
+++ b/spec/controllers/comments_controller_spec.rb
@@ -25,8 +25,7 @@ describe CommentsController do
         it 'redirects to comment on project page' do
           post :create, params: { comment: params.merge(valid_attributes) }
           comment = assigns(:comment)
-          anchor  = "comment_#{comment.id}"
-          expect(response).to redirect_to [project, anchor: anchor]
+          expect(response).to redirect_to [project, anchor: "comment_#{comment.id}"]
         end
       end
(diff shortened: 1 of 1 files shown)


spec/controllers/comments_controller_spec.rb
```diff
@@ -24,7 +24,9 @@
 
         it 'redirects to comment on project page' do
           post :create, params: { comment: params.merge(valid_attributes) }
-          expect(response).to redirect_to [project, anchor: 'comment_1']
+          comment = assigns(:comment)
+          anchor  = "comment_#{comment.id}"
+          expect(response).to redirect_to [project, anchor: anchor]
```
F3PiX: I am moving away more and more from local vars. Do you like  

  ```expect(response).to redirect_to [project, anchor: "comment_#{comment.id}"]```

as well?

klappradla:
Hm, you mean it wasn't flaky? I thought I have seen it quite often 🤔  But could also be I'm mixing things up.

klappradla:
Hm, closing this for now, since I don't find any examples of this failing on CI any more 😹

PR #861, opened 2017-09-18
[DO NOT MERGE] Playing with Travis

klappradla:
As said, I wanted to quickly look into Travis' [beta pipeline feature](https://blog.travis-ci.com/2017-05-11-introducing-build-stages). It in short allows to group the build into `stages` that run sequentially. However, these stages can contain multiple `job` that run in parallel 🤓 

Here I for instance grouped the linters:
screen shot 2017-09-18 at 23 30 12
☝️  note that build usually take about twice the time

Here I had steps with only one job each and the first one failing => the build already stops
screen shot 2017-09-18 at 23 37 26

All in all a nifty feature to speed up and have faster failing builds. Maybe also to have faster access to a stage or job's output. But it does not add more information into Github directly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,12 +17,12 @@ before_script:
   - npm install -g jshint
 jobs:
   include:
-    - stage: tests
+    - stage: test
       script: bundle exec rake spec
-    - stage: rubocp
+    - stage: lint
       script: bundle exec rubocop
-    - stage: jshint
-      script: jshing app/assets/javascripts
+    -
+      script: jshint app/assets/javascripts
     - stage: deploy
       script: skip
       deploy:
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ cache: bundler
 sudo: false
 dist: trusty
 rvm:
-- 2.4.2
+  - 2.4.2
 bundler_args: "-j4 --without development production --quiet"
 notifications:
   notifications:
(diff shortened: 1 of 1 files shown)


klappradla:
Hm, @carpodaster & @F3PiX what do you think of making the build a 2-step process: `build` and `deploy`? This would mean tests, and the 2 linters could run in parallel which would probably speed up the suite... 🏃 💨

F3PiX:
Sounds cool! Would it make sense to run the linters before the tests? Assuming that provides the speediest feedback from the linters. ?

F3PiX:
O... crossTyping again... ;-)

klappradla:
Ok, my idea made sense in theory. In practise, Travis of course spins up a new environment for each job - so the overall time for the build is of course way longer then just running everything sequentially 😹 (at least of our rather simple setup)

But yeah, otherwise what you say @F3PiX - I'd like to switch the order for the normal Travis setup. Plus I'm going to clean up the `.travis.yml` file, it's a bit cluttered and actually neither valid yml nor valid in Travis' linter https://lint.travis-ci.org/

klappradla:
For the sake of completeness:

Staged build:
screen shot 2017-09-20 at 23 51 40

Normal build:
screen shot 2017-09-20 at 23 52 06

klappradla:
Follow up on cleaning up the config is 👉  #864

PR #863, opened 2017-09-19
Community/add tech fields

branquinhoaa:
Related issue #612 

- Includes 2 new fields (not required) for users (Technologies of expertise and Technologies of interest)
tech_fields_edit


app/views/users/edit.html.slim
```diff
@@ -54,6 +54,11 @@
 
   = f.collection_check_boxes :interested_in, User::INTERESTS.to_a, :first, :last, item_wrapper_tag: 'div', item_wrapper_class: 'interested_in'
 
+  h3.page-header Technologies:
+
+  = f.input :tech_expertise_list, placeholder: 'Languages, techniques, etc', label: 'Technologies of expertise', hint: 'Separate tags by comma.'
```
klappradla: How about also phrasing things from a "me"-perspective here? Same as with the other form fields in the user profile?

Maybe something along
> Programming Languages, Frameworks, Technologies, etc. I'm experienced with

and 
> Other Languages, Technologies, etc. I'm interested in

???

Plus, even though we treat these things like "tags" under the hood, for the user it's "just" a text input 😉 
So I would suggest to change the hint to a simple "separate with comma".

app/views/users/edit.html.slim
```diff
@@ -54,6 +54,11 @@
 
   = f.collection_check_boxes :interested_in, User::INTERESTS.to_a, :first, :last, item_wrapper_tag: 'div', item_wrapper_class: 'interested_in'
 
+  h3.page-header Technologies:
```
klappradla: Please skip the `:` - feels too 90s 👯

app/models/user.rb
```diff
@@ -216,6 +216,22 @@ def student_team
     teams.in_current_season.last if student?
   end
 
+  def tech_expertise_list
```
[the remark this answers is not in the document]
klappradla: Oh 🙈 forget everything I wrote ☝️ 

Combined with the usage in the form, your solution is probably way better then moving things out the model 👍

branquinhoaa:
@klappradla I think this one is ready for another review! :+1:

mkalininait:
@branquinhoaa @klappradla Is there a reason why we don't show Technologies of expertise and Technologies of interest on the user/show page?

branquinhoaa:
@mkalininait We can include it! Do you want to?

mkalininait:
@branquinhoaa I think it makes sense. I also don't know why we don't show the company related info and the "interested in" values. I'd show them all :)

branquinhoaa:
@mkalininait ok, I will do it!!

klappradla:
Makes sense ☝️ 

Just ping me again when you're done with these changes @branquinhoaa ✌️

branquinhoaa:
How should I display all this info? :thinking: 

user_show

klappradla:
Looking at you screenshot on how to visualize all the info, I'd say yes, goes in the right direction. Some things still 😉 
* please use a whitespace after the comma
* rephrase "my interests" into "interested in" (since we don't have me / my elsewhere)
* capitalize the field names - I think it's best practice to have headlines like "Technologies of Expertise" rather then "Technologies of expertise" (Rails has helpers for this ✌️  )

branquinhoaa:
@klappradla Where do I should put the space? Because I think that it is already there. The other changes were made! :+1:

klappradla:
Cool @branquinhoaa 🎉  Gonna have a look 👀 
The spaces - I meant the tag lists, but I'll just comment on it in the code.

Also: can you please merge back master to get a green build again?

mkalininait:
@branquinhoaa I'm late for the party, but the changes look great to me! Thank you!

PR #864, opened 2017-09-20
Clean up and lint travis.yml

klappradla:
This fixes some formatting problems in the `travis.yml` and adds classic yml-indentation to it for a better readability.

Background: While experimenting with Travis' pipeline feature in #861, I realized our `travis.yml` had some formatting problems, such as a duplicated "notifications" key with a weird nesting (wonder why it still worked..) 😉

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,42 +3,41 @@ cache: bundler
 sudo: false
 dist: trusty
 rvm:
-- 2.4.2
+  - 2.4.2
 bundler_args: "-j4 --without development production --quiet"
 notifications:
-  notifications:
-    webhooks:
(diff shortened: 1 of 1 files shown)


klappradla:
Having the keys in the `.travis.yml` file is the "official" way to go (they are encrypted before they are put there). I could also define env-vars on Travis directly for them. _However_ their values would be overridden by anything defined in the `.travis.yml` file and they don't work for forks...

Not sure if this makes the file simpler 🤔  I can give it a try though. Separate PR?

F3PiX:
If this is the official way to go, I'd say keep it as it is now. 👍

carpodaster:
:+1: Awesome!

PR #865, opened 2017-10-02
Fix/adapt tests for season transition

klappradla:
This adapts the tests to work regardless if the current season ended or not.

Background: some controller tests were under the hood tied to the fact that the current season is still on and not already transitioning to the next. So I put in some before hooks making sure tests which are meant to target a non-transition phase also work when run during a "real" transition phase 😉

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -42,6 +42,7 @@ group :production do
 end
 
 group :development, :test do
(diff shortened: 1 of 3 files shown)


carpodaster:
👌 🎊

Issue #618, opened 2017-01-18
User Profile: emails to users who have been added to a team


mkalininait:
This issue originates from this one: #593 (see [this comment](https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/593#issuecomment-272316901))

We're gonna send an email once a profile has been saved (to confirm an email address). We should probably try to send an email to team members once someone added them to a team, too (to ask them fill out their profiles). 

We already send emails to coaches (once they are added to a team) and ask them to confirm their coach status:
```
**Subject:** [rgsoc-teams] You have been added to the [TeamName] team.
**Body:** You have been added to Team [TeamName] as a coach.
Please confirm that you are indeed willing to join the team as their coach. 
You can do so on the team's overview page listed above (make sure you 
are logged in using your Github account).
```
It doesn't seem to be sufficient.

Here how the whole process could look like:
![user-profiles](https://cloud.githubusercontent.com/assets/1745735/22045524/1ab3cd24-dd1b-11e6-8572-49ba20e6de4e.png)

It looks a bit bulky, in fact it reflects the following use cases:
**1. Coaches with an email address:**
**1a. Straightforward:** follow the link from the email, get their email confirmed, edit their profile info, save changes, get their coach status confirmed automatically (we want as little clicks as possible I guess).
**1b. With email address change:** follow the link from the email, get their email confirmed, edit their profile info including the email address, save changes, confirm their new email, get another email with a link to confirm their coach status.
**2. Coaches without an email address:** can't confirm their coach status until they fill out their profile and confirm their email; when they do so, they get an email with a link to confirm their coach status (or maybe get redirected to the team's page?).
**3. Other users with an email address:**
**3a. Straightforward:** follow the link from the email, get their email confirmed, edit their profile info, save changes.
**3b. With email address change:** follow the link from the email, get their email confirmed, edit their profile info including the email address, save changes, confirm their new email.
**4. Other users without an email address:** edit their profile one day and confirm their email address.

Please share your opinions. Maybe we could make it all simpler?

28Ritu:
my partner is facing a problem an hasn't received any confirmation mail. Please resolve this asap.

alicetragedy:
hey @28Ritu, please send us an email to our support email address (contact@rgsoc.org) to handle this. thanks

28Ritu:
Please resolve this asap. She is still facing the problem.

28Ritu:
[...]

PR #870, opened 2017-11-09
Sends email to users that were added to a team with any role

LuisaAPF:
Related issue #618 

 - Sends an email to any user that has been added to a team, with any role (not only coaches) 
 - Includes a line on the email asking the user to fill out profile and confirm email

### DIFF
--- a/app/views/role_mailer/user_added_to_team.html.slim
+++ b/app/views/role_mailer/user_added_to_team.html.slim
@@ -13,8 +13,13 @@ html
       a href=team_url(@role.team) Team #{@role.team.name}
       |  as a #{@role.name}.
 
-      - if @role.name == 'coach'
-        p
-          ' Please confirm that you are indeed willing to join the team as their coach.
-          ' You can do so on the team's overview page listed above (make sure you are
-          ' logged in using your Github account).
+    - if @role.name == 'coach'
+      p
+        ' Please confirm that you are indeed willing to join the team as their coach.
+        ' You can do so on the team's overview page listed above (make sure you are
+        ' logged in using your Github account).
+
+    p
+      ' Please complete your profile
+      a href=edit_user_url(@role.user_id) here
+      '  and be sure to confirm your email address.
--- a/app/models/role.rb
+++ b/app/models/role.rb
@@ -20,7 +20,7 @@ class Role < ActiveRecord::Base
   validates :user_id, uniqueness: { scope: [:name, :team_id] }
 
   before_create :set_confirmation_token
-  after_commit :send_notification, on: :create, if: -> { GUIDE_ROLES.include?(name) }
+  after_commit :send_notification, on: :create
 
   after_create do |role|
     role.confirm! unless role.name == 'coach'
--- a/app/views/role_mailer/user_added_to_team.text.erb
+++ b/app/views/role_mailer/user_added_to_team.text.erb
@@ -7,3 +7,6 @@ Please confirm that you are indeed willing to join the team as their coach.
 You can do so on the team's overview page listed above (make sure you are
 logged in using your Github account).
 <% end %>
+
+Please complete your profile at <%= edit_user_url(@role.user_id) %> 
+and be sure to confirm your email address.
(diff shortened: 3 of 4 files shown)


F3PiX:
Hi Luisa! I am not sure if you access to the CI details, but in case not: here is the failing spec
```Failures:
  1) Role create when the user is added as a student does not send a notification
     Failure/Error: expect(RoleMailer).to_not have_received(:user_added_to_team).with(subject)
     
       (RoleMailer (class)).user_added_to_team(no args)
           expected: 0 times with arguments: (#<Role id: 1, team_id: 1, user_id: 1, name: "student", created_at: "2017-11-09 11:09:15", updated_at: "2017-11-09 11:09:15", state: "confirmed", confirmation_token: "cb0f42079448300e">)
           received: 1 time
     # ./spec/models/role_spec.rb:83:in `block (4 levels) in <top (required)>'
Finished in 1 minute 44.6 seconds (files took 5.03 seconds to load)
1456 examples, 1 failure, 1 pending
Failed examples:
rspec ./spec/models/role_spec.rb:82 # Role create when the user is added as a student does not send a notification
```
FYI 
You can run the specs locally with 
`$ bundle exec rspec`
and you can run the single file with:
`$ bundle exec rspec spec/models/role_spec.rb`
or a single spec by adding the line number:
`$ bundle exec rspec spec/models/role_spec.rb:83`
 
Don't hesitate to ping me if you get stuck.

LuisaAPF:
Thanks @F3PiX! I'm running the test locally :grin:

LuisaAPF:
I'm just not sure about the line mentioning the 30 days, because it will not necessarily be true, as the account could have been created some time ago.

F3PiX:
O, of course, you are right, that goes for the account confirmation mail only. Sorry!

app/views/role_mailer/user_added_to_team.html.slim
```diff
@@ -18,3 +18,5 @@ html
           ' Please confirm that you are indeed willing to join the team as their coach.
           ' You can do so on the team's overview page listed above (make sure you are
           ' logged in using your Github account).
+
+      ' Please visit your profile page, to complete your profile and confirm your email address.
```
klappradla: I'd suggest to straight add a link that the receiver can follow to edit their profile:
```ruby
link_to('your profile page', edit_user_url(@role.user))
```
Maybe make `profile page` in this sentence the link.

LuisaAPF: This URL will only work if the user is already signed in. Should I use it anyway or should I include instead a link to the homepage and ask the user to go to the profile page? Which one do you think is better @klappradla , @F3PiX ?

F3PiX: Good point, Luisa. If the user is not logged in, wouldn't the profile link redirect to the login page automatically? Did you check that? Or what happens?

LuisaAPF: In this case I'm redirected to the homepage and I see this message: You are not authorized to access this page.

F3PiX: And if you log in after this redirect, does that automatically redirect you to the profile page?

LuisaAPF: Currently, when I log in I see the activities page

F3PiX: Yeah, that is not very user-friendly. It should be better if we redirect to the last request. 
IIRC that was discussed before, I'll have to look into that. (And maybe add a new issue.)
Thanks for pointing this out! 

Conclusion for now: (cc @klappradla)
I'd say: put the link to the profile in the email anyway. (so that the user can find the link, if they get lost).

app/views/role_mailer/user_added_to_team.html.slim
```diff
@@ -13,8 +13,13 @@ html
...
-          ' logged in using your Github account).
+    - if @role.name == 'coach'
+      p
+        ' Please confirm that you are indeed willing to join the team as their coach.
+        ' You can do so on the team's overview page listed above (make sure you are
+        ' logged in using your Github account).
+
+    p
+      ' Please complete your profile
+      a href=edit_user_url(@role.user_id) here
```
klappradla: Interesting way how links are rendered in this view 👾  But yeah, seems to be fine to aligning with what's already there 😉

klappradla:
Thanx again @LuisaAPF 💚

PR #871, opened 2017-11-10
Move FactoryGirl to 🤖

klappradla:
Simple maintenance PR: the `factory_girl` gem has been renamed. To not look at deprecation warnings, this PR moves the codebase over to the new gem name.

PS: would also be interested in removing the `FactoryGirl` prefix in all files again 🙈

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -44,7 +44,7 @@ end
 group :development, :test do
   gem 'pry'
   gem 'rspec-rails'
-  gem 'factory_girl_rails'
+  gem 'factory_bot_rails'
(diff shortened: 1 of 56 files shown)


F3PiX:
I didn't understand your PS. 🤔

klappradla:
Oh, we made a small effort some months ago to have explicit `FactoryGirl.create(...` instead of just `create(:thing)` in the specs. But after discussing it with Carsten and seeing the other extreme day to day, I think I changed my mind and will change it back 😉

Issue #726, opened 2017-03-31
Cleanup Rating - Application reference


beanieboi:
Right now the ratings tables has 2 possibilities to reference an application.

you can reference an application using the polymorphic `rateable` association (`rateable_type` and `rateable_id`) or you can use `application_id` directly.

We should settle on one style of connecting applications and ratings, either via `rateable`  or `application_id`

klappradla:
Yes, voting for `application_id` - this will however mean, that last / this season's ratings aren't compatible with the schema any more. But I don't think this should be a problem, the whole process changes over time anyways.

carpodaster:
I stumbled over this inconsistency today myself. I also +1 `application_id` – we probably only ever rate applications, no other model.

klappradla:
Closing via https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/875

PR #875, opened 2017-11-14
[#726] Cleanup Rating - Application reference

klappradla:
This removes the redundant polymorphic association for `Ratings` as specified in #726 and uses the already existing `application_id` column for referencing the rated `Application`.

I included a migration to move all the existing "usable" data over to this new data-model. Please note that this migration also **deletes a bunch of `rating` records from 2015** - namely those rating `Users` or `Teams`.
Rating these other entities was possible in the 2015 edition of RGSoC. However, with the new datamodel these records don't make sense any more and I thought it's better to not have them at all instead of running the risk of having these invalid records maybe impacting the current `ratings`...

I also cleaned up the models playing into `rating` (`Todo`, `Application`, `Rating`) to reflect these changes to the data-model and removed a few unused methods from back in the days.

While being already on it, I set all our models to inherit from `ApplicationRecord` the Rails 5 style and add a personal goal to the `new_framework_defaults_5_1.rb` file 😹

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -307,10 +307,6 @@ GEM
     sax-machine (1.3.2)
     sentry-raven (2.7.1)
       faraday (>= 0.7.6, < 1.0)
-    shoulda (3.5.0)
-      shoulda-context (~> 1.0, >= 1.0.1)
-      shoulda-matchers (>= 1.4.1, < 3.0)
-    shoulda-context (1.2.2)
     shoulda-matchers (2.8.0)
       activesupport (>= 3.0.0)
     simple_form (3.5.0)
@@ -414,7 +410,6 @@ DEPENDENCIES
   rubocop
   sass-rails
   sentry-raven
-  shoulda
   shoulda-matchers
   simple_form
   slim-rails
--- a/app/models/rating.rb
+++ b/app/models/rating.rb
@@ -1,10 +1,8 @@
 class Rating < ActiveRecord::Base
+  belongs_to :application, inverse_of: :ratings
+  belongs_to :user
 
   serialize :data
-
-  belongs_to :application
-  belongs_to :user
-  belongs_to :rateable, polymorphic: true
 
   FIELDS = ActiveSupport::HashWithIndifferentAccess.new({
     diversity:
@@ -110,8 +108,11 @@ class Rating < ActiveRecord::Base
 
   before_validation :set_data
 
-  validates :rateable, presence: true
-  validates :user_id, presence: true, uniqueness: { scope: [:rateable_type, :rateable_id] }
+  validates :application, presence: true
+  validates :user,        presence: true
+  validates :user_id,     uniqueness: { scope: :application_id }
+
+  scope :by, -> (user) { where(user_id: user.id) }
 
   class << self
     def user_names
@@ -120,14 +121,6 @@ class Rating < ActiveRecord::Base
 
     def excluding(names)
       joins(:user).where('users.name NOT IN (?)', names)
-    end
-
-    def by(user)
-      where(user_id: user.id)
-    end
-
-    def for(type, id)
-      where(rateable_type: type, rateable_id: id)
     end
   end
 
--- a/app/models/rating.rb
+++ b/app/models/rating.rb
@@ -1,5 +1,5 @@
 class Rating < ApplicationRecord
-  belongs_to :application, inverse_of: :ratings, required: true
+  belongs_to :application, required: true
   belongs_to :user, required: true
 
   serialize :data
--- a/app/models/rating.rb
+++ b/app/models/rating.rb
@@ -1,6 +1,6 @@
 class Rating < ActiveRecord::Base
-  belongs_to :application, inverse_of: :ratings
-  belongs_to :user
+  belongs_to :application, inverse_of: :ratings, required: true
+  belongs_to :user, required: true
 
   serialize :data
 
@@ -108,25 +108,9 @@ class Rating < ActiveRecord::Base
 
   before_validation :set_data
 
-  validates :application, presence: true
-  validates :user,        presence: true
-  validates :user_id,     uniqueness: { scope: :application_id }
+  validates :user_id, uniqueness: { scope: :application_id }
 
   scope :by, -> (user) { where(user_id: user.id) }
-
-  class << self
-    def user_names
-      User.includes(:roles).where('roles.name = ?', 'reviewer').pluck(:name)
-    end
-
-    def excluding(names)
-      joins(:user).where('users.name NOT IN (?)', names)
-    end
-  end
-
-  def woman?
-    data[:is_woman] == 1
-  end
 
   # public: The weighted sum of the points that the reviewer gave.
   def points
--- a/app/models/rating.rb
+++ b/app/models/rating.rb
@@ -1,4 +1,4 @@
-class Rating < ActiveRecord::Base
+class Rating < ApplicationRecord
   belongs_to :application, inverse_of: :ratings, required: true
   belongs_to :user, required: true
 
--- a/app/models/rating/strictness.rb
+++ b/app/models/rating/strictness.rb
@@ -44,9 +44,7 @@ class Rating::Strictness
   private
 
   def ratings
-    @ratings ||= Rating
-                  .includes(:application)
-                  .where('applications.season_id' => season.id)
+    @ratings ||= Rating.joins(:application).where('applications.season_id' => season.id)
   end
 
   # @return [Array<Integer>] list of IDs for all rated applications
--- a/app/models/rating/strictness.rb
+++ b/app/models/rating/strictness.rb
@@ -45,14 +45,13 @@ class Rating::Strictness
 
   def ratings
     @ratings ||= Rating
-                  .joins('JOIN applications ON ratings.rateable_id = applications.id')
-                  .where(rateable_type: 'Application')
+                  .includes(:application)
                   .where('applications.season_id' => season.id)
   end
 
   # @return [Array<Integer>] list of IDs for all rated applications
   def application_ids
-    @application_ids ||= ratings.pluck(:rateable_id)
+    @application_ids ||= ratings.pluck(:application_id)
   end
 
   # @return [Array<Application>] all rated applications
--- a/app/models/todo.rb
+++ b/app/models/todo.rb
@@ -32,6 +32,6 @@ class Todo < ApplicationRecord
   end
 
   def validate_number_of_reviewers
-    errors.add(:user, 'too many reviewers') if application.todos.count > 3
+    errors.add(:user, :too_many_reviewers) if application.todos.count > 3
   end
 end
--- a/app/models/todo.rb
+++ b/app/models/todo.rb
@@ -5,8 +5,7 @@ class Todo < ApplicationRecord
   belongs_to :user, inverse_of: :todos, required: true
   belongs_to :application, inverse_of: :todos, required: true
 
-  delegate :season, to: :application
-  delegate :application_data, to: :application
+  delegate :application_data, :season, to: :application
 
   validate :validate_number_of_reviewers, if: :application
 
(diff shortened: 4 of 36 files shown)


spec/models/application_spec.rb
```diff
@@ -1,61 +1,59 @@
...
     end
+
+    describe '.rateable' do
+      subject(:applications) { described_class.rateable }
+
+      let!(:application)       { create(:application, :in_current_season) }
+      let!(:from_other_season) { create(:application, season: build(:season)) }
+      let!(:without_team)      { create(:application, :in_current_season, :skip_validations, team: nil) }
+
+      it 'returns only applications from the current season which a team' do
```
F3PiX: 'with' a team instead of 'which', I presume

spec/models/todo_spec.rb
```diff
@@ -71,19 +78,14 @@
...
-      rating.save(validate: false) # TODO: tmp solution
+    subject(:todo) { create(:todo) }
+
+    it 'returns true if the application has been rated' do
+      rating = create(:rating, user: todo.user, application: todo.application)
       expect(todo).to be_done
     end
 
-    it 'returns false if application not rated yet' do
+    it 'returns false if the application not rated yet' do
```
F3PiX: missing 'is'

app/models/todo.rb
```diff
@@ -33,6 +33,6 @@ def sign_offs?
   end
 
   def validate_number_of_reviewers
-    errors.add(:user, "too many reviewers") if application.todos.size > 3
+    errors.add(:user, 'too many reviewers') if application.todos.size > 3
```
F3PiX: How about we start to add error messages to the en.yml?

app/models/rating.rb
```diff
@@ -110,30 +108,9 @@ class Rating < ActiveRecord::Base
...
-    end
-
-    def for(type, id)
-      where(rateable_type: type, rateable_id: id)
-    end
-  end
-
-  def woman?
-    data[:is_woman] == 1
-  end
```
F3PiX: Bouble check:  this criterium is not in use anymore?

app/models/todo.rb
```diff
@@ -33,6 +32,6 @@ def sign_offs?
   end
 
   def validate_number_of_reviewers
-    errors.add(:user, "too many reviewers") if application.todos.size > 3
+    errors.add(:user, 'too many reviewers') if application.todos.count > 3
```
F3PiX: How about we start with adding the errors to the en.yml?

config/initializers/new_framework_defaults_5_1.rb
```diff
@@ -9,6 +9,10 @@
 # Make `form_with` generate non-remote forms.
 Rails.application.config.action_view.form_with_generates_remote_forms = false
 
+# The belongs_to associations are required by default in Rails 5
+# we cannot opt into this right now, but it would be a goal for the future
+Rails.application.config.active_record.belongs_to_required_by_default = false
```
F3PiX: Does this goal involve more than adding `inverse of` everywhere? (checking if it is a beginner-friendly issue)

spec/controllers/rating/applications_controller_spec.rb
```diff
@@ -1,6 +1,6 @@
 require 'spec_helper'
 
-describe Rating::ApplicationsController do
+describe Rating::ApplicationsController, type: :controller do
```
F3PiX: I thought we don't need to set the type because we have:
https://github.com/rails-girls-summer-of-code/rgsoc-teams/blob/32e7940730b453245cbdfd42a7ec9f1e625b0771/spec/support/database_cleaner.rb#L26

app/models/rating.rb
klappradla: I couldn't find any usage of it in the code. Wasn't there a bigger change to the way people would specify their gender?

config/initializers/new_framework_defaults_5_1.rb
klappradla: Hm, to be honest, I don't know 🙈 
I would guess it's probably going over all associations, adjusting / removing the validations, updating specs and factories, etc.

spec/controllers/rating/applications_controller_spec.rb
klappradla: For having `database_cleaner` work as before, yes. But maybe someone wants to add a specific config for controller-types specs, e.g. including certain matchers there...?

I actually did not put much thought into this 😉 

However, I think it would be good to have, especially if we finally start to add feature-specs (we should def. do this) - so that we can distinguish which parts of the suite require a headless browser, etc.
Plus, we could split the build to run feature-specs and non feature specs in parallel...

app/models/rating.rb
F3PiX: If it is not in the code, than it is not in the rating #thingsRailsGirlsSay 😂

config/initializers/new_framework_defaults_5_1.rb
F3PiX: Let's NOT make it a beginners issue ;-)

Gemfile.lock
```diff
@@ -307,10 +307,6 @@ GEM
     sax-machine (1.3.2)
     sentry-raven (2.7.1)
       faraday (>= 0.7.6, < 1.0)
-    shoulda (3.5.0)
```
carpodaster: Interesting, I always thought that shoulda-matchers would depend on it. The more you know :) Good bye superfluous dependency :wave:

app/models/rating/strictness.rb
```diff
@@ -45,14 +45,13 @@ def strictness_per_reviewer
 
   def ratings
     @ratings ||= Rating
-                  .joins('JOIN applications ON ratings.rateable_id = applications.id')
-                  .where(rateable_type: 'Application')
+                  .includes(:application)
```
carpodaster: Not sure this is an equivalent refactor. `joins` is an INNER and `includes` is an OUTER JOIN if I recall correctly, meaning you will project values don't have matching association.

config/initializers/new_framework_defaults_5_1.rb
carpodaster: :thinking: I'm not familiar with this inverse_of. I remember that at one point I was starting to use it, only to learn that Rails core was thinking about removing it again. Never followed up on that one. It would seem it's still there and heaved upon the `belongs_to` macro at that?

db/migrate/20171112174323_remove_polymorphism_from_ratings.rb
```diff
@@ -0,0 +1,32 @@
+class RemovePolymorphismFromRatings < ActiveRecord::Migration[5.1]
+  class Rating < ActiveRecord::Base; end
```
carpodaster: Does this need a `self.table_name = 'ratings'`? Because of this being an inline class I'm thinking the table name would be auto-resolved to `remove_polymorphisms_from_ratings_ratings`

db/migrate/20171112174323_remove_polymorphism_from_ratings.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  def up
+    add_index :ratings, :application_id
```
carpodaster: We're definitely not working with big data here, but I've recently come across the `algorithm: :concurrently` option in [this article](https://robots.thoughtbot.com/how-to-create-postgres-indexes-concurrently-in)

spec/controllers/rating/applications_controller_spec.rb
carpodaster: If we are to add the `type` key, can we also change the `describe` to `RSpec.describe`? No, no need to update this PR, just thinking about it while reading this.

config/initializers/new_framework_defaults_5_1.rb
klappradla: 🙈 thanx a million for pointing this out @carpodaster 🙏 
I since forever "just used"™️ `inverse_of`, because I thought it would do some behind the scenes optimization for looking up the association (see the beginning of [this blogpost from back in the days](https://www.viget.com/articles/exploring-the-inverse-of-option-on-rails-model-associations)).
**BUT** as it turns out, this was already added as a default behavior to Rails 4 (and therefore probably also Rails 5). It's probably still around for associations where the inverse cannot automatically be determined by the name - but for basic cases like most of the associations in the Teams App, this is obsolete 💡 

So I guess I can ditch the `inverse_of`s I added in this PR 🎉 

Learned something - yay!

Gemfile.lock
klappradla: I've never used `shoulda`, but I think it adds some `RSpec` like macros to `minitest`. So you probably need it for the "should" part if you use these matchers with `minitest`.
For RSpec, we won't need it.

config/initializers/new_framework_defaults_5_1.rb
F3PiX: If I understand it correctly, the 'inverse of"  is (now?) required when validating associated records and/or on polymorphic associations.  https://github.com/rails/rails/pull/18937
I can't see if this is relevant for this PR.  If it isn't, please ignore this :-)

app/models/rating/strictness.rb
klappradla: Yep, you're totally right!

Should be
```ruby
Rating.joins(applications)
```

Will do the correct query then:
```ruby
> Rating.joins(:application).where('applications.season_id' => season.id)
  Rating Load (5.4ms)  SELECT  "ratings".* FROM "ratings" INNER JOIN "applications" ON "applications"."id" = "ratings"."application_id" WHERE "applications"."season_id" = $1 LIMIT $2  [["season_id", 1], ["LIMIT", 11]]
```

db/migrate/20171112174323_remove_polymorphism_from_ratings.rb
klappradla: Hm, good point! This is actually a scenario I never thought of before - but sounds plausible 🤔 

The reason why I thought this wouldn't be a problem: the migration-class is not an `ActiveRecord::Base` class and therefore does _(at least that's how I understand it)_ not add a prefix to its nested classes...

But what you say sounds plausible - I'll add it just to play save ✌️

db/migrate/20171112174323_remove_polymorphism_from_ratings.rb
klappradla: Neat! 🤘

klappradla:
Million thanx for the feedback @carpodaster & @F3PiX - updated accordingly ✅ 

And yes, sorry for the noise 🙈 coming across weird files is indeed tempting for "drive-by-fixes" _(that's how we call this at work...)_, because I of course love things nice and tidy in the Teams App 😉

F3PiX:
Thank you, @klappradla!
and.. uhm...  
> How about we start with adding the errors to the en.yml?

Shall we make that a separate issue, maybe? If we show an example and layout the steps, it may be a nice beginner_friendly issue. WDYT?

And out of curiousity. I am doing a similar 'drive-by-fix' in a PR I am currently working on (in another project). What is the neat way to clean up, given that we want to leave the code cleaner when we leave?

klappradla:
About the error message @F3PiX : I quickly added a commit doing this. I did not forget about it, I just thought it was more of a "general" question 😉 
However, yes, that can def. be a beginner-friendly issue 👍  If you've already got the words for it in your head, feel free to add it. I can also do it myself tomorrow.

For driving by: depends on the rottenness of the codebase one's on. For a normal state of legacy, I guess it's best practise to clean up all the areas that the current task touches. Where "touches" can either be on file-base or closer associated with the feature or task itself. However, "cleaning up" is of course far from refactoring everything to a new code style, but more like small improvements that can be done in seconds and save hours later 😉

db/migrate/20171112174323_remove_polymorphism_from_ratings.rb
klappradla: Just for reference @carpodaster: you were so right and I by coincidence also ran over a (or the 🤔 ) blog post about it today👇 
http://blog.testdouble.com/posts/2014-11-04-healthy-migration-habits.html

carpodaster: > The ~ring~ article _wants_ to be found!

– Gandalf the Grey on Active Record migration techniques

klappradla:
Do you wanna take another look or should I straight merge @F3PiX and @carpodaster ?

_you can emoji-vote on this comment_

carpodaster:
Do it! :)

PR #876, opened 2017-11-14
Add .ruby-version file and update travis.yml

klappradla:
This updates our config for Travis to basically not build every PR twice. It should be enough to build every PR and the master / staging branch. If individual branches should be build, one can still do this by hand 👇 

screen shot 2017-11-14 at 23 31 23

I also added a `.ruby-version` file so that Travis can pick it up. We won't be testing against any other versions + most people probably already use that file anyways ✌️  - therefore I don't see a problem in having it in the repo _(maybe also makes it harder to accidentally use the wrong Ruby version)_.

Spoiler: I also started to use an interesting pattern to have fail-fast Travis builds at work. Will add a PR to the Teams App as well as soon as I find the time.


klappradla:
> Way to save the environment!

💚

F3PiX:
I am late to the party ;-) 
One question: we do have the ruby version in the gemfile. [I learned that build will fail with different ruby versions](https://github.com/travis-ci/travis-ci/issues/953). Do we need the ruby version file anyway?   
(Also, The gemfile declaration prevents that you use the wrong ruby version.)

klappradla:
Thanx for pointing out @F3PiX 🙏 

Yes, it's true that Travis won't let you use versions different to the one specified in your Gemfile - or at least it was like this for the last 2 years. Specifically if you specify different versions from `rvm` for Travis to chose from. If I remember correctly the build then fails even before any build-steps are run, since Travis cannot match the configuration and the files in the repo correctly.
However, all ruby version managers (rbenv, chruby, rvm, etc.) can use a `.ruby-version` file to automatically pick up the right version if it's installed. And so does Travis 🎉  _(in fact, that's how I usually use Travis in other projects: only have a `.ruby-version` file in the repo which solves work picking the right version in development, ci and on production)_.

The Gemfile-declaration is for Heroku _(no clue why they don't use `.ruby-version` files)_ and from what I remember it does not have any impact on the Ruby version you use on your own machine with it. I could imagine that `rvm` is able to read and pick a version from a Gemfile, but it's very unlikely that it has preference over one specified in a `.ruby-version`.

So **tl;dr** it should not neither break our builds nor your local dev setup ✌️

F3PiX:
Thanks to you! Interesting stuff. I'll go and add  .ruby-version's to my projects. ;-)

PR #877, opened 2017-11-19
updated styling for teams page dropdown

### LINKED ISSUE #874: Improve UI of “Past teams” dropdown
The [list of teams](https://teams.railsgirlssummerofcode.org/teams) currently has a dropdown that allows you to view past teams, which is great! However the styling of the dropdown could be improved with the following:
1) Text and button shouldn't be grey/greyed out, as it gives the impression that it's disabled
2) some left padding for the content of the dropdown list would be great, right now the first number of each year is completely stuck to the left dropdown border, which is a little ugly and makes it difficult to read

I've attached a screenshot below. I haven't looked into the code so I don't know if this dropdown is using bootstrap styles or not, but I think it's a fairly beginner-friendly issue for someone interested in UI and CSS.

screen shot 2017-11-13 at 14 57 53

mbrien12:
Improved styling of the dropdown menu on the teams page
* The text colour is black so doesn't look inactive - I over-wrote the bootstrap class for form inputs with this colour. I checked the site for other places this could impact and couldn't find anything, but if preferable not to override this class then instead could create id for this specific dropdown?
* Added padding to the items in the list so matches the main navigation

screen shot 2017-11-19 at 18 11 07

Fixes #874


app/views/teams/index.html.slim
```diff
@@ -2,7 +2,7 @@ nav.actions
   ul.list-inline
     li = link_to 'My Team', current_student.current_team, class: 'btn btn-default btn-sm' if current_student.current_team
     li = link_to icon('plus', 'New Team'), new_team_path, class: 'btn btn-primary btn-sm' if can? :create, Team.new(season: Season.current)
-  div.pull-right.dropdown
+  div.pull-right.dropdown.center-dropdown
     button.dropdown-toggle.form-control  data-toggle="dropdown"
```
klappradla: I think one of the problems with this button is that it's just using the _"wrong"_ css classes - one that's normally intended for forms. Maybe have a look at [the old bootstrap docs](https://bootstrapdocs.com/v3.3.5/docs/components/#btn-dropdowns-single) what markup they use in their examples.

I could imagine something like:
```ruby
button.dropdown-toggle.btn.btn-default data-toggle='dropdown'
```
to work with the right colors etc.

Do you wanna give it a try? Maybe better on the long run compared to overriding bootstrap defaults 🤔

mbrien12:
Yep you're right! Have removed the bootstrap over-ride and updated with old class. This fixed the overall styling of the button so doesn't appear grey anymore, but couldn't find a way to fix the padding with just bootstrap classes.

It's not following the convention of the main nav bar dropdown because they are <a> tags rather than just <li> so I still included a new class here to include padding; does that work?

app/views/teams/index.html.slim
```diff
@@ -2,12 +2,28 @@ nav.actions
   ul.list-inline
     li = link_to 'My Team', current_student.current_team, class: 'btn btn-default btn-sm' if current_student.current_team
     li = link_to icon('plus', 'New Team'), new_team_path, class: 'btn btn-primary btn-sm' if can? :create, Team.new(season: Season.current)
-  div.pull-right.dropdown
-    button.dropdown-toggle.form-control  data-toggle="dropdown"
+  div.pull-right.dropdown.padding-dropdown
```
klappradla: Hm, I think I don't really understand this part 🙈  why do we also need the padding on the `div`...?

app/views/teams/index.html.slim
```diff
@@ -2,12 +2,28 @@ nav.actions
...
+    button.btn.btn-default.dropdown-toggle data-toggle="dropdown" aria-haspopup="true"
```
klappradla: 👍 
Just curious: where does the `aria` attribute come from and what does it mean? We don't have a popup here, that's why I'm wondering... But I'm also not familiar with these tags  🐣

app/views/teams/index.html.slim
```diff
@@ -2,12 +2,28 @@ nav.actions
...
-        li data-behaviour="switch-teams-year" value=year =year
+        li.dropdown-padding data-behaviour="switch-teams-year" value=year =year
 
-= render 'table', namespace: nil
\ No newline at end of file
+= render 'table', namespace: nil
+
+
+
+/ <!-- Single button -->
```
klappradla: Probably a left over from fiddling with bootstrap 😉  Can you please remove this...?

klappradla:
Hey @mbrien12 👋 

Just wanted to check back if you're still on this and if you need more information, help or anything else?

mbrien12:
Sorry for the delay! I'm going to be taking a look at this tomorrow so
should be done then 👍

mbrien12:
Yeah think `li` items works much better thanks! Means it inherits the correct padding now too, how's this now?

app/assets/stylesheets/controllers/teams.sass
```diff
@@ -25,4 +25,7 @@
 
 .icon-star-volunteer
   @include icon-before($star-empty)
-  color: $railsred
\ No newline at end of file
+  color: $railsred
+  
+.padding-dropdown
```
klappradla: Can you please remove this class again? It's not used anymore and having unused code lying around is never good 😉

app/views/teams/index.html.slim
```diff
@@ -3,11 +3,15 @@ nav.actions
     li = link_to 'My Team', current_student.current_team, class: 'btn btn-default btn-sm' if current_student.current_team
     li = link_to icon('plus', 'New Team'), new_team_path, class: 'btn btn-primary btn-sm' if can? :create, Team.new(season: Season.current)
   div.pull-right.dropdown
-    button.dropdown-toggle.form-control  data-toggle="dropdown"
+    button.btn.btn-default.dropdown-toggle data-toggle="dropdown"
       | Past Teams
       span.caret
```
klappradla: Just tried it out locally. Works perfect, but is missing a whitespace between "past teams" and the caret 👾 

It's only a subtle difference, but I think it makes it look way cleaner with a separation:
_without whitespace:_
screenshot-2017-12-4 rgsoc - teams app 2

_with whitespace:_
screenshot-2017-12-4 rgsoc - teams app 1


The syntax in slim therefore is a bit weird though 😉  See [the docs on it](http://www.rubydoc.info/gems/slim/frames#Trailing_and_leading_whitespace_______)

So here it would be:
```ruby
span.caret<
```

mbrien12: I was trying to figure out how to do that thanks!!🙈

mbrien12:
Pushed changes requested above 👆

app/views/teams/index.html.slim
klappradla: Hm, maybe you forgot to push the change? I don't see it 🙈

It's literally this syntax:
```
span.caret<
```

In the changes I just see a "normal" whitespace after the `Past Teams` - this unfortunately does not work with all browsers plus most editors (mine for instance) remove trailing whitespace so this won't be around for too long 😞 

Could you update to use the slim syntax? Then we're good to merge.

F3PiX:
Hi @mbrien Could you check if the dropdown is actually working, and selecting the year's teams? I happened to notice it is not selecting anything on master.

mbrien12:
Hi @F3PiX I received a notification of another message which isn't displaying here, about the dropdown not selecting anything on master, is that still happening?

I realised I made this change without having any data in the app so when I tried to seed now I get an error saying:
```NoMethodError: undefined method `create_list' for main:Object```

This was after running `rails db:seed` - can you advise how I can seed the app so I can test the dropdown properly? Thanks in advance!

F3PiX:
Hi @mbrien12,
We just discovered - and solved - the seeding issue yesterday!  
Just pull the latest changes. 
Now, the seeding should work - on your master branch.

Now you can either rebase onto master, or create a new local branch off of master, and copy the changes from `ui-enhancement-team-dropdown` to your new local branch.  

And - sorry about the misleading notification - then the dropdown should work!

mbrien12:
Have ran the seeds and tested the dropdown again and is selecting teams corresponding to the year. But have noticed that '2018' is empty. Which makes sense as selecting 'past' teams!  Should it display as an option still? If so, would be good to at least re-order the list so displays in date order, I can look at this week?

F3PiX:
Good catch, @mbrien! Thanks!
You are right, we can do without the _succeeding_ year, in this case 2018.
It is a bit tricky though, because in a week or 3, 2018 is the _current_ year, and we _do_ want to show that. (If you want to go another step: we do want to have that from the moment we do have teams for 2018. :-)  )

I'd say: you can change the `:past_years` method in the `Season` model, and let it return the year names without the `succ`eeding season.
Let us know if you need help; I just sketched it here so you can take a stab on it, but feel free to ask for help or hints :-)
 
And now that we are on it: as the current season is also in the dropdown list, _and_ that is the only link to get back to the current season's team:   How about we change the label to: 'By Year' or 'Teams by year' instead of 'Past Teams'?

cc @klappradla Agree?

klappradla:
Not sure I got every detail (my brain is already a bit melted today) - but my ➕  on displaying the years in order and rephrasing the text on the button 👍 

I would however prefer this in a separate PR, so that we don't have this one linger around forever with the efforts of constantly keeping it up to date 😉 

So, if you address [my previous comment about 👾 ](https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/877#discussion_r155032510), we could merge this one and do the naming and sorting in a separate one @mbrien12 .

mbrien12:
I have pushed a change with `<` on the caret but it shows up in the dropdown when added to 'Past teams'. I replied to comment in review but may have got lost in thread, see here. How is it now?

screen shot 2017-12-14 at 10 26 28

klappradla:
Yay 🎉  that's perfect and exactly what I was talking about. Sorry if I did not make myself that clear before 🙈  . As for the comment: could also have been me who overlooked / lost it - my Github notifications are a constant information overload... 🚚 

Anyways: do you wanna continue the additions proposes by @F3PiX in a separate PR then? Do you want a ticket for it or is it enough for you if you can go back to this thread right here @mbrien12 ?

And again: million thanx for your efforts 💚

F3PiX:
I already created the issue: #895

mbrien12:
Ah me too! I will close! #896

F3PiX:
Sorry! Thanks!

PR #879, opened 2017-12-02
Move ApplicationForm constants from User  into its own class

F3PiX:
Related issue #(none)  Simplify working with the User model 

- These constants are used only in the ApplicationForm. 
Step 1: They get their own class, named (for now) `ApplicantFields`. 

- Questions:
  - 1 Is there a better name for the Class? An `ApplicationForm` object already exists. I thought about naming the new class: `ApplicationFormData` or `ApplicationFormFields`. That describes the relation to the Form better, but there are more fields than just the personal ApplicantFields. 
Also: should it be singular ApplicantField? 
  - 2 There is a separate SkillLevel class, that is a similar 'applicant field', and also used only in the form. Does it make sense to combine the ApplicantFields and SkillLevel into one class?
  - 3 The  application form has a gender list field, but it doesn't show the select options as stated in `ApplicantFields::GENDER_LIST`.  
https://github.com/rails-girls-summer-of-code/rgsoc-teams/blob/0cf4b724db8085f8c78c6a56b2e9c31e4690e8c6/app/views/application_drafts/_student_fields.html.slim#L12 
screen shot 2017-12-02 at 17 04 21
Is should show as a select, shouldn't it?


F3PiX:
@klappradla Can you take a look please?

app/models/applicant_fields.rb
```diff
@@ -0,0 +1,34 @@
+class ApplicantFields
+  AGE = [
```
klappradla: Does not really make a difference in usage, but how about starting to use `%w(` syntax here and then `.freeze` these arrays? I think the [% literal syntax](https://docs.ruby-lang.org/en/2.4.0/syntax/literals_rdoc.html#label-Percent+Strings) could be a bit more readable for these lists.

klappradla:
For your other questions:
1. _I kind of answered this in my comment on the PR I think_
1. Not 100%. From what I remember it still does something slightly different. It does not just define some constants, but also has some utility functions for rendering data in form inputs. So straight combining them is probably not possible and may again just make things harder to distinguish. But moving it somewhere else, closer to the form and other form related things could be an option.
1. From what I remember it's a text input with autocomplete triggerd by some JS. So it's actually using the values - but not as a select _(you're still free to input whatever you like)_.

☝️ is that of any help? I feel like my brain is a 🎈 today...

F3PiX:
It is very helpful! Thanks. 

> But for the quick goal of getting things out of the way in the user model but not growing the already huge ApplicationForm object, this is the first solution that comes to my mind.

I realise I am thinking ahead, and maybe overthinking it. Because: it turns out we do have an [application_form_helper](https://github.com/rails-girls-summer-of-code/rgsoc-teams/blob/master/app/helpers/application_form_helper.rb). How about I put it just in there?

The reason I am not (yet) wildly enthousiastic about the `forms/` directory, is because we do have the application_draft as well. (Which has its own namespace already.) And is also tightly related to the application_form.
I also looked at presenters, but that looks like overengineering for the current simple goal.

So, although I do not like helpers at all, in this special case it might be serving the quick goal?  
WDYT? Helper or module?

F3PiX:
@klappradla ping ^^

klappradla:
So you mean module that gets included into the form class? 👍 sold

I'm not a huge fan of helpers. They're global namespace and tend grow out of hand real quick.

I have used [draper](https://github.com/drapergem/draper) for decorators before which worked quite good - kind of as an alternative to helpers.

klappradla:
In addition to that and helpers tending to rot: I think we can safely remove the `ApplicationHelper` module as it seems not to be used: 😹 
```bash
❯ ag time_span_array
app/helpers/application_form_helper.rb
3:  def time_span_array

spec/helpers/application_form_helper_spec.rb
4:  describe '.time_span_array' do
6:      expect(time_span_array.size).to eq(6)
```

F3PiX:
Yes, I noticed. 👍   Delete! Delete! Delete!

F3PiX:
@klappradla Can you take a look please? 
* I - reluctantly - renamed to StudentFields, because that name is already in use for those fields.
* I really like the idea of the Decorators. But I'd propose not to do it in this pr. Okay?

app/models/student_fields.rb
```diff
@@ -0,0 +1,34 @@
+module StudentFields
```
klappradla: "Concern" I have with this: we're polluting the already **huge** `models` folder with things that aren't models and only make sense when combined with an actual *model*. Therefore I'd vote for: making it a `concern` 😂 

There's the `concerns` directory in the `models` directory. You could add a fitting namespace, maybe `Fields` and then stuff it there:
```ruby
# in app/models/concerns/fields/student_fields.rb

module Fields
  module StudentFields
    # all the fun
```

What do you think?

app/views/application_drafts/_student_fields.html.slim
```diff
@@ -8,8 +8,8 @@
...
-  = s.input :application_gender_identification, as: :string, label: 'Your gender self-identification', input_html: { disabled: !may_edit?(s.object), "data-genders-src" => User::GENDER_LIST.to_json, class: "gender-form-field"}, wrapper_html: { class: student_field_error_class(index, application_draft, :application_gender_identification) }, hint: (student_field_error_hint(index, application_draft, :application_gender_identification) ||  'No-one is excluded from applying on the basis of gender, but people who self-identify as female, or have experiences being socialized as female, are given preference during selection.')
+    = s.input :application_age, :collection => StudentFields::AGE, include_blank: true, label: 'Your age', disabled: !may_edit?(s.object), wrapper_html: { class: student_field_error_class(index, application_draft, :application_age)}, hint: student_field_error_hint(index, application_draft, :application_age)
```
klappradla: Not sure how relevant for the moment, but I would vote for **not** decoupling the fields / constants from where they actually belong to and just reference them straight from the model.

So in this case we're dealing with an `ApplicationDraft` model / instance right? My initial idea would have been to simply include the fields module there
```ruby
class ApplicationDraft
  include Fields::StudentFields
```
This would add all the constants and functionality to the `ApplicationDraft` and you'd use it directly from there.

app/models/student_fields.rb
carpodaster: can you add the frozen_string_literal magic comment to this class? That way the values in the arrays are also frozen (and the memory objects can be reused)

F3PiX: I was in doubt about it, because I learned not to use `concerns` for things that are in one model only. But it looks so much better, so  👍   totally agreed.

carpodaster: On one hand, I agree that this module shouldn't reside in app/models. Otoh, why do we care? There is no real definition what should go into models. Only persistence objects? We would have to through out quite a few files then. And were to put them? Coming up with new subfolders under `app/` makes it more difficult to find the implementation of a class than browsing a densely  packed app/models folder.

The current implementation of `StudentFields` does look as if it belonged into `app/concerns`.

That being said, I'm not so much a fan of models (concerns) as a bandaid to remove the line count of a fat model when they're only used in a single class. If anything, it makes it only less explicit what a model is capable of.

If we don't want to expose constants to the form, can we transform these to module functions? Or, better yet (and certainly not in this PR), add them as methods to a custom form builder for the the students fields form?

F3PiX: We can decide later on a more permanent decision? I'll keep all this in mind, together with the option of decorators.

klappradla: 👍 

I think the important part was just to realize that these things should be part of the form object rather then the user model - and we have that now.

F3PiX:
Look at thaaaat! Isn't it neat now?

app/models/concerns/field_populators/skill_level_fields.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+module FieldPopulators
+  module SkillLevelFields
```
klappradla: Hm, I fear this does not really fit a module. From what I remember it was implemented as a separate class (and not adding a constant to the `ApplicationDraft` model, since it also has some **bahavior** (see the metaprogramming fun on the Eigenclass) and is used in 2 different places (`Mentor` views and `ApplicationDraft` views).

Given it has behavior attached to the Eigenclass, I'd say moving this into a module and including it somewhere else is *entering dangerzone* - I just would not do it because it's super hard to grasp all the potential sideeffects this can have 🔥

app/models/mentor/student.rb
```diff
@@ -1,6 +1,7 @@
 module Mentor
   class Student
     include ActiveModel::Model
+    include FieldPopulators::SkillLevelFields
```
klappradla: As I said: dangerzone _(plus it's actually not even used in the view like this)_

F3PiX:
Sorry to you Max! I'll revert the SkillLevel. 
About the other two though. Would you mind terribly to keep them concerns? I'd still prefer that over having view-things in the model itself. Also because it will be REALLY crowdy.

klappradla:
Looking at the model, to be honest, yes 🙈 

It contains mostly decorator functions for the view, so the constants would actually fit best there.

One could then later try to reduce the complexity / LOC of this class, but with a different approach (form builder, decorator, etc.). Moving things to concerns would just split the domain, Carsten was right with that...

F3PiX:
Better turned at half than to get lost altogether :-)

klappradla:
💚 
Btw. could it be that the `ApplicationForm` class is not even used? 🙈 I couldn't find a usage on the first quick look...

F3PiX:
OMG!!! Really 🤣   

Thanks for the lessons and the merge 🙇

PR #880, opened 2017-12-05
Lint spec for factories

klappradla:
This is a follow up for #871. It adds a lint spec to the test suite which "lints" all our `FactoryBot` factories based upon the [build in lint task](https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#linting-factories) 👮 .

The idea / motivation have factories build the most minimal **but valid** object. Being able to rely on a factory always returning a valid objects makes writing tests overall more easy - especially when it comes to test setup, the in my opinion hardest part in legacy Rails apps 😉 

**Sidenote**
As a *drive-by-refactoring*  🚜  I also changed a flaky spec and removed to obsolete helper files.


spec/models/conference_preference_spec.rb
```diff
@@ -83,25 +82,29 @@
   end
 
   describe '#terms accepted?' do
```
F3PiX: Should the method name have an underscore? #terms_accepted?

klappradla: Yep. I did not look into the details of specs for this PR 😉 

But I can fix it on the way.

PR #881, opened 2017-12-05
Remove some RSpec legacy stuff

klappradla:
This is a follow up to #880 and #871 to remove leftover of the pre v3 era of RSpec:
1. Inferring the spec type from the file location
  see [the RSpec docs](https://relishapp.com/rspec/rspec-rails/docs/directory-structure) for details on this
1. Have 'describe' available from the main object
  see [global namespace](https://relishapp.com/rspec/rspec-core/v/3-7/docs/configuration/global-namespace-dsl) for details

Motivation for this: I just thing it's a) generally healthy to keep the application up to data with current standards and b) easier to understand for people new to the project then.

PS: I know this touches loads of files... The important parts are the configuration and that the suite is still ✅ 

PS 2: No need to look at this until #880 is agreed upon and I've rebased this branch on it.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -34,6 +34,7 @@ Dir[Rails.root.join('spec/support/**/*.rb')].each { |f| require f }
 ActiveRecord::Migration.maintain_test_schema!
 
 RSpec.configure do |config|
+  config.expose_dsl_globally = false # disable RSpec < 3 global monkey patching
   config.use_transactional_fixtures = false
   config.infer_base_class_for_anonymous_controllers = false
 
@@ -41,8 +42,6 @@ RSpec.configure do |config|
 
   config.include ActiveJob::TestHelper
   config.include RSpecHtmlMatchers
-
-  config.infer_spec_type_from_file_location!
 
   config.before(:suite) do
     WebMock.disable_net_connect!(
--- a/spec/controllers/activities_controller_spec.rb
+++ b/spec/controllers/activities_controller_spec.rb
@@ -1,6 +1,6 @@
 require 'spec_helper'
 
-describe ActivitiesController do
+RSpec.describe ActivitiesController, type: :controller do
   render_views
 
   describe 'GET index' do
--- a/spec/controllers/application_drafts_controller_spec.rb
+++ b/spec/controllers/application_drafts_controller_spec.rb
@@ -1,6 +1,6 @@
 require 'spec_helper'
 
-RSpec.describe ApplicationDraftsController do
+RSpec.describe ApplicationDraftsController, type: :controller do
   render_views
 
   let(:team) { create :team, :in_current_season }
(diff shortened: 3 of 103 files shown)


spec/factories/conference_preferences.rb
```diff
@@ -1,19 +1,16 @@
 FactoryBot.define do
   factory :conference_preference do
-    team { create(:team, :in_current_season) }
-    first_conference { create(:conference, :in_current_season) }
-    second_conference { create(:conference, :in_current_season) }
+    association :team, :in_current_season
```
carpodaster: neat!

spec/factories/season.rb
```diff
@@ -1,6 +1,6 @@
 FactoryBot.define do
   factory :season do
-    name { Date.today.year.to_s }
+    sequence(:name) { |n| "#{Date.today.year}-#{n}" }
```
carpodaster: I think the name has to correspond to a year for its automatic lookup to work, no?

spec/spec_helper.rb
```diff
@@ -34,16 +34,14 @@
...
   config.use_transactional_fixtures = false
   config.infer_base_class_for_anonymous_controllers = false
 
   config.order = 'random'
 
   config.include ActiveJob::TestHelper
   config.include RSpecHtmlMatchers
-  config.extend ControllerMacros, type: :controller
-
-  config.infer_spec_type_from_file_location!
```
carpodaster: I kinda like this auto-detection system 😢

spec/spec_helper.rb
```diff
@@ -34,16 +34,14 @@
 ActiveRecord::Migration.maintain_test_schema!
 
 RSpec.configure do |config|
+  config.expose_dsl_globally = false # disable RSpec < 3 global monkey patching
```
carpodaster: 👍 did not know this config knob existed

spec/factories/season.rb
klappradla: Yes. I even first thought about excluding the `season` model from the factory linting. But then I thought, there are some explicit tests for this functionality in place anyways and the DB is wiped after each test anyways, so I thought it would be fine.

Can undo it if you have concerns about it ✌️

spec/spec_helper.rb
klappradla: 😿 
Interesting / promissing thing about the current setup: it *almost* works with transactional fixtures 🎉 
There have apparently been some changes to Rails' test internals that may allow to ditch `database_cleaner` and switch on `transactional_fixtures` instead. I tried it out - it runs super fast. But unfortunately some tests around the `season` model do not really work with this setup 😉

klappradla:
"Standard RSpec specs" are for POROs:
```ruby
class Thing
  # fun
end

RSpec.describe Thing do
  # fun
end
```

If you have `model` or `controller` spec, rspec-rails includes additional functions, such as `response`, `get` `post`, etc.

spec/factories/season.rb
carpodaster: You're right about the flushing after each example. I'm not sure if anyone looks into the factory definitions to understand the inner workings of a class, but it would confuse me. I guess I'd prefer sticking with `%r(\d{4})` but we should also add a format validation (not necessarily in this PR)

Issue #858, opened 2017-09-18
How to Rubocop: CodeClimate, Travis or ...?


F3PiX:
Dev happiness!

As of #857, we have Rubocop for static code analysis. 🎁 

We can run the analysis locally, with `bundle exec rubocop`, to solve issues before pushing a PR. 
 
With the current implementation, the analysis is part of the CI build. That means that the build will fail if there are Rubocop issues ( is this correct, @klappradla?) 

Another option is using CodeClimate. It has a very easy and beginner-friendly UI. Both for explaining issues and solving them. 

We already use CodeClimate. [Adding Rubocop is easy](https://docs.codeclimate.com/docs/rubocop):  add a codeclimate.yml,  enable rubocop, and then CC uses our rubocop.yml to enforce 🚔  the rules.
Then connect it to the GitHub PR's. 

With CC set up,  we would have 2 checks after submitting/updating a PR: one for the Travis build, and one for CodeClimate.  Which makes it superclear if there are either failing specs, or formatting/style issues. 

![jrducdgrgwghdg6zqmyw_github1](https://user-images.githubusercontent.com/6314687/30536408-fbe47104-9c65-11e7-9d9a-84405d5e897a.png)

Another option could be Travis build stages, see https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/857#discussion_r139322795 
I think that still misses the nice UI that CC offers. 

I'd vote for using CodeClimate, mostly because of the beginner-friendly UI plus the separate process in the PR workflow.  

Votes and thoughts please!

carpodaster:
I'm also very much in favor of using CC for Rubocop.

klappradla:
> That means that the build will fail if there are Rubocop issues

Yes, that's correct.

Some questions on the Code Climate thing (also related to ☝️ ): I don't understand the point of shelling out to a different service for two of the 4 steps in our CI pipeline 🙈  How can Code Climate prevent Travis from deploying to Heroku if a build does not comply to Rubocop or JSHint?

It's true, that things would be more readable if Travis would add a separate link on Github for each step - that's probably a reason why CircleCI is so popular nowadays 😉  - but I would say things are still reasonably easy to find there.

So tl;dr: I don't understand the benefits we get from using Code Climate and fear that it just adds a huge layer of complexity to the otherwise super basic and straight forward CI pipeline.
Maybe I just don't understand the tool yet.

F3PiX:
[...]

PR #883, opened 2017-12-07
Add codeclimate.yml 

F3PiX:
Related issue #858 

Added a codeclimate.yml to make codeclimate do the first Rubocop checks on the PR page.
- I added the `exclude_patterns` that CodeClimate uses by default; these may need fine tuning. 
- I am not 100% sure that this is all we need.


F3PiX:
This is ready for review!

.codeclimate.yml
```diff
@@ -0,0 +1,16 @@
+version: "2"
+plugins:
+  rubocop:
+    enabled: true
+exclude_patterns:
+- "config/"
+- "db/"
+- "dist/"
```
carpodaster: Do we need this exclude pattern? Shouldn't it reuse the ones defined by rubocop?

carpodaster: This comment is actually not about line 8, but the entire block. Sorry, I misclicked :(

F3PiX: I am not sure. 
The reason I added it is because CodeClimate uses the rubocop.yml only for the rubocop check, obviously, but it is doing [other checks as well](https://docs.codeclimate.com/v1.0/docs/default-analysis-configuration#section-maintainability-checks).
And we want to keep those, right? So, with the default settings copied, the behaviour of the Code Quality check as we have it now, should stay the same.

carpodaster: As I understand it, we explicitly only enable rubocop. There are more, yes, and we can consider adding them in the future. Looking at the docs you linked, I see that the exclude pattern is already the default one for CC so I would prefer to remove them to be in sync with what CC thinks best. But I don't have hard feelings about this.

F3PiX: I'll remove them. If we miss them, we'll bring them back.

F3PiX:
Thanks Carsten!!

PR #891, opened 2017-12-09
Update season factory and validations

klappradla:
Quick follow up to @carpodaster's [comment](https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/881#discussion_r155162131) on #881: removing some unnecessary stuff from the `season` factory, use a more reasonable `sequence` there (actual year-strings) and add a validation for the format.

See [this example]( http://rubular.com/r/DfaZcaWuez) for how the regex works 🤓

### DIFF
--- a/spec/factories/season.rb
+++ b/spec/factories/season.rb
@@ -1,9 +1,5 @@
 FactoryBot.define do
   factory :season do
-    sequence(:name) { |n| "#{Date.today.year}-#{n}" }
-    starts_at { Time.utc(Date.today.year, 7, 1) }
-    ends_at { Time.utc(Date.today.year, 9, 30) }
-    applications_open_at { Time.utc(Date.today.year, 3, 1) }
-    applications_close_at { Time.utc(Date.today.year, 3, 31) }
+    sequence(:name, '2000')
   end
 end
--- a/app/models/season.rb
+++ b/app/models/season.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 class Season < ApplicationRecord
 
-  validates :name, presence: true, uniqueness: true, format: { with: /\A\d{4}\z/ }
+  validates :name, presence: true, uniqueness: true, inclusion: { in: ('1999'..'2050') }
 
   before_validation :set_application_dates
 
--- a/app/models/season.rb
+++ b/app/models/season.rb
@@ -1,17 +1,17 @@
 # frozen_string_literal: true
 class Season < ApplicationRecord
 
-  validates :name, presence: true, uniqueness: true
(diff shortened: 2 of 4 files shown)


F3PiX:
Would it make sense to use the auto-creating of Season in the Fabricator?
Instead of validations in the model itself, while the real Seasons are always been autocreated, and the name can't be changed?

I am afraid I can't explain it very well, but generally I hesitate to add logic to the model that is there only for the tests' sake.

F3PiX:
I guess I mean: wouldn't it make sense to validate if the year is an actual and meaningful year.
 
With a custom validator that validates something like:
`year = "2017"`
`("2000".."2050").include?(year)`
Validation will fail for the cases in your examples. And furthermore for cases like "0017", "1700" and "1718".  

WDYT?

klappradla:
Nifty idea using the range @F3PiX 👍 
I can update it to work with that or leave the format validation all together, no strong feelings there. My point was just to have the factory build real-world data and remove unnecessary stuff from there.

F3PiX:
In that case, I would appreciate the range validation! 🙏

spec/factories/season.rb
```diff
@@ -1,9 +1,5 @@
 FactoryBot.define do
   factory :season do
-    sequence(:name) { |n| "#{Date.today.year}-#{n}" }
-    starts_at { Time.utc(Date.today.year, 7, 1) }
-    ends_at { Time.utc(Date.today.year, 9, 30) }
-    applications_open_at { Time.utc(Date.today.year, 3, 1) }
-    applications_close_at { Time.utc(Date.today.year, 3, 31) }
+    sequence(:name, '2000')
```
carpodaster: Ist that a sequence? There‘s no iterator var being evaluated

klappradla: That's because it's a even simpler sequence 😉 : starts at 2000 and just increments by 1.

carpodaster: Interesting, So it `to_i`s it under the hood? Neat.

carpodaster:
Oh, and PS: excellent idea to give a Range, Maud! 💯

spec/factories/season.rb
klappradla: Way simpler. It just calls `next` ✌️ 
Try it yourself:
```ruby
[1] pry(main)> "2015".next
=> "2016"
```
😹  Ruby I guess 😹

carpodaster: Whoa! TIL about `String#(next|succ)`

carpodaster: (on 2nd thought, this is ofc how `"2000".."2050"` works in the first place. D'oh.

PR #893, opened 2017-12-10
Cleanup nested modules and namespaces

klappradla:
Warning upfront: this touches a lot of files ⚠️ 

I made this as a *small*  Sunday project to prepare for updating the layout for orga-users. Turned out it was way harder then expected, since it involved one of the hardest things in CS: **naming things** 🙈 

**The motivation / idea**
Get rid of the "shorthand" syntax for modules in preference of real, actual modules:
```ruby
# from this
class Foo::Bar

# to this
module Foo
  class Bar
```

This also means: it is barely possible (or at least super dangerous) to have a class with the same name as a module on the same "level". So having a model `Mentor` and define a `Mentor::OtherThing` is no longer possible. Therefore I had to rename some things.

**The solution™️**
I could not - and also did not really want to - use the same module for both, nesting models and namespacing controllers, because in my opinion and especially in the Teams App this stands for 2 different things:
1. nesting models: grouping things by domain
1. namespacing controllers: grouping things by user-role / access-level

Therefore I went with domain-fitting (happy on suggestions there still...) names for model namespaces and used the plural role for controller namespaces since this also matches the general naming of routes and controllers and won't collide with our model modules _(we're not going to use plural names for model modules if they aren't encapsulating collections)_

Once I did ☝️ for one role I though it looks weird if I would not align all the roles 🙈 . I first thought about keeping "orga" as "orga". Then I tried to add the "orga" role to my user with `name: 'orga'` and realized it would probably be better to also align this 😹 

**TL;DR**
Grouping some models by domain and namespacing respective controllers with the plural name of the role which is allowed in.
I know these are a lot of in theory unnecessary changes, but I think it makes the controller schema quickly to understand when e.g. starting from the routes and gives us sort of "futureproof" foundation for now.


app/models/dev_utils/season_phase_switcher.rb
```diff
@@ -0,0 +1,69 @@
+# frozen_string_literal: true
+module DevUtils
```
F3PiX: 😍

F3PiX:
Wow!! 🥇 
This _must_ have been hard, because the result looks deceptively simple. :code_perfume:
I love the new controller namespaces, also the `DevUtils` module.

The only two _names_ I am not so sure about: `Selection` and `Services`. After reading your explanation, I see why you use the name `Services`, so if you don't have reservations on this one, I am fine. (Does the 'classic' service object have a controller, though?).
About `Selection`: we have two types of selection: the projects and the teams. The formerly-known-as-ratings were only about the team selection process, right? How about naming it `TeamSelection`?

And the new routing file 😍

klappradla:
Valid point about `Selection` @F3PiX - thought about this as well... renaming to `TeamSelection` could be an option 👍 

For `Services`: that's not a module 😉 
It's simply a new directory under `app/services` that hold callable objects that encapsulate certain business logic - some people also call them *"service objects".

Services aren't *models*  or *resources* in general and also don't have a corresponding controller. They just contain a piece of business logic to make these bits of code easier to test, share etc. and better describe the domain.
Being callable objects, you'd instantiate them with all the basic information they need to do their operation and then just "call" them via their only public method: `call` ✌️ 

I usually implement them like this:
```ruby
class Service
  def initialize(user)
    @user = user
  end

  def call
    # have fun
  end

  private

  attr_reader :user
end
```

And then e.g. cal them like this:
```ruby
statistics = StatisticsComputer.new(user: user).call
```

The conference importer that I moved over is implemented *slightly* different but still looks and behaves very much like a service to me - that's why I went that way.

klappradla:
Just wondering 🤔 
Should I fix the issues code climate raised in this PR or in a follow up? Technically speaking they don't originate from this PR since I just renamed files and paths 😉 

But I probably should not merge with code climate being 🔴 ...

F3PiX:
I wouldn't fix them, we did not even decide if we keep the maintainability checks. That is something CC decided to offer us ;-)
Normally, there would be a dropdown if you follow the `details` link, and hover over the `new` tag.
I don't see that myself, though, in this repo. Maybe because of the rights I have?
Can you access the dropdown menu? 
If so, just set the issue status to 'won't fix'. It means: it's a legit issue, but not fixing it right now. 

Either way. I'd like to cleanup the duplication in the season switch anyway. Could you leave them to me :-) ?

klappradla:
Ok, so `won't fix` means we'll find it again afterwards? Because I agree with what the tools says and would for sure fix it - but it would make this PR even huger. I'd like to stay with just renaming files in this PR for now 😉

F3PiX:
Well, you can find them in the 'issues' tab on CodeClimate. It won't bark if you are not going to look for them. (Unless you touch the same file again.) Or create an issue on GH. 
"won't fix" is exactly meant for what you want: Yes it needs fixing, just not in this PR.

carpodaster:
Hey @klappradla. Before I get to a "real" review: did you use `git mv`? The diff looks like Git didn't recognise the move which makes the changeset very hard to read and will also lose our git history.

Also, I approved the CC complaints since we're just moving files here and not adding anything new.

klappradla:
Oh, good point @carpodaster! I tbh didn't even know about `git mv` 🙈 and also did not look into the details of the changeset...
However, the way I understand it, `git mv` is only a shortcut for moving a file and then `git add` the changes in one command. So I think the history of files should still be fine - it's just that Github cannot display the changes in another way apparently...

If you for instance look at an apparently new file, you'll still see the history - [see this for example](https://github.com/rails-girls-summer-of-code/rgsoc-teams/blame/2ee67565a11d1acd4114673beb0bcbf212da9dad/app/controllers/mentors/applications_controller.rb).

Do you know of anything I could do to make the changeset easier to read or pick up the changes in a different way? Would `git rm` do this?

klappradla:
I'll merge this for now @carpodaster to not get into troubles with other PRs touching multiple files, such as the approaching one for translating error messages 🙈 

If you find the time to look into the details of this PR around renaming namespaces and routes, you can still comment here and I'll address everything in follow up PRs to have this 🗻 out the way ✌️

F3PiX:
Yay!!!! 🏅 🎇  Fireworks are early this year!

carpodaster:
@klappradla sorry I did not come back to this PR, I'm a bit slow … again 😞  Good that you moved ahead – and indeed: early fireworks :)

PR #898, opened 2017-12-28
Fix which navigation items to show when and how to mark them "active"

klappradla:
This fixes the problem of a) having the same items multiple times in the dropdown navigation when being between seasons and b) having multiple items marked as `active` in the dropdown.

**Before**
screen shot 2017-12-27 at 19 29 07

**After**
screen shot 2017-12-28 at 14 02 18

This is specifically about hiding the `projects` and `conferences` item again after the season has ended and we're transitioning to the next season.

I will add a follow-up PR with better feature-testing the navigation.

### DIFF
--- a/app/helpers/nav_helper.rb
+++ b/app/helpers/nav_helper.rb
@@ -1,13 +1,17 @@
 # frozen_string_literal: true
 module NavHelper
-  SOC_CONTROLLER = %w(conferences projects applications application_drafts teams users)
+  SOC_CONTROLLER = %w(conferences projects applications application_drafts teams community).freeze
 
-  def dropdown_active?
+  def active_if_soc_dropdown_active
     'active' if SOC_CONTROLLER.include?(params[:controller])
   end
 
-  def active?(controller_name)
-    'active' if params[:controller] == controller_name.to_s
+  def active_if(path)
+    'active' if current_page?(path)
+  end
+
+  def active_if_controller(controller)
+    'active' if params[:controller] == controller
   end
 
   def during_application_phase?
--- a/app/helpers/nav_helper.rb
+++ b/app/helpers/nav_helper.rb
@@ -9,4 +9,13 @@ module NavHelper
   def active?(controller_name)
     'active' if params[:controller] == controller_name.to_s
   end
+
+  def during_application_phase?
+    current_season.application_period? ||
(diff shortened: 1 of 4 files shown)


app/views/application/_nav_right.html.slim
```diff
@@ -1,27 +1,27 @@
...
-        li class=active?(:projects) = link_to 'Projects', projects_path      
-      - if show_application_link?
+        li class=active_if(new_project_path) = link_to 'Submit your Project', new_project_path
+        li class=active_if(projects_path) = link_to 'Projects', projects_path
+      - elsif during_application_phase?
         li.divider.hidden-xs
-        li class=active?(:application_drafts) = application_disambiguation_link
-        li class=active?(:projects) = link_to 'Projects', projects_path
-      - if current_season.started?
+        li class=active_if_controller('application_drafts') = application_disambiguation_link
```
F3PiX: Is there a reason why you can't use the application_drafts_path here?
Because this one seems to be the only one that uses the `active_if_controller` helper.  Maybe we can do without that, or...?

app/helpers/nav_helper.rb
```diff
@@ -1,12 +1,25 @@
 # frozen_string_literal: true
 module NavHelper
-  SOC_CONTROLLER = %w(conferences projects applications application_drafts teams users)
+  SOC_CONTROLLER = %w(conferences projects applications application_drafts teams community).freeze
```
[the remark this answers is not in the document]
klappradla: The reason why I used this prefix: the menu / nav item this is about has the name `Summer of Code`, plus it's basically a global constant 😉 

So I would go for the plural version, yes, but I would keep some prefix to have it still closely tied to the menu item.

app/views/application/_nav_right.html.slim
klappradla: The `application_disambiguation_link` helper can return two different paths, depending on whether the user already has an `application_draft` or not. That's why a matching on the exact path is not possible 😞 
Both paths originate from the same controller though, that's why the controller approach still works here.

F3PiX: Brilliant.

F3PiX:
🎉  (LOL that I missed the wip tag )

PR #900, opened 2017-12-30
Don't show a "my team" button to guest users / users without a team

klappradla:
This fixes a bug on the team index page (likely only visible in production): both guest users and users who aren't students or part of a team see a `my team` button which will bring them to the show view for some random team where at least one student deleted their account **(we should try to avoid getting these invalid team records in the first place...)**

Looked like this _(see I'm not signed in)_:
screen shot 2017-12-28 at 12 22 03

Tiny bit of background: the `current_student` helper defined in the `ApplicationController` would always return a `Student`, even if there's no `current_user` (the student would then be called `nil`) and would therefore always find a `current_team` - one with a `user_id` being nil.

### DIFF
--- a/app/views/teams/index.html.slim
+++ b/app/views/teams/index.html.slim
@@ -1,6 +1,6 @@
 nav.actions
   ul.list-inline
-    li = link_to 'My Team', current_student.current_team, class: 'btn btn-default btn-sm' if current_student.current_team
+    li = link_to 'My Team', current_student.current_team, class: 'btn btn-default btn-sm' if current_student&.current_team
     li = link_to icon('plus', 'New Team'), new_team_path, class: 'btn btn-primary btn-sm' if can? :create, Team.new(season: Season.current)
   div.pull-right.dropdown
     button.btn.btn-default.dropdown-toggle data-toggle="dropdown"
@@ -12,6 +12,3 @@ nav.actions
          a data-behaviour="switch-teams-year" data-value=year =year
 
 = render 'table', namespace: nil
-
-
-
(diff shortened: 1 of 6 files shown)


app/views/teams/index.html.slim
```diff
@@ -1,6 +1,6 @@
 nav.actions
   ul.list-inline
-    li = link_to 'My Team', current_student.current_team, class: 'btn btn-default btn-sm' if current_student.current_team
+    li = link_to 'My Team', current_student.current_team, class: 'btn btn-default btn-sm' if current_student&.current_team
```
F3PiX: We have a `:student_team method` in User.rb. I'd think you can use that instead?
Also, you could use the [`link_to_if`](http://api.rubyonrails.org/classes/ActionView/Helpers/UrlHelper.html#method-i-link_to_if) helper. 
OTOH.... shouldn't coaches and supervisors have access to the button as well?

klappradla: Good idea! I'll switch to `student_team` - and can probably fully get rid of the method in `Student`...

For `link_to_if`: I personally don't find it fitting the use case here. The helper **always** returns something (e.g. a string) and I explicitly don't want to display anything if the condition is false.... I could accomplish this with this helper as well of course (passing an empty block) but it's not really what it is intended for 😉 

For whether coaches and supervisors should see the button too: I don't think so.
I don't remember why the button was added, but it looks like a "shortcut" for students to find their team. Coaches and Supervisors are likely to be in multiple teams in the same season, so the button does not make too much sense for them I'd say.

klappradla: Oh, on the other hand: `student_team` is not used anywhere. I could also just get rid of it and stick with `current_team`... 🤔

klappradla: Ok, went with what I said above @F3PiX ☝️

F3PiX:
I totally agree with all your conclusions, thanks for the explanations. 
🎆

Issue #901, opened 2018-01-02
Incorrect author name for a blog post 


teambundledore:
In the blog post feed, under the title of the article, the correct name is not displayed. I am sorry if this has already been reported, please close the issue if you feel it has already been reported. Attaching a screenshot. 
![screenshot from 2018-01-02 18-51-54](https://user-images.githubusercontent.com/29810436/34485170-1d6b4b5e-efef-11e7-8ef2-ce6776b57007.png)

nerdyvaishali:
I think this was reported in teams app. There was an issue with one team id where instead of team's daily logs, blog posts were appeared. I remember abourt some discussion, but I don't remember if it was solved or not.

F3PiX:
@nerdyvaishali There was another issue with the blog posts, but that was a bit different. 

I'll follow up later with a description of the problem.

klappradla:
It's 403s for all the images. I remember we had something similar when we tried to switch to the new AWS gem...

carpodaster:
The root problem seems to be that a team has updated the list of web resources in their profile with a direct link to their blog post on rgsoc.org. Our feed url detection does the rest:

```ruby
require 'feed'
source = Source.new(url: "https://railsgirlssummerofcode.org/blog/2017-08-07-two-geeks-and-a-summer-of-code")
feed = Feed.new(source)
feed.send(:discover_feed_url)
# => "http://railsgirlssummerofcode.org/blog.xml"
```

nerdyvaishali:
@F3PiX I found it I guess. It is issue 853: https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/853 

The problem seems to be with the same team's team id.

F3PiX:
@carpodaster @nerdyvaishali 

I am not sure that is the problem here. These blogposts are by the orga's. 
  
![schermafbeelding 2018-01-03 om 11 29 08](https://user-images.githubusercontent.com/6314687/34516968-5d23fe78-f079-11e7-87df-c18c6613fb15.png)

I was thinking to look into two things:
1) the :display team name method, when there is not a team 
2) the way we pull in blogposts from the orga website.

carpodaster:
@F3PiX the blog posts are posted by Team prodyoGEEKY and as you can see on [thir profile page](https://teams.railsgirlssummerofcode.org/teams/338), they are the ones referencing a railsgirlssummerofcode.org blog post. It was not added to the Teams App by the orga team.

[...]

PR #904, opened 2018-01-03
Fix/allow single page web source

carpodaster:
Related issue #901 

While this won't block the error outlined in the referenced issue, it does offer a chance to add a URL to a team's blog that doesn't trigger a synchronisation of all available articles available on that blog site.

It removes the `SourcesController` to remove some duplication. Only the CUD actions where used but their functionality is covered by the `Team` model's `accepts_nested_attributes`.

It displays a conditional help block when the "blog" source type is selected. It is only displayed when the radio button is _changed to that value_. The JS I wrote for this is probably very, very naive and I'm looking forward to your suggestions to make it better :smile: 

screenshot-2018-1-3 rgsoc - teams app


app/assets/javascripts/teams.js
```diff
@@ -0,0 +1,15 @@
+$(document).ready(function() {
+    $('#team-sources input[id^=team_sources_attributes]:radio').change(function() {
```
klappradla: You could use a data-attribute instead of the ID. From my experience, this usually makes the intention clearer (especially given the fact that we don't test JS code 🙈 ). So maybe: `data-behavior="source-switcher"` or something like this.

app/assets/javascripts/teams.js
```diff
@@ -0,0 +1,15 @@
...
+        var helpText,
+            helpNode = $($(this).parents()[2]).find('.help-block');
```
klappradla: Same here, I'd suggest to use a data attribute as well - makes selection easier.

app/views/teams/_form.html.slim
```diff
@@ -53,10 +53,9 @@
         Please make sure to add at least your team log feed and repositories that
         you are working on:
 
-    fieldset
+    fieldset id="team-sources"
```
klappradla: As said before, I'd suggest to switch to data attributes and put it directly on the radio button rather than this high up in the DOM.

F3PiX:
Just checking:  this doesn't need the actual `accepts_attributes_for :sources` on the team model ?

carpodaster:
> Just checking:  this doesn't need the actual `accepts_attributes_for :sources` on the team model ? 
 
@F3PiX yes it does. It's included here: https://github.com/rails-girls-summer-of-code/rgsoc-teams/blob/f36d5804e2afe43688b77984995645275b646786/app/models/team.rb#L36

And here: https://github.com/rails-girls-summer-of-code/rgsoc-teams/blob/f36d5804e2afe43688b77984995645275b646786/app/views/teams/_form.html.slim#L57

F3PiX:
Sorry!  🙈

carpodaster:
Thanks for the suggestions, @klappradla. I moved to data-behaviour attributes. Better now? As for the four spaces: I was told this is how it's done in the JS world. Is that not the case?

It will not immediately fix the problem, but we can edit the "offending" blog source to be a `page` instead. After that we'd have to remove all the feed entries for that blog on the prod console but after that, we'd have our normal activity feed back.

But as mentioned in the PR description: if anyone adds a URL with kind `blog` were not every article belongs to them, it will start all over. Not sure if we can fix that without disabling the RSS/Atom feed auto-detection in `lib/feed.rb`

klappradla:
Well "better" is kinda relative there, since this mostly comes down to personal preference - I think data-attributes are a better fit for these things however. This leaves IDs for styling and QA ✌️ 

About 👾 : I'm probably following the AirBnB style-guide _([here the section on whitespace](https://github.com/airbnb/javascript#whitespace))_ which does not do the 4 space thing. But as always with Javascript, there are millions of opinions out there 😹  So no worries.

Are you also taking care of the follow-up hotfixes on production?

carpodaster:
Offtopic for the PR, but do we want to enforce a whitespace style?

And yes, I will do the prod console dance tomorrow.

klappradla:
For now not.

I hope to find time for the Webpacker and ES6 topic in the next months. If we have this setup in place, we can utilize [prettier](https://github.com/prettier/prettier) to do all this automatically in a pre-commit hook 😎

PR #905, opened 2018-01-05
Add rails_helper

carpodaster:
This PR moves the Rails specific config to `spec/rails_helper.rb`. It's the default for the rspec-rails gem for quite a while now and also enables unit-testing without loading the Rails environment (= takes looong) by just requiring `spec/spec_helper.rb` (which is indeed required automatically by `.rspec` now).


spec/spec_helper.rb
```diff
@@ -34,7 +34,7 @@
     # Prevents you from mocking or stubbing a method that does not exist on
     # a real object. This is generally recommended, and will default to
     # `true` in RSpec 4.
-    mocks.verify_partial_doubles = true
+    mocks.verify_partial_doubles = false
```
carpodaster: I've created a follow up issue: #906

spec/rails_helper.rb
```diff
@@ -0,0 +1,72 @@
...
+    [SimpleCov::Formatter::HTMLFormatter, Coveralls::SimpleCov::Formatter]
+  )
+  SimpleCov.start 'rails' do
+    add_filter '/spec'
+    add_group 'Models', 'app/models'
+    add_group 'Controllers', 'app/controllers'
+    coverage_dir File.join('coverage', Time.now.strftime('%Y%m%d-%H%M%S'))
+  end
+end
+
```
F3PiX: Would you mind to add a comment that the Coverage things need to be at the top of the file? 
I had to look that up, because it looks contrary to line 26 "Add additional requires below this line".

F3PiX: 🙇  thanks :-)

carpodaster:
@F3PiX I have no hard feelings for or against moving the config calls in `spec/support`. I wouldn't do this in this PR, though.

PR #907, opened 2018-01-05
Create an error reporting facade

carpodaster:
While exception tracker libraries tie into Ruby's/Rails' system automatically for uncaught exceptions, using the exception tracker's log/system/dashboard for rescued errors requires special API calls.

We still haven't figured out which exception tracker we want to use (we've used Sentry, Honeybadger, and Rollbar so far). This simple class makes it easy to swap error logging providers for manual error reports.

Note: the explicit `require` statements are for a seemless tie-in with PR #905. 

**Edit**: The `Raven.capture_exception` does work:

bildschirmfoto vom 2018-01-07 16-34-08

### DIFF
--- a/spec/lib/error_reporting_spec.rb
+++ b/spec/lib/error_reporting_spec.rb
@@ -1,0 +1,25 @@
+require 'spec_helper'
+require 'error_reporting'
+
+RSpec.describe ErrorReporting do
+  describe '.call' do
+    let(:raven_double) { double(capture_message: :ok, capture_exception: :ok) }
+
+    before do
+      stub_const('Raven', raven_double)
+    end
+
+    let(:error) { Class.new(Exception).new("Nope.") }
+    let(:strng) { "Err, what about … no?" }
+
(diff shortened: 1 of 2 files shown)


carpodaster:
For an application of this, see #908

spec/lib/error_reporting_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+RSpec.describe ErrorReporting do
+  describe '.call' do
+    let(:raven_double) { double(capture_message: :ok, capture_exception: :ok) }
+
+    before do
+      stub_const('Raven', raven_double)
+    end
+
+    let(:error) { Class.new(Exception).new("Nope.") }
+    let(:strng) { "Err, what about … no?" }
```
klappradla: Is this typo on purpose? Or is this nowadays already internetz-slang enough to no longer be considered a typo?

carpodaster: haha, this wasn't a typo, but my way of having both statements aligned ;)

Issue #903, opened 2018-01-03
Investigate possible AWS errors


carpodaster:
Follow-up to #853 where @klappradla [pointed out](https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/853#issuecomment-329408374) that the 403s on S3 feed images might be related to the updated AWS gem.

PR #908, opened 2018-01-07
Slightly improve feed handling (and allow better error inspection)

carpodaster:
**Note**: This PR is based on #907. Please do not merge before that one.

* Removes `autoload`ing
* Becomes specific about expected `option` keys
* Related issue #903 - tracks error in our exception tracking service for later inspection


lib/feed.rb
```diff
@@ -1,12 +1,13 @@
 # frozen_string_literal: true
 require 'feedjira'
 
-class Feed
-  autoload :Discovery, 'feed/discovery'
-  autoload :Image,     'feed/image'
-  autoload :Item,      'feed/item'
-  autoload :S3,        'feed/s3'
+require 'error_reporting'
```
klappradla: 👍

lib/feed.rb
```diff
@@ -17,9 +18,11 @@ def update_all
 
   attr_reader :source, :logger
 
-  def initialize(source, options = {})
+  # @param source [Source]
+  # @param logger [Logger] an optional logger instance. Logs to STDOUT by default.
+  def initialize(source, logger: Logger.new(STDOUT))
```
klappradla: 💚

PR #909, opened 2018-01-09
Update Gems

beanieboi:
Updating our dependencies.

I needed to fix 2 little things to make the tests pass. shoulda-matchers has a configuration now and the order for some validations is important now.

cheers


carpodaster:
Great, thanks! Aside from the specs being green, I pushed it to staging and clicked around a bit – everything seems to be working

PR #910, opened 2018-01-16
Feature/browse past projects

carpodaster:
**What?**
Adds the ability to browse past projects. For past seasons, only **accepted and assigned** projects will be shown, just as we do for the current season once it started.

The orga-version of the projects view hasn't been touched. No need to mingle in the past ^^

desktop-animation

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -25,11 +25,12 @@ class ProjectsController < ApplicationController
   end
 
   def index
-    if Season.current.active?
-      @projects = Project.selected
-    else
-      @projects = Project.in_current_season.where(aasm_state: %w(accepted proposed))
-    end
+    season = Season.find_by(name: params['filter']) || Season.current
+    @projects = if season.current? and !season.active?
+                  Project.in_current_season.not_rejected
+                else
+                  Project.selected(season: season)
+                end
   end
 
   def destroy
--- a/app/models/concerns/has_season.rb
+++ b/app/models/concerns/has_season.rb
@@ -4,5 +4,12 @@ module HasSeason
 
   included do
     belongs_to :season
+
+    # @param season [#name, String, Integer]
+    # @return [ActiveRecord::Relation]
+    def self.in_season(season)
+      season = season.try(:name) || season
+      joins(:season).where('seasons.name' => season)
+    end
   end
 end
--- a/app/views/projects/index.html.slim
+++ b/app/views/projects/index.html.slim
@@ -7,6 +7,10 @@ h1.header
   = icon('users')
   span Projects
 
+nav.past-projects
+  - project_years.each do |filter|
+    = link_to filter.titleize, projects_path(filter: filter), class: "label label-default"
+
 table.table.table-striped.table-bordered.table-condensed.hidden-sm.hidden-xs
   tr
     th Name
--- a/app/assets/stylesheets/controllers/projects.sass
+++ b/app/assets/stylesheets/controllers/projects.sass
@@ -30,3 +30,6 @@
 
 .space_btns
   margin-left: 10px
+
+nav.past-projects
+  margin-bottom: 10px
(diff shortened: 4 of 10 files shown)


app/models/concerns/has_season.rb
```diff
@@ -4,5 +4,12 @@ module HasSeason
 
   included do
     belongs_to :season
+
+    # @param season [#name, String, Integer]
+    # @return [ActiveRecord::Relation]
+    def self.in_season(season)
+      season = season.try(:name) || season
```
klappradla: Personal preference, but I'd find
```ruby
name = season&.name || season
```
easier on the eyes... Plus: I think it's good practice to stop using `try` since it's basically in the language now 😉

app/views/projects/index.html.slim
```diff
@@ -7,6 +7,10 @@ h1.header
   = icon('users')
   span Projects
 
+nav.past-projects
+  - project_years.each do |filter|
+    = link_to filter.titleize, projects_path(filter: filter), class: "label label-default"
```
klappradla: Using these labels is a nice idea 💡 I'd even suggest to align the `teams` page to this pattern and remove the dropdown there in favor of these labels _(would also help us spare some untested JS code there)_

klappradla:
Build is failing for reasons I don't see though 🙈

app/models/concerns/has_season.rb
carpodaster: The argument "union type" requires the `try`. There is a difference in `try` and `&`, they're not the same thing:

Object#try is
```ruby
def try(meth)
  send(meth) if respond_to?(meth)
end
```

The safe navigation is:
```
send(meth) unless nil?
```

```
$ rails c
Loading development environment (Rails 5.1.4)
2.5.0 :001 > RUBY_VERSION
 => "2.5.0"
2.5.0 :002 > "".try(:name)
 => nil
2.5.0 :003 > nil.try(:name)
 => nil
2.5.0 :004 > nil&.name
 => nil
2.5.0 :005 > ""&.name
Traceback (most recent call last):
NoMethodError (undefined method `name' for "":String)
```

app/controllers/projects_controller.rb
```diff
@@ -25,11 +25,12 @@ def edit
   end
 
   def index
-    if Season.current.active?
-      @projects = Project.selected
-    else
-      @projects = Project.in_current_season.where(aasm_state: %w(accepted proposed))
-    end
+    season = Season.find_by(name: params['filter']) || Season.current
+    @projects = if season.current? and !season.active?
```
F3PiX: Does it make sense to make this a 'pre-season' method?
And doublecheck: we need `and` instead of `&&` ? (No need for explanation ;-), just checking)

app/models/concerns/has_season.rb
F3PiX: 🙃  I didn't realize this. Thnks

app/controllers/projects_controller.rb
carpodaster: Pre-season: there is `Season#transition?` already. Is that what you mean?

`and` vs. `&&`: I never know what to use. I know I will read up on that after I click the "Comment" button. I know I will have forgotten about it tomorrow   ;)

F3PiX: `Season#transition` is from october 1 through dec 31, and I think you want jan 1 through `active`. ? 

The way I remember it is: Use `&&` unless you have a very specific goal for `and`.  
http://www.virtuouscode.com/2014/08/26/how-to-use-rubys-english-andor-operators-without-going-nuts/

carpodaster: `and`/`or`  for control flow! Currently tattooing this in my knees! 😄   Thanks!

app/models/concerns/has_season.rb
klappradla: Oh, did not think this through 🙈 thanx for pointing out @carpodaster 🙏

alicetragedy:
:partyparrot:

app/models/concerns/has_season.rb
ramonh: Whoa, thanks! Didn't know this 👍

ramonh:
Nice one! Thanks! 🎉

Issue #882, opened 2017-12-07
Limiting the number of coaches per team


nerdyvaishali:
Often we see that participants end up adding multiple coaches and not all of them are quite involved during the program. So, as discussed in the offsite meeting how about adding a feature in teams app which puts a limitation on adding a number of coaches per team [ideally 4 I assume]?

One of the other thing, which is slightly related to this is we want to allow project maintainers to add coaches this year. So, here is my idea of how things may look like at the end:

- For the project submission form, add in the description field that project maintainers can list coaches [basically just a better decsription for description field] and once the project is accepted I think coaches/co-maintainers can sign up as coaches.
- Add feature to teams app that doesn't allow more than 4 (?) coaches per team. These 4 coaches can be either from students or from projects. 

... But as we allow 2 project choices from students, it may possible that some project will have coaches from their side and some projects will not. And in that case how does this limitation feature will work? Should we implement this as a per team, per project feature? Any other ideas to implement this in a better way?

klappradla:
Thanx for reporting this 💡 

**Limit number of coaches**
I get the problem of coaches not being involved enough into the program, but I don't understand how limiting the number of coaches solves or at least touches this problem 🙈 
Let's say a team picks 10 coaches and only 1 of them does actual *"work"*. If we now limit the number of coaches to 4, the team may pick 4 out of the 9 non-involved once and end up with *"no coaching"* at all. Or am I getting this wrong?

I agree that it does not make sense to add more then 4 coaches, plus that any number above that is likely to decrease quality of coaching. I would also add the respective logic for that, but I don't think it solved this problem:
> participants end up adding multiple coaches and not all of them are quite involved during the program

**Project-bound-coaches**
I think with the new community page (added as part of last year's summer of code) something along the lines of what you describe is already possible. It was actually one of the motivations for building the page - to allow students who don't already have a super strong coaching team to their side to look for people who are willing to coach and have a background in the technologies involved.

The only difference: coaches have to sign up **before** someone fills out the project application.

I'll quickly outline the scenario:
1. Mentor decides to submit project
1. All the coaches she has to her side sign up as users on the Teams App, confirm their profile, set their status to `available for coaching` _(or however it is phrased, don't know it by heart)_ and cite that they are proficient in the technologies involved in the project
[...]

Issue #878, opened 2017-11-21
Move error messages to translation file


F3PiX:
We have some custom error messages, for example here.
https://github.com/rails-girls-summer-of-code/rgsoc-teams/blob/db178cfe9d65c5c9f4c58e4bcf0d33148a221177/app/models/application_draft.rb#L141-L145

We want to move these custom error messages to the translation file `en.yml`. In [this commit ](https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/875/commits/acbd0b55a8a6661bd467c182a525af45d2a27e25)you can see how it is done, for an error message on the Todo model and its user attribute.

Collect all the error messages and put them into the translation file, at the right place.

F3PiX:
@aSquare14 The best way to start is by selecting one of the issues with the 'beginner-friendly' label. Like this issue (or click on the yellow label, to see all issues with that label). Even if you are not a beginner at coding, they are suitable for first time contributors. 
Does this work for you?
  
After that, if you are looking for something more challenging, you can take a look at the issues labeled 'rgsoc18'. I wouldn't recommend all of them, but we can chat about that later. 

To learn more about projects: see https://railsgirlssummerofcode.org/guide/projects/ 
I don't know much more that what is in the guides ;-) , there is an email address in the guide for further inquiries.

aSquare14:
Thank You !

nileshgulia1:
Hey @F3PiX ,I would like to work on the  issue,will send a PR soon :)

F3PiX:
Hi @nileshgulia1 Welcome at the Teams App!!! Great that you want to contribute. 👍  🎉 

@aSquare14 Are you currently working on this issue? Or can we pass it to @nileshgulia1 ?

@nileshgulia1 I'd hate to see you do double work for nothing, so please wait a few days until we know if @aSquare14 is working on this.

aSquare14:
Nopes I'm not. He's free to work on it. All the best!

F3PiX:
@nileshgulia1 All yours!!! 💃 

@aSquare14 Thanks for your swift reaction! Much appreciated.

F3PiX:
@nileshgulia1 One more thing: be sure to _first_ pull in the latest changes - we just merged a huge PR - . Then create your branch, so that your commits are on top of the changes. 
This will save you a lot of merge conflicts later on. 
 
(Ask away if you need more explanation/help.)

nileshgulia1:
[...]

PR #913, opened 2018-01-22
Feature/882 team coach limit

ramonh:
Related issue #882 

This aims to limit the number of coaches a team can have to four.

### DIFF
--- a/app/models/team.rb
+++ b/app/models/team.rb
@@ -11,6 +11,7 @@ class Team < ApplicationRecord
   validate :disallow_multiple_student_roles
   validate :disallow_duplicate_members
   validate :limit_number_of_students
+  validate :limit_number_of_coaches
 
   attr_accessor :checked
 
@@ -187,6 +188,12 @@ class Team < ApplicationRecord
     students = roles.select{|r| r.name == 'student' && !r.marked_for_destruction?}
     return unless students.size > 2
     errors.add(:roles, 'there cannot be more than 2 students on a team.')
+  end
+
+  def limit_number_of_coaches
+    coaches = roles.select{|r| r.name == 'coach' && !r.marked_for_destruction?}
+    return unless coaches.size > 4
+    errors.add(:roles, 'there cannot be more than 4 coaches on a team.')
   end
 
   def two_students_present?
--- a/app/models/team.rb
+++ b/app/models/team.rb
@@ -193,7 +193,7 @@ class Team < ApplicationRecord
   def limit_number_of_coaches
     coaches = members_with_role 'coach'
     return unless coaches.size > 4
-    errors.add(:roles, 'there cannot be more than 4 coaches on a team.')
+    errors.add(:roles, :too_many_coaches)
   end
 
   def members_with_role(role)
--- a/app/models/team.rb
+++ b/app/models/team.rb
@@ -185,15 +185,19 @@ class Team < ApplicationRecord
   end
 
   def limit_number_of_students
-    students = roles.select{|r| r.name == 'student' && !r.marked_for_destruction?}
+    students = members_with_role 'student'
     return unless students.size > 2
     errors.add(:roles, 'there cannot be more than 2 students on a team.')
   end
 
   def limit_number_of_coaches
-    coaches = roles.select{|r| r.name == 'coach' && !r.marked_for_destruction?}
+    coaches = members_with_role 'coach'
     return unless coaches.size > 4
     errors.add(:roles, 'there cannot be more than 4 coaches on a team.')
+  end
+
+  def members_with_role(role)
+    roles.select{|r| r.name == role && !r.marked_for_destruction?}
   end
 
   def two_students_present?
--- a/app/models/team.rb
+++ b/app/models/team.rb
@@ -185,14 +185,12 @@ class Team < ApplicationRecord
   end
 
   def limit_number_of_students
-    students = members_with_role 'student'
-    return unless students.size > 2
+    return unless members_with_role('student').size > 2
     errors.add(:roles, 'there cannot be more than 2 students on a team.')
   end
 
   def limit_number_of_coaches
-    coaches = members_with_role 'coach'
-    return unless coaches.size > 4
+    return unless members_with_role('coach').size > 4
     errors.add(:roles, :too_many_coaches)
   end
 
--- a/spec/models/team_spec.rb
+++ b/spec/models/team_spec.rb
@@ -156,7 +156,7 @@ RSpec.describe Team, type: :model do
     let(:remove_coach) { { name: 'coach', team_id: team.id, user_id: create(:user).id, _destroy: true } }
 
     context 'when team has no coaches yet' do
-      it 'allows to add 2 new coaches' do
+      it 'allows to add 4 new coaches' do
         expect {
           team.update roles_attributes: coach_attributes
         }.to change { team.members.count }.by 4
--- a/spec/models/team_spec.rb
+++ b/spec/models/team_spec.rb
@@ -147,6 +147,86 @@ RSpec.describe Team, type: :model do
     end
   end
 
+  describe 'limit of coaches' do
+    let(:team) { create :team }
+    let(:new_user) { create :user }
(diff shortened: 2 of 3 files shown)


app/models/team.rb
```diff
@@ -184,11 +185,21 @@ def disallow_duplicate_members
...
-    students = roles.select{|r| r.name == 'student' && !r.marked_for_destruction?}
+    students = members_with_role 'student'
     return unless students.size > 2
     errors.add(:roles, 'there cannot be more than 2 students on a team.')
   end
 
+  def limit_number_of_coaches
+    coaches = members_with_role 'coach'
+    return unless coaches.size > 4
+    errors.add(:roles, 'there cannot be more than 4 coaches on a team.')
```
F3PiX: Can you please add the error to the locale? 
https://github.com/rails-girls-summer-of-code/rgsoc-teams/blob/3fd55ebe3b221273bb0112bdbcb0cae783e6c471/config/locales/en.yml#L86-L91

F3PiX: See https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/878

ramonh: Thanks for the feedback! I'm on it

ramonh: Hey @F3PiX, I've added the localisation as discussed. 

I've left it out for the student limit error, as I'm not sure it makes sense to add it to this PR. 

What do you think?

app/models/team.rb
```diff
@@ -184,11 +185,21 @@ def disallow_duplicate_members
...
     errors.add(:roles, 'there cannot be more than 2 students on a team.')
   end
 
+  def limit_number_of_coaches
+    coaches = members_with_role 'coach'
+    return unless coaches.size > 4
+    errors.add(:roles, :too_many_coaches)
+  end
+
+  def members_with_role(role)
```
carpodaster: Oh great, that's way more idiomatic!

app/models/team.rb
```diff
@@ -184,11 +185,21 @@ def disallow_duplicate_members
...
   def limit_number_of_students
-    students = roles.select{|r| r.name == 'student' && !r.marked_for_destruction?}
+    students = members_with_role 'student'
     return unless students.size > 2
     errors.add(:roles, 'there cannot be more than 2 students on a team.')
   end
 
+  def limit_number_of_coaches
+    coaches = members_with_role 'coach'
+    return unless coaches.size > 4
```
carpodaster: Can you pull that on one line, eliminating the need for the local variable?

spec/models/team_spec.rb
```diff
@@ -147,6 +147,86 @@
     end
   end
 
+  describe 'limit of coaches' do
+    let(:team) { create :team }
+    let(:new_user) { create :user }
+    let(:new_coach) { -> {{ name: 'coach', team_id: team.id, user_id: new_user.id }} }
```
carpodaster: Personal taste: I tend to make "generator spec helpers" real methods with `def foo(arg1)` instead of defining a proc/lambda in a `let`.

spec/models/team_spec.rb
```diff
@@ -147,6 +147,86 @@
...
+    let(:new_coach_as_student) { -> {{ name: 'student', team_id: team.id, user_id: new_user.id }} }
+    let(:second_new_coach) { -> {{ name: 'coach', team_id: team.id, user_id: create(:user).id }} }
```
carpodaster: Can't all these lambdas (expect the remove coach bit) be transformed to a single `let(:coach_attributes)` that returns a Hash?

spec/models/team_spec.rb
```diff
@@ -147,6 +147,86 @@
...
+
+      it 'does not allow the same coach to add themselves as a student' do
+        roles_attributes = [new_coach.call, new_coach_as_student.call]
+        team.attributes = { roles_attributes: roles_attributes }
+        expect { team.save }.not_to change { team.members.count }
+        expect(team.errors[:base].first).to eql "#{new_user.name} can't have more than one role in this team!"
+      end
+    end
+    context 'when team has coaches' do
+      let!(:coach_1) { create :coach_role, team: team, user: create(:user) }
```
carpodaster: You can probably use FactoryBot's  `create_list` for that

spec/models/team_spec.rb
ramonh: Nice! In your opinion, where would these be defined? The same spec file?

spec/models/team_spec.rb
ramonh: The only problem I've encountered here is that this would apply the same instance of `User` to all attributes. I'll look into a way to solve this.

app/models/team.rb
F3PiX: I totally agree.

app/models/team.rb
ramonh: Done! Thanks for the suggestion

spec/models/team_spec.rb
ramonh: Instead, using `4.times.map { create :coach_role, team: team, user: create(:user) }` does the trick. What do you think?

spec/models/team_spec.rb
```diff
@@ -147,6 +147,80 @@
...
+  describe 'limit of coaches' do
+    let(:team) { create :team }
+    let(:new_user) { create :user }
+    let(:coach_attributes) { [{ name: 'coach', team_id: team.id, user_id: new_user.id }] + 3.times.map { { name: 'coach', team_id: team.id, user_id: create(:user).id } } }
+    let(:new_coach_as_student) { { name: 'student', team_id: team.id, user_id: new_user.id } }
+    let(:fifth_new_coach) { { name: 'coach', team_id: team.id, user_id: create(:user).id } }
+    let(:remove_coach) { { name: 'coach', team_id: team.id, user_id: create(:user).id, _destroy: true } }
+
+    context 'when team has no coaches yet' do
+      it 'allows to add 2 new coaches' do
```
carpodaster: the description says 2, but it's about 4 in the example :)

spec/models/team_spec.rb
```diff
@@ -147,6 +147,80 @@
     end
   end
 
+  describe 'limit of coaches' do
+    let(:team) { create :team }
+    let(:new_user) { create :user }
+    let(:coach_attributes) { [{ name: 'coach', team_id: team.id, user_id: new_user.id }] + 3.times.map { { name: 'coach', team_id: team.id, user_id: create(:user).id } } }
```
carpodaster: This line is a bit long now. Can you make it a multi-line `do … end` let? Also see below for a suggestion on the 3.times.map(&block)

spec/models/team_spec.rb
```diff
@@ -147,6 +147,80 @@
...
+
+      it 'does not allow the same coach to add themselves as a student' do
+        roles_attributes = [coach_attributes.first, new_coach_as_student]
+        team.attributes = { roles_attributes: roles_attributes }
+        expect { team.save }.not_to change { team.members.count }
+        expect(team.errors[:base].first).to eql "#{new_user.name} can't have more than one role in this team!"
+      end
+    end
+    context 'when team has coaches' do
+      let!(:existing_coaches) { 4.times.map { create :coach_role, team: team, user: create(:user) } }
```
carpodaster: The `4.times.map` works, of course, but it looks a bit hacky to me. What about 

```ruby
let!(:existing_coaches) { create_list :coach_role, 4, team: team }  # `user` is implied in the factory
```

ramonh: This worked great, thank you!

spec/models/team_spec.rb
ramonh: Done!

spec/models/team_spec.rb
ramonh: Thanks for the suggestion! I've split it up, but can't use create_list for a set of attributes, can I?

carpodaster: No, I don't think you can

ramonh:
Cool! Thank you for the feedback <3

PR #916, opened 2018-01-26
Fix/redirect after deleting user

### LINKED ISSUE #915: deleting a user redirects to a non-existing page
When you delete a user (either as an organiser or, I'm guessing, as a user trying to delete your own profile, if that's a possible use case), you get redirected to `/users` which does not exist anymore (i.e. visitor gets a 404). The visitor should probably get redirected to `/community` instead, which is the “new” user list page.

carpodaster:
Fixes #915 

* Fixes the redirect target
* Improves the naming of the named route for `/community` (i.e. `community_path` now)
* Logs out the current user if they just deleted themselves.

### DIFF
--- a/spec/controllers/users_controller_spec.rb
+++ b/spec/controllers/users_controller_spec.rb
@@ -210,7 +210,7 @@ RSpec.describe UsersController, type: :controller do
 
     it 'redirects to user#show' do
       post :impersonate, params: { id: impersonated_user.id }
-      expect(response).to redirect_to community_index_path
+      expect(response).to redirect_to community_path
       expect(flash[:notice]).to include "Now impersonating #{impersonated_user.name}"
     end
   end
@@ -230,7 +230,7 @@ RSpec.describe UsersController, type: :controller do
 
     it 'redirects to community#index' do
       post :stop_impersonating
-      expect(response).to redirect_to community_index_path
+      expect(response).to redirect_to community_path
       expect(flash[:notice]).to include "Impersonation stopped"
     end
   end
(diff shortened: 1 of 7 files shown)


spec/controllers/users_controller_spec.rb
```diff
@@ -165,20 +165,15 @@
...
-    end
+    before { sign_in user }
 
     context "its own profile" do
-      it "destroys the requested user" do
-        expect {
-          delete :destroy, params: { id: user.to_param }
-        }.to change(User, :count).by(-1)
-      end
+      it "destroys the requested user and redirects to /community" do
```
carpodaster: I know this violates the "one assertion per example" rule. But that's how I roll in controller/request/feature/functional tests because the setup of DB objects is too expensive to not reuse the existing setup.

F3PiX: I am totally convinced ;-) 
In our first Dev Sync we talked about moving from controller tests to request tests. Would that help in this case?

carpodaster: Not as long the "one assertion per example" mantra is still being held up; the test would look almost the same, only the test macros change a little. I did not rewrite this as a request spec (even though we decided to gradually move request specs in the first DevSync as you mentioned) since we already had an exisiting test and I think the move to request specs should be one big PR on its own.

carpodaster:
@F3PiX I don't know much about Devise's magic tricks. What do you have in mind?

ramonh:
Thanks @carpodaster ! Looks good to me, too!

Issue #765, opened 2017-06-23
[feature request] send copy of application to applicants after submission


alicetragedy:
When applicants submit their application, they don't get a confirmation that the application has gone through. Furthermore, after the application deadline, applicants don't have access to their application anymore. 

I think a solid way to solve both problems at once would be to send a “copy” of the application data to both applicants. This would allow them to have a basic confirmation of the application going through, as well as a copy of their answers for future reference. This will be especially helpful before the summer, when applicants might want to review the features or issues they wanted to work on and adapt them to current needs of the project.

Any thoughts?

carpodaster:
👍 very good!

mkalininait:
I think it's a good idea. Additionally, it would be cool to show applicants their application in the Teams App (even after the applications deadline) with the status "Submitted".

klappradla:
Up next: show a link to the submitted application.

PR #917, opened 2018-01-28
[#765] [feature request] send copy of application to applicants after submission

klappradla:
This is supposed to fix https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/765: sending an email with the contents of the just submitted application to the two students of a team.

***edit:** this is still missing the logic to show the submitted application within the app as well.*

I added a multipart email that looks aprox. like this when rendered as `html`: 💿 
screen shot 2018-01-28 at 17 42 14

And approx. like this when rendered as `text`: 📵 
screen shot 2018-01-28 at 19 38 00

**Sidenote**
I know there should be something like `formats: [:html]` available for rendering partials from Rails views, but I somehow always (at least in the mailer tests) ended up with a mashup of both emails when naming the partials identically. So I for now just appended a `_text` to the text ones... May find a better solution to this at some point.

I also had to add some translations to the application draft form - otherwise it would basically be impossible to keep the form and the mail templates in sync 🙈 
While coming across it, I also removed a legacy intializer for `suckerpunch`. 

I came across some places where one could refactor a bit of "cruft", but I'll do this in follow up PRs, am very tight on time at the moment 😿


klappradla:
Oh, just had another look at the issue: totally forgot about Maria's comment, showing the filled out application to students even **after** they submitted and the summer started.

I guess we could show them the disabled form, similar to when people already submitted... I'm just not 100% sure about the navigation and logic for that. Will try to come up with something.

app/views/layouts/application_form_mailer.slim
```diff
@@ -0,0 +1,63 @@
...
+            table[width='640' cellpadding='0' cellspacing='0' border='0' bgcolor='#d82222']
+              tr
+                td[height='20' style='font-size:10px; line-height:10px;']
+                  |
+              tr
+                td[align='center' valign='top']
+                  table[width='600' cellpadding='0' cellspacing='0' border='0']
+                    tr
+                      td[align='left' valign='top']
+                        = image_tag('logo.png', width: 150)
```
klappradla: I cannot remember the last time I dealt with images in mails send *directly* from a Rails app... I'm not 100% sure about the use of the `image_tag` helper there. Is it a save thing to do? Or do I need to do a "normal" `` tag and then point the source to the asset **url**?

ramonh: I think this is fine, but don't have a lot of experience in using images in emails

ramonh:
Thank you very much @klappradla! I've had a look at the changes and they look great! 👍 

Do you reckon it makes sense to split the disabled application form viewing into a different PR?

app/views/application_form_mailer/_project_info.slim
```diff
@@ -0,0 +1,26 @@
...
+h3 About your project
+
+b = Application.data_label('project1_id')
+p = link_to_if project1, project1&.name, project1
+
+b = Application.data_label('why_selected_project1')
+= simple_format(auto_link(show_or_na(data.why_selected_project1)))
+
+b = Application.data_label('plan_project1')
+= simple_format(show_or_na(render_markdown(data.plan_project1)))
```
carpodaster: I feel at some point we should introduce a presenter 🙃   Let's not bother us with it in this PR, though

app/models/team.rb
```diff
@@ -201,12 +205,4 @@ def members_with_role(role)
   def two_students_present?
     students.reload.select(&:persisted?).size == 2
   end
-
-  # def must_have_members
```
carpodaster: 🎊

app/controllers/application_drafts_controller.rb
```diff
@@ -63,7 +63,7 @@ def apply
     if current_team.coaches_confirmed?
       if application_draft.ready? && application_draft.submit_application!
         flash[:notice] = 'Your application has been submitted!'
-        ApplicationFormMailer.new_application(application_draft.application).deliver_later
+        application_draft.application.notify_orga_and_submitters
```
carpodaster: (ugh, for some reason that comment didn't make it into my review – PEBCAC for sure)

Personal taste: I tend to `!` methods with IO/network side effects. Then again, I also like it that the method doesn't say that the type of notification is an email. /shrug :) 

One other thing, though: can we move this notification into the state transition? The method call on `application` looked a bit suspicious (Mrs Demeter called) and that got me thinking that it's probably not the controller's responsibility

klappradla: Haha, nice that we're on the same page here @carpodaster !

I actually had both, the `ApplicationDraft` model's state machine and the `CreateApplicationEtc...` service refactored to do exactly this 😹 

But I had to adjust too many tests and did not have the time to finish this on Sunday. So I'll do it in another PR.

Btw., a thing I changed there: I'd prefer the `CreateApplicationEtc.` service to actually **return** the persisted application record to then use this instance for notifying orga and the submitters. Using `deliver_later` for the emails requires all DB transactions for the `draft` and the `application` to be completed - assigning the `application` makes this a bit more explicit to me ✌️

klappradla: So, are you fine with "fixing" this in a follow up @carpodaster ?

klappradla:
For the disabled application view @ramonh: yes I think it would be best to do this in a separate PR. We just have to repopen the associated issue again for this.
I can start doing this probably even today... just not sure yet where to best place the link and how long to best have it being available.

I guess it should disappear once the season changes to the next year right (cc @carpodaster )

ramonh:
Good point, I would suggest linking it to the season change, yeah

alicetragedy:
❤️

PR #921, opened 2018-01-29
Show application as submitted

klappradla:
This is an attempt on part 2 of #765 : show the application as "submitted" to the students after it was submitted.

I for now only updated the application draft form to show a different message once an application was submitted and is therefore also no longer editable. Looks like this in the wild:
screen shot 2018-01-29 at 22 21 56

This 👆 is probably not sufficient though. I was thinking of also giving the users access to this disabled form during the time from when applications close until eventually the end of the season... Is this something we want (cc @ramonh & @alicetragedy) ?
In order to do this I have to do some "rethinking" on the navigation items 😉  I'm already putting some though in there, but would be super happy about suggestions 🤗


F3PiX:
So cool! About the navigation:  we do have a link to the appl form during application phase, pointing to the draft. Would you like one of these?  a) change the existing link after submitting, to point to the disabled form? Or b) add an extra (greyed out) link that will stay active as long as we chose to show the form.  
Either way, the navigation is not in the form itself. Which, to me, makes sense.

klappradla:
Sort of b) @F3PiX 

I was thinking about adding link to something like `application#show` only showing it to users with a submitted draft and only from the end of the application phase to maybe the end of summer.
There, I can just take the `application_draft` from the application and render the disabled form again... What do you think?

F3PiX:
That sounds totally logical 👍 .  One thing though: I guess the times I re-read the submitted application the most, was between submitting and the application phase was closed. So I'd vote for making the 'copy' accessible right from after the user submitted the final version.

klappradla:
That is already possible @F3PiX - the `my application` link is available until the end of the application phase.

alicetragedy:
@klappradla sorry, I somehow didn't see this notification. Yes, I think giving them access to the disabled form would be good, and it would make sense if it was until end of the season, that way they can look over the application and their project plan. 
Sidenote: I'm not sure if I missed the conversation about this — was there a plan to send them a copy of their application by email which was then dropped?

klappradla:
It wasn't dropped, I already added the email-part @alicetragedy, some screenshots here 👉 https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/917

alicetragedy:
@klappradla oh my. completely missed that, I'm sorry. thank you! Both look great 😍 
(and tbh I think that if the applicants already have that information, showing the disabled form is not _necessary_ if you're short on time). Totally grateful for all the work you're doing 🙌

Issue #920, opened 2018-01-29
Remove volunteering checkbox from application form


ramonh:
As discussed, this will no longer be offered to applicants.

However, we should not remove the field from the database, for our records.

klappradla:
Why not also remove the field and all logic connected with it? If we're not using it, I don't see a point in having it around... 🤔

ramonh:
I thought of this too @klappradla but I think we might want to keep it around, say, if we want to display stats from previous years. Wdyt?

klappradla:
Valid point. But won't the contents of the DB and the current state of the application be exported by the end of the season anyways..?

If not, I'd suggest to approach it sorta like we did it when deprecating other application attributes last year: rename the column, prefix it with `deprecated_<original-name>` so that we know we can throw it out at some point.

However I'd still get rid of the logic inside the app related to this attribute - cause this looks like unused code to me then 🚜

ramonh:
That sounds good to me! Sorry I was unclear, I intend to remove the volunteer logic as well 👍 

So I'll rename the column to `deprecated_<name>`

F3PiX:
(Heads up: we use the `rgsoc-2018` label for issues that the student team can work on during Summer. Instead, I created a new Milestone for issues that need to be done before the Application Phase starts.)

ramonh:
Oh cool, thanks for putting up with this github newbie 😅

Issue #919, opened 2018-01-29
Integrate part-time scholarship options into application process


ramonh:
As discussed, this will take the shape of two checkboxes:

- Full-time (40 hours per week)
- Part-time (20 hours per week)

PR #923, opened 2018-01-31
Replace volunteer with part time in applications

ramonh:
Related issue #919 #920

This does the following:

- Deprecate the `voluntary` and `voluntary_hours` fields
- Add the `WorkWeek` model with two default values: part time and full time
- Add the WorkWeek values to the application copy sent to students


ramonh:
Oops, sorry for the build fail, folks, gotta fix these spacings 🙈

app/views/application_drafts/_form.slim
```diff
@@ -58,10 +58,9 @@
...
-      p You can read up on volunteer teams in the #{link_to 'application guide', 'http://railsgirlssummerofcode.org/students/application/#volunteer-teams', target: :blank}.
-      = f.input :voluntary, as: :boolean, hint: 'This will not exclude you from being in the running for a sponsored seat.', label: 'We could join the RGSoC as a volunteer team', input_html: { disabled: !application_draft.as_student? || application_draft.applied? }
-      = f.input :voluntary_hours_per_week, as: :integer, label: 'As a volunteer team, how many hours per week would you be able to work on the project?', hint: 'Please only enter a full number (e.g., 15).', input_html: { disabled: !application_draft.as_student? || application_draft.applied? }
+      = f.input :work_week_ids, collection: WorkWeek.all.map {|choice| [choice.human_readable_label, choice.id]}, as: :check_boxes, label: "How would you prefer to work? (Please pick at least one)", :input_html => {multiple: true }
+
+      = f.input :work_week_explanation, label: 'Please explain why you made the selection above', input_html: { rows: 5, disabled: !application_draft.as_student? || application_draft.applied? }
```
klappradla: Same here: the label should go into the translations file

app/views/organizers/users_info/index.html.slim
```diff
@@ -9,7 +9,7 @@ h1.header User info
         = role.capitalize
 
     h5 Type
-    - %w(all sponsored voluntary).each do |kind|
+    - %w(all sponsored deprecated_voluntary).each do |kind|
```
klappradla: I'd again suggest to remove the deprecated column here but instead maybe look for a way to include the new type of team "half-time" or however it should be called.

app/views/reviewers/applications/_application.html.slim
```diff
@@ -13,9 +13,9 @@ tr
   td = application.coaching_company
   td.text-right = application.application_data['student0_application_money']
   td.text-left = application.application_data['student1_application_money']
-  td.text-center = icon('check') if application.application_data['voluntary'].to_s == 'true'
-  td.text-center = application.application_data['voluntary_hours_per_week']
+  td.text-center = WorkWeek.where(id: JSON.parse(application.application_data['work_week_ids'])).map(&:human_readable_label) if application.application_data['work_weeks_ids']
```
klappradla: Didn't you add a method to the `application` to spare this `JSON` parse part?

app/views/reviewers/applications/_application_data.html.slim
```diff
@@ -147,13 +147,13 @@
   b = Application.data_label("working_together")
   = simple_format(auto_link(show_or_na(ad.working_together)))
 
-  h3 Volunteering
+  h3 Work Week
```
klappradla: This should probably be plural?

app/views/reviewers/applications/_application_data.html.slim
```diff
@@ -147,13 +147,13 @@
...
 
-  b = Application.data_label("voluntary")
-  p = auto_link(show_or_na(ad.voluntary))
+  b = Application.data_label("work_weeks")
+  p = auto_link(show_or_na(WorkWeek.where(id: ad.work_week_ids).map(&:human_readable_label)))
```
klappradla: I think we can spare the `auto_link` here.

db/migrate/20180129133617_create_work_schedules.rb
```diff
@@ -0,0 +1,19 @@
...
+    end
+
+    create_table :application_draft_work_schedules do |t|
+      t.belongs_to :application_draft, index: true
+      t.belongs_to :work_schedule, index: true
+      t.timestamps
+    end
+
+    timestamp = Time.now.strftime("%Y-%m-%d %H:%M:%S")
+    connection.execute("INSERT INTO work_schedules (label, human_readable_label, created_at, updated_at) VALUES ('full_time', 'Full-time (40 hours per week)', '#{timestamp}', '#{timestamp}');")
```
klappradla: Having this code inside the `change` method makes the migration no longer reversible....

You could in theory wrap it in something like this:
```ruby
reversible do |dir|
      dir.up do
        # code
```
_(written outta my head, not tested)_

spec/controllers/organizers/teams_controller_spec.rb
```diff
@@ -18,17 +18,17 @@
     include_context 'with admin logged in'
 
     describe 'GET index' do
```
klappradla: Please remove any tests and logic related to the deprecated columns and test for the new / current types of teams instead.

spec/controllers/teams_controller_spec.rb
```diff
@@ -60,9 +60,9 @@
     end
 
     context 'after acceptance letters have been sent' do
-      let(:last_season)      { Season.create name: Date.today.year - 1 }
-      let!(:voluntary_team)  { create :team, :in_current_season, kind: 'voluntary' }
-      let!(:sponsored_team)  { create :team, :in_current_season, kind: 'sponsored' }
+      let(:last_season) { Season.create name: Date.today.year - 1 }
+      let!(:deprecated_voluntary_team)  { create :team, :in_current_season, kind: 'deprecated_voluntary' }
```
klappradla: Same here: please remove anything related to the deprecated column and test for the new types of teams instead.

spec/factories/application.rb
```diff
@@ -5,6 +5,7 @@
       student0_application_coding_level: 2,
       student1_application_coding_level: 2,
       student_name: FFaker::Name.name,
+      work_week_ids: [create(:work_week).id],
```
klappradla: Let's keep the factories minimal: all that's needed to have a valid application. So let's just skip the `work_weeks` here, they're not mandatory.

spec/factories/application.rb
```diff
@@ -37,6 +38,7 @@
         'student1_application_code_samples': FFaker::Lorem.paragraph,
         'student1_application_learning_history': FFaker::Lorem.paragraph,
         'student1_application_skills': FFaker::Lorem.paragraph,
+        'work_week_ids': [create(:work_week).id],
```
klappradla: Same here: let's skip the non-mandatory things.

spec/factories/application_drafts.rb
```diff
@@ -26,9 +31,9 @@
       plan_project2         { FFaker::Lorem.paragraph }
     end
 
-    trait :voluntary do
-      voluntary true
-      voluntary_hours_per_week 20
+    trait :deprecated_voluntary do
```
klappradla: Please remove this.

spec/factories/work_weeks.rb
```diff
@@ -0,0 +1,11 @@
+FactoryBot.define do
+  factory :work_week do
+    label                   { FFaker::Lorem.sentence }
```
klappradla: I wonder if using `FFaker` makes sense here 🤔  The values we have are two set sets that only make sense in conjunction....

spec/models/application_draft_spec.rb
```diff
@@ -57,15 +57,16 @@ def build_draft
       it_behaves_like 'proxies :apply validation', :heard_about_it
       it_behaves_like 'proxies :apply validation', :working_together
       it_behaves_like 'proxies :apply validation', :why_selected_project1
+      it_behaves_like 'proxies :apply validation', :work_weeks
 
-      context 'required fields for voluntary mode' do
-        it { is_expected.not_to validate_presence_of :voluntary_hours_per_week }
-        it { is_expected.not_to validate_presence_of(:voluntary_hours_per_week).on(:apply) }
+      context 'required fields for deprecated_voluntary mode' do
```
klappradla: Please remove all the logic and tests related to the deprecated columns _(in the whole file)_

spec/models/source_spec.rb
```diff
@@ -5,7 +5,7 @@
   it { is_expected.to validate_presence_of(:url) }
 
   describe '.for_accepted_teams' do
-    let!(:accepted_team)   { create :team, kind: %w(sponsored voluntary).sample }
+    let!(:accepted_team)   { create :team, kind: %w(sponsored deprecated_voluntary).sample }
```
klappradla: Same here: please remove the deprecated columns and test for the new things instead.

spec/models/team_spec.rb
```diff
@@ -455,8 +455,8 @@
       expect(subject).not_to be_accepted
     end
 
-    it 'returns false for a voluntary team' do
-      subject.kind = 'voluntary'
+    it 'returns false for a deprecated_voluntary team' do
```
klappradla: All over the file: please remove anything related to the deprecated columns.

app/models/application_draft.rb
```diff
@@ -75,6 +75,8 @@ class ApplicationDraft < ApplicationRecord
   belongs_to :updater, class_name: 'User'
   belongs_to :project1, class_name: 'Project'
   belongs_to :project2, class_name: 'Project'
+  has_many   :application_draft_work_weeks
```
klappradla: Hm, I fear I don't get the benefits of this approach 🙈 

It appears rather complex / overly complex to me for what we want to archive: have working-weeks set to either "full time" or "half time".

An more lightweight approach would for instance be: add a new integer column `work_weeks`to the `application_draft` model, make it an [Rails enum](http://api.rubyonrails.org/classes/ActiveRecord/Enum.html) (defined with hash syntax at best) and add the respective translations for it and use a simple select in the form.

Or even simpler: make it a string column and use something like [this gem](https://github.com/makandra/assignable_values) to restrict the values and get the naming right.

ramonh: Hmm, another project where I was collaborated insisted on having this approach over others. Now I'm not sure which is best..

The definite advantage of having WorkWeeks be tied to a table is that we're not tied to hard-coded values and can later on add more options, say a 10 hour week (highly improbable)

ramonh:
Thank you @klappradla ! I'll get to these right away :)

app/models/application_draft.rb
klappradla: Yes, makes sense, I get this point!

I think enum defined with a hash will also save you in this case - so I'd consider it a similar flexible but a bit more lightweight option... What do you think?

app/models/team.rb
```diff
@@ -2,9 +2,9 @@
 class Team < ApplicationRecord
   include ProfilesHelper, HasSeason
 
-  delegate :sponsored?, :voluntary?, to: :kind
+  delegate :sponsored?, :deprecated_voluntary?, to: :kind
 
-  KINDS = %w(sponsored voluntary)
+  KINDS = %w(sponsored deprecated_voluntary)
```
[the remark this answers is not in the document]
ramonh: My aim is to have this done in a later PR, as this one is more related to the application form

app/models/application_draft.rb
ramonh: On the other hand, we have the `heard_about_it` field, which is a string array. I could go with this instead

ramonh:
Hey @klappradla, I've taken care of most of the feedback, thank you!

The only thing I couldn't figure out was localizing the field. I noticed that all the localized fields are in `app/views/application_drafts/_student_fields.html.slim`, but the ones in `app/views/application_drafts/_form.slim` are not. 

I'm afraid I'm not sure how to get these localized, could you please point me in the right direction?

ramonh:
Ahhh scratch that @klappradla I figured it out 🙈

Thanks again for the feedback!

db/migrate/20180129133617_create_work_schedules.rb
```diff
@@ -0,0 +1,24 @@
+class CreateWorkSchedules < ActiveRecord::Migration[5.1]
```
klappradla: Let's just fully remove these migrations then if we don't use these tables - makes it easier to run them again.

spec/factories/application.rb
```diff
@@ -5,6 +5,7 @@
       student0_application_coding_level: 2,
       student1_application_coding_level: 2,
       student_name: FFaker::Name.name,
+      work_weeks: FFaker::Lorem.sentence,
```
klappradla: I guess this should be an array now.

spec/factories/application_drafts.rb
```diff
@@ -8,11 +8,12 @@
       association :project1, :accepted, factory: :project
       plan_project1         { FFaker::Lorem.paragraph }
       heard_about_it        { [FFaker::Lorem.paragraph] }
+      work_weeks            { [FFaker::Lorem.paragraph] }
       why_selected_project1 { FFaker::Lorem.paragraph }
       working_together      { FFaker::Lorem.paragraph }
       misc_info             { FFaker::Lorem.paragraph } # NOTE not a required field
 
-      after(:create) do |draft|
+      after(:create) do |draft, evaluator|
```
klappradla: Hm, I don't get this change I'm afraid 🙈

ramonh: That is a leftover from older commits, thank you for finding it! 😱

db/migrate/20180129133617_create_work_schedules.rb
ramonh: Hm... Currently, I have a migration to replace these with the array. Should I throw them all out and start over in that regard?

klappradla: Yep. Just get the `schema.rb` from `master`, remove the migration files we don't need and adjust the others to just add the 2 columns and everything's nice and clean.

I can also quickly do it - no more need for more work for you. You did an awesome job already.

klappradla:
Solid 💪 

I'll merge once Travis ran 🏃‍♀️

Issue #926, opened 2018-02-05
Crash when user has no `current_team` upon applying


ramonh:
If a user has no `current_team` and they try to click on "Apply Now", the app will throw an error

klappradla:
Can we close this now since #927 has been merged?

ramonh:
Yes, thank you @klappradla!

PR #927, opened 2018-02-05
Fix crash when `current_team` is nil

ramonh:
Related issue #926 

- Students with teams in previous seasons will get a crash when applying with no current team


ramonh:
Thank you @carpodaster, done!

ramonh:
Yay! Thanks @carpodaster

Issue #930, opened 2018-02-09
Wording on code samples in application form needs some clarification


ramonh:
As it stands right now:

> E.g., links to completed coding challenge provided by the mentor (if present), commits or pull requests on Github or other concise examples of your code. Links to GitHub profiles and repositories won't be considered. Please do not directly copy-paste code here.

A question was brought up on Slack that made us decide tweaking the phrasing to:

> E.g., links to completed coding challenge provided by the mentor (if present), commits or pull requests on Github or other concise examples of your code (including links to single files in repositories). Links to your Github profile or to entire repositories won\'t be considered. Please do not directly copy-paste code here

PR #931, opened 2018-02-09
Application form wording tweak

ramonh:
Related issue #930 

This adds some clarification to the coding samples field in the application form.


app/views/application_drafts/_student_fields.html.slim
```diff
@@ -53,7 +53,7 @@
...
+  = s.input :application_code_samples, as: :character_limited, input_html: { rows: 4, disabled: !may_edit?(s.object) }, wrapper_html: { class: student_field_error_class(index, application_draft, :application_code_samples) }, hint: (student_field_error_hint(index, application_draft, :application_code_samples) || "E.g., links to completed coding challenge provided by the mentor (if present), commits or pull requests on Github or other concise examples of your code (including links to single files in repositories). Links to your Github profile or to entire repositories won't be considered. Please do not directly copy-paste code here")
```
alicetragedy: there is a missing `.` at the end of this sentence 😇 
Also, for consistency's sake, I would suggest using single quotes here as well and escaping the apostrophe in `won't`, like this: `won\'t`. What do you think?

ramonh: Oh sure! I had the habit of using doubles for when there are singles inside, but am not picky 👍

Issue #925, opened 2018-02-05
Differentiating between part-time and full-time teams


ramonh:
Up until recently, we were differentiating between `sponsored` and `volunteer` teams.

With the introduction of part-time and full-time, as well as replacing volunteer teams with these, we'll need a mechanism to differentiate between these.

For this, we'll need:

- [x] Integrate these changes into the reviewing process
- [x] Add these to the `kinds` for `Team`s
- [x] Show and filter these in the different `index` views

klappradla:
With #948 being merged, can this be considered done now @ramonh ?

ramonh:
Yes! Totally, thanks @klappradla

PR #936, opened 2018-02-12
Replace `sponsored` kind with full- and part-time

ramonh:
Related issue #925 

This is the first of a series of steps towards rolling out the new team kinds, replacing the team logic for sponsored and volunteering teams with full and part-time teams

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -118,7 +118,7 @@ module ApplicationHelper
 
   def links_to_user_teams(user)
     user.teams.map do |team|
-      link_to(team.name || team.project, team, class: "team #{team.sponsored? ? 'sponsored' : ''}")
+      link_to(team.name || team.project, team, class: "team #{team.accepted? ? 'sponsored' : ''}")
     end.map(&:html_safe)
   end
 
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -118,7 +118,7 @@ module ApplicationHelper
 
   def links_to_user_teams(user)
     user.teams.map do |team|
-      link_to(team.name || team.project, team, class: "team #{team.accepted? ? 'sponsored' : ''}")
+      link_to(team.name || team.project, team, class: "team #{team.accepted? ? 'full-time' : ''}")
     end.map(&:html_safe)
   end
 
--- a/app/models/team.rb
+++ b/app/models/team.rb
@@ -2,9 +2,10 @@
 class Team < ApplicationRecord
   include ProfilesHelper, HasSeason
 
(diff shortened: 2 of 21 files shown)


app/helpers/application_helper.rb
```diff
@@ -118,7 +118,7 @@ def team_group_emails(team)
 
   def links_to_user_teams(user)
     user.teams.map do |team|
-      link_to(team.name || team.project, team, class: "team #{team.sponsored? ? 'sponsored' : ''}")
+      link_to(team.name || team.project, team, class: "team #{team.accepted? ? 'sponsored' : ''}")
```
klappradla: Hm, did you not just rename the `sponsored` css class to `full-time`... 🤔 
I fear this then won't work anymore.

spec/controllers/organizers/teams_controller_spec.rb
```diff
@@ -18,7 +18,7 @@
     include_context 'with admin logged in'
 
     describe 'GET index' do
-      let!(:sponsored_team) { create :team, :in_current_season, kind: 'sponsored' }
+      let!(:sponsored_team) { create :team, :in_current_season, kind: 'full_time' }
```
klappradla: I'd suggest to also rename this to `full_time`throughout the tests, just to make the intend more obvious _(we probably now would consider both sort of teams "sponsored")_

spec/models/source_spec.rb
```diff
@@ -5,7 +5,7 @@
   it { is_expected.to validate_presence_of(:url) }
 
   describe '.for_accepted_teams' do
-    let!(:accepted_team)   { create :team, kind: %w(sponsored).sample }
+    let!(:accepted_team)   { create :team, kind: %w(full_time).sample }
```
klappradla: Can you please remove this whole thing around creating an array with a single element and then taking a "sample" from it? I think it's fine to just to with the string `full_time` here.

app/models/team.rb
```diff
@@ -2,9 +2,10 @@
 class Team < ApplicationRecord
   include ProfilesHelper, HasSeason
 
-  delegate :sponsored?, to: :kind
+  delegate :full_time?, to: :kind
```
klappradla: I'm sorry, but I think I don't understand what this should be 🙈

ramonh:
Thanks very much for the feedback, @klappradla ! I'll be cleaning these up.

app/models/team.rb
ramonh: These allow us to call `full_time?` or `part_time?` on the team models. Removing them would give me a no method error.

ramonh:
Done! Could you please take another look, @klappradla?

Issue #932, opened 2018-02-11
Teams decided to not listed in the team overview should have more privacy


ArushiSinghal:
If the team decides not to get listed in the team overview, then their team name should also not be shown in the team-member profile. This can help the members to remain more private until their selection.

ramonh:
Thanks for pointing this out, @ArushiSinghal! Will look into this 👍

klappradla:
Should be fixed via #938 now @ArushiSinghal - thanx again for reporting 💚

PR #938, opened 2018-02-15
Bugfix/hide team on profile if invisible

ramonh:
Related issue #932 

This aims to hide a user's team on their profile if invisible, unless they're accepted.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -86,10 +86,6 @@ class User < ApplicationRecord
     @github_import = import
   end
 
-  def visible_teams
-    teams.accepted.merge(teams.visible)
-  end
-
   class << self
     def ordered(order = nil, direction = 'asc')
       direction = direction == 'asc' ? 'ASC' : 'DESC'
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -86,6 +86,10 @@ class User < ApplicationRecord
     @github_import = import
   end
 
+  def visible_teams
+    teams.accepted.merge(teams.visible)
+  end
+
   class << self
     def ordered(order = nil, direction = 'asc')
       direction = direction == 'asc' ? 'ASC' : 'DESC'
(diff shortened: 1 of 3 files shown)


app/models/user.rb
```diff
@@ -86,6 +86,10 @@ def github_import=(import)
     @github_import = import
   end
 
+  def visible_teams
```
klappradla: Personal taste, but I think it's overkill to a) introduce a new method in the `user` model for this and b) merge the two queries which roughly do the same...

How about refactoring the `visible` class method on the `Team` model into a scope? It's technically speaking already this... Then you'll be able to simply chain the scopes into something like
```ruby
user.teams.accepted.visible
```

What do you think?

app/views/users/show.html.slim
```diff
@@ -42,10 +42,10 @@ nav.actions
     h4 About me
     p = render_markdown(@user.bio).html_safe
 
-  - if @user.teams.any?
+  - if @user.visible_teams.any?
```
klappradla: I'm still not 100% sure if I got the feature and the issue right. The way I understand it, teams set to "invisible" should be hidden from the profile until they are accepted. That's also how I read the current implementation of the `visible` method on the `team`:
```ruby
def visible
  where("teams.invisible IS NOT TRUE OR teams.kind IN (?)", KINDS)
end
```
So once a team is accepted, it's always visible. Which also means, the chaining of scopes you do in the user model should actually not even be necessary. It should be enough to do a
```ruby
@user.teams.visible.any?
```

spec/models/user_spec.rb
```diff
@@ -317,6 +317,25 @@
     end
   end
 
+  describe '#visible_teams' do
```
klappradla: As said in my previous comment: I'd prefer to make the `visible` class method a scope in the `team` model. So you could probably move the specs over there if it's not tested accordingly yet.

ramonh:
@klappradla You were absolutely right. This is what happens when you don't read the scope implementation 🙈 Thanks for the second pair of eyes! I've just pushed my fix. Could you please take a look?

klappradla:
😹 ok, wouldn't have expected it to be such a tiny change after all 🐭

ramonh:
Indeed! Thanks for your help!

Issue #934, opened 2018-02-11
Populate Team dropdown on the Activity page between seasons


ArushiSinghal:
On the Activity page, the "Team drop-down" is of no use.
![screenshot-2018-2-11 rgsoc - teams app](https://user-images.githubusercontent.com/20858804/36079324-d9f06926-0fa7-11e8-9d48-3245c435568d.png)

Edit: Summing up the discussion below. The requirements are:
1) During the `current season`, the dropdown is populated with the teams that are `accepted` in the current season
2) Until the `current season` starts, the dropdown is populated with the teams that were `accepted` in the previous season. 

Updated the issue title accordingly.

ramonh:
Hey @ArushiSinghal thanks for pointing this out!

The purpose of this dropdown is to display the accepted teams of the current Summer of Code season. 

So as you correctly found, the dropdown is empty between seasons. 

One thing we could do is hide the dropdown if there are no teams to display and show it otherwise.

What do you think?

ArushiSinghal:
Yes, we should hide the dropdown when there is no team. @ramonh  should I go ahead to solve this issue?

ramonh:
Sounds good, thanks @ArushiSinghal go for it!

carpodaster:
We could also fill the menu with the selected teams of the most recent season when we don't have selected teams for the current one, yet. Wdyt?

ramonh:
Good point, this is much better!

PR #940, opened 2018-02-21
Fixed redundant styling on the Activity Page of Teams App

ArushiSinghal:
Fixed Issue #934 
When the there is no team to be displayed then the presence of team drop-down is redundant. So drop down will only be shown when there are teams in the database, else it won't be shown.
If there is no team then:-

screenshot-2018-2-22 rgsoc - teams app 1
If it is:-
screenshot-2018-2-22 rgsoc - teams app


app/views/activities/index.html.slim
```diff
@@ -11,9 +11,9 @@ div.row
           label.radio
             = radio_button_tag :kind, kind == 'all' ? '' : kind, params[:kind] == kind
             = " " + kind == 'feed_entry' ? 'Blog Post' : kind.titleize
-        label.team_filter
-          ' Team:
-          = select_tag :team_id, options_for_select(teams.map { |t| [t.display_name, t.id] }, params[:team_id]), include_blank: true, class: 'form-control'
+        - if teams.count != 0
```
F3PiX: Because of readabilty, I'd vote for:
- `unless teams.count = 0` 
or
- `if teams.any?`

app/views/activities/index.html.slim
```diff
@@ -11,9 +11,9 @@ div.row
...
+          label.team_filter
+            = select_tag :team_id, options_for_select(teams.map { |t| [t.display_name, t.id] }, params[:team_id]), include_blank: 'Teams', class: 'form-control'
```
F3PiX: Nice, the string instead of a blank field!!

F3PiX:
O, and could you add a before and after screenshot to the [description](https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/940#issue-170542777) ?

app/views/activities/index.html.slim
ArushiSinghal: Yes I again make a pull request after changing it to 
```
if teams.any?
```
Thanks for the suggestion.

ArushiSinghal:
Thanks @F3PiX for the suggestion, I made the required changes.
Also, I did ``` git pull origin master ```, but I didn't make the new branch for doing this task.
If not acceptable I'll again make a pull request.

F3PiX:
Yes please, close this one and open a new Pull Request 👍

F3PiX:
I just realised that the issue is not solved with just checking if there 'are' teams. 
See this [comment:](https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/934#issuecomment-364842498)  The teams are present in the database. They are not in the dropdown, because we are 'between seasons' (= between two Summers of Code). 

And in a later comment, the proposed solution is to show the teams from the previous season, instead of hiding the button: https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/934#issuecomment-364868821

So, there are two situations:
1) During the current season, the dropdown is populated with the teams that are `accepted` `in current season`.
2) Before the current season starts, the dropdown is populated with the teams that were `accepted` in the previous season. 

Let me know if you need help!!

simransinghal:
Hi @F3PiX 
Arushi and I are working on this issue. We are stuck, that what is the structure of the database. We tried to figure it out and come to know that drop-down is showing only those teams which are part-time with season id as 1. Drop-down is not showing "sponsored" and "voluntary" teams even though their season id is 1.
So, we want to clarify is the season id which helps in finding the present and past years team.

F3PiX:
Aha. In previous years, teams were not 'part time' or 'full time' but 'sponsored' or 'volunteer'. We changed that recently. The weird thing :-) is: that doesn't matter for a team being [ `accepted` ](https://github.com/rails-girls-summer-of-code/rgsoc-teams/blob/master/app/models/team.rb#L48) or selected.

1) BUT I just discovered that the seeds were not updated when that ^ change was made! https://github.com/rails-girls-summer-of-code/rgsoc-teams/blob/master/db/seeds.rb#L1-L8
When you forked the repo, you got some database records, but not with the new 'kinds'. 
So, the first thing to do would be updating the seeds. That will make your life so much easier!
You can update the seeds, see the lines in my link, then do `rake db:reset` (to get rid of the existing records) and then `rake db:seed` (to load the new seed data).

Hope that helps!! 

After that, all 'accepted' or 'selected' teams of a season should show up in the dropdown, according to this line:  
https://github.com/rails-girls-summer-of-code/rgsoc-teams/blob/master/app/controllers/activities_controller.rb#L30 

2) Then to the season. You are close. We don't use the season.id though. In the teams model, you''ll find the [`in_current_season` scope for teams](https://github.com/rails-girls-summer-of-code/rgsoc-teams/blob/master/app/models/team.rb#L80). That should get you started.

The dropdown should show the 'accepted' or 'selected' teams only, for any given season only. I'd think the `teams` helper method (see link above) takes care of excluding the not-accepted teams. So yes,  for the dropdown logic you can focus on the team's season. 
 
Hope that this helps! Happy coding!

ArushiSinghal:
Thanks @F3PiX for giving us the detailed explanation. It really helped us to get more familiar with the code and also we learnt a lot when trying to fix the issue. We made the required changes and the new pull request #941  is created by @simransinghal.

Issue #942, opened 2018-02-25
Display teams after filtering the year from the drop down menu


dimi-23:
Currently when a particular year is selected from the drop-down menu as shown in the figure below, no search results are obtained.

![image](https://user-images.githubusercontent.com/28829807/36640173-f93989d2-1a3f-11e8-9a03-53960f386938.png)
This issue needs to be rectified.

dimi-23:
I would like to work on this issue if it seems important.

klappradla:
Hey @dimi-23 👋 

Thanx a lot for reporting 🙏 

The problem originates from changes to the possible `kinds` of teams we recently introduced before this upcoming season (cc @ramonh ). In short: the old data of course does not adhere to the current datamodel 😉 

The "fix" would be to simply run a migration on the production database. Therefore I'm not really sure if this issue is a good fit for you to work on 🤔 It's a) a bit boring and b) required access to several internal resources (e.g. the production database) which we - as you probably understand - cannot just give to you 😉

ramonh:
Looks like this is working again by the look of it, is that right?

@klappradla, @dimi-23 can we close this?

PR #946, opened 2018-02-25
Migrate old team "kinds" to new datamodel

klappradla:
As #942 reports, displaying teams based on the season currently does not really work, as the old records don't adhere to the new datamodel for `team` - we have different **kinds** now: `part_time` and `full_time`.

I checked with the production db and added a small migration to move `sponsored` to `full_time` and `voluntary` to `part_time`. Even though that's not a 💯 fit, I'd say it's a better option then redefining what an `accepted` team is, or not showing any past teams at all... What do you think @ramonh ?

### DIFF
--- a/app/models/team.rb
+++ b/app/models/team.rb
@@ -45,9 +45,9 @@ class Team < ApplicationRecord
     where.not(id: Activity.where(kind: ['status_update', 'feed_entry']).where("created_at > ?", 26.hours.ago).pluck(:team_id))
   }
 
-  scope :accepted, -> { where(kind: %w(full_time part_time)) }
-  scope :full_time, -> { where(kind: 'full_time') }
-  scope :part_time, -> { where(kind: 'part_time') }
+  scope :full_time, -> { where(kind: %w(full_time sponsored)) }
+  scope :part_time, -> { where(kind: %w(part_time voluntary)) }
+  scope :accepted, -> { full_time.or(part_time) }
 
   scope :by_season, ->(year_or_season) do
     case year_or_season
--- a/app/models/team.rb
+++ b/app/models/team.rb
@@ -48,6 +48,8 @@ class Team < ApplicationRecord
   scope :full_time, -> { where(kind: %w(full_time sponsored)) }
   scope :part_time, -> { where(kind: %w(part_time voluntary)) }
   scope :accepted, -> { full_time.or(part_time) }
+  scope :visible, -> { where.not(invisible: true).or(accepted) }
+  scope :in_current_season, -> { where(season: Season.current) }
 
   scope :by_season, ->(year_or_season) do
     case year_or_season
@@ -65,10 +67,6 @@ class Team < ApplicationRecord
       order([sort[:order] || 'kind, name', sort[:direction] || 'asc'].join(' '))
     end
 
-    def visible
-      where("teams.invisible IS NOT TRUE OR teams.kind IN (?)", KINDS)
-    end
-
     # Returns a base scope reflecting the relevant teams depending on what
     # phase of the running season we're currently in.
     def by_season_phase
@@ -77,10 +75,6 @@ class Team < ApplicationRecord
       else
         Team.in_current_season.visible
       end
-    end
-
-    def in_current_season
-      where(season: Season.current)
     end
   end
 
--- a/app/models/team.rb
+++ b/app/models/team.rb
@@ -16,23 +16,23 @@ class Team < ApplicationRecord
 
   attr_accessor :checked
 
+  belongs_to :project
+
+  has_one :last_activity, -> { order('id DESC') }, class_name: 'Activity'
+  has_one :conference_preference, dependent: :destroy
+
   has_many :applications, dependent: :nullify, inverse_of: :team
   has_many :application_drafts, dependent: :nullify
   has_many :roles, dependent: :destroy, inverse_of: :team
-  has_many :members, class_name: 'User', through: :roles, source: :user
+  has_many :members, through: :roles, source: :user
   Role::ROLES.each do |role|
-    has_many role.pluralize.to_sym, -> { where(roles: { name: role }) }, class_name: 'User', through: :roles, source: :user
+    has_many role.pluralize.to_sym, -> { where(roles: { name: role }) }, through: :roles, source: :user
   end
   has_many :sources, dependent: :destroy
   has_many :activities, dependent: :destroy
-  has_one :last_activity, -> { order('id DESC') }, class_name: 'Activity'
   has_many :comments, as: :commentable
   has_many :status_updates, -> { where(kind: 'status_update') }, class_name: 'Activity'
   has_many :conference_attendances, dependent: :destroy
-
-  has_one :conference_preference, dependent: :destroy
-  has_many :conferences, through: :conference_preference
-  belongs_to :project
 
   accepts_nested_attributes_for :conference_preference, allow_destroy: true
   accepts_nested_attributes_for :roles, :sources, allow_destroy: true
--- a/app/models/team.rb
+++ b/app/models/team.rb
@@ -73,7 +73,7 @@ class Team < ApplicationRecord
     # phase of the running season we're currently in.
     def by_season_phase
       if Time.now.utc > Season.current.acceptance_notification_at
-        Team.in_current_season.selected
+        Team.in_current_season.accepted
       else
         Team.in_current_season.visible
       end
@@ -81,10 +81,6 @@ class Team < ApplicationRecord
 
     def in_current_season
       where(season: Season.current)
-    end
-
-    def selected
-      where(kind: %w(full_time part_time))
     end
   end
 
--- a/app/models/team.rb
+++ b/app/models/team.rb
@@ -8,7 +8,6 @@ class Team < ApplicationRecord
   KINDS = %w(full_time part_time)
 
   validates :name, presence: true, uniqueness: true
-  # validate :must_have_members
   validate :disallow_multiple_student_roles
   validate :disallow_duplicate_members
   validate :limit_number_of_students
--- a/app/models/team.rb
+++ b/app/models/team.rb
@@ -49,17 +49,7 @@ class Team < ApplicationRecord
   scope :accepted, -> { full_time.or(part_time) }
   scope :visible, -> { where.not(invisible: true).or(accepted) }
   scope :in_current_season, -> { where(season: Season.current) }
-
-  scope :by_season, ->(year_or_season) do
-    case year_or_season
-    when Integer, String
-      joins(:season).where('seasons.name': year_or_season)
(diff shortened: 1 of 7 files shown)


ramonh:
Thanks for this @klappradla ! 

I'm a bit torn about this, because we'd be losing past information, that is, who was sponsored and who was volunteering. Maybe we'd want to use this in the future?

What do you think? Also pinging @alicetragedy if you don't mind :)

db/migrate/20180225173740_migrate_deprecated_kinds_teams.rb
```diff
@@ -0,0 +1,33 @@
+class MigrateDeprecatedKindsTeams < ActiveRecord::Migration[5.1]
+  class Team < ApplicationRecord
+    self.table_name = 'teams'
```
ramonh: Minor note here: I've understood that it's better to use `connection.execute` with an SQL statement instead of using the `ActiveRecord` functionality. Is this overkill?

klappradla: Hm, since I redefine the class with its table name, using AR is generally "save" in this migration.

But you're of course right, I'd also prefer using the raw connection. I just thought it may be harder to read and less in the line of the existing migrations... But I'll see how it looks like 👍

klappradla:
Ok, I updated the migration to use `execute` directly on the DB.

For your concern about past data @ramonh: I was also asking myself this... 🤔 

My conclusion was: doing statistics about past seasons vs. having a working Teams app are two different pairs of shoes. The app being a "app", the datamodel is of course constantly changing from season to season. By implementing / offering pages like the `teams` where users would be able to browse teams from previous seasons, I feel like we bought in to also caring about past seasons' data needing to be usable with the current version of the app, rather then sleeping in some corner of the database forever ⚰️ 

Similar to removing deprecated columns, in my opinion, for the sake of maintainability anything that is old and no longer used should be removed from the app. Therefore, I went in favor of losing these bits of information. Afaik (correcty me if I'm wrong @alicetragedy ) we have DB dumps from each season that can be used for doing "statistics" on, which in my opinion is also better scalable then trying to keep the db schema backwards compatible forever.

ramonh:
Oh! @klappradla if that's the case then I'm all for it. Well put, thank you! 🙌

carpodaster:
I generally agree to the notion that the app is moving and we cannot guarantee backwards compatibility forever. It's the price we pay for several other gains (like not having to set up / fork the app every year, re-using users/organizers, general visibility of past teams, …).

Fwiw. can we not _add_ the new `kind` values and map the past values to what we define here as their present day counterpart in the AR scope on team? We'd have to make sure that we don't do `where(kind: 'value')` (which we do, I think :thinking: ) and go through the proper AR scope instead.

klappradla:
Also a nifty solution @carpodaster 👍 

I'd be up for doing it.

Do you have any personal preferences for an approach on this? Also @ramonh ? All for not-migrating but updating the app logic?

alicetragedy:
Quickly leaving my thoughts in here: I think it would be a good idea to think about doing regular data dumps (e.g. maybe at the beginning and end of each season?) anyway 😇 and generally about tracking the kind of information we need better than we currently do.
Maybe trying to avoid too much backwards compatibility can help with that :)
I just checked though, and we've got the list of sponsored + volunteer teams in our internal wiki (except for 2013 and 2014). Looking for this information in a current dump and saving it in the same way as we did after 2014 would take care of the issue raised by @ramonh! (and if the information is unavailable in there for whatever reason — iirc there might not have been a teams app in 2013 — then there's other ways to figure this out.

ramonh:
Thanks for your thoughts everyone! 

So if I understood correctly, we wouldn't actually change the kinds, but have them be linked somehow, correct?

Also, what's the best way to do a db dump? And to store them?

klappradla:
Ok, should I update this PR accordingly to do the mapping tonight @ramonh ?

For the dumps: Heroku's Postgres plugin / extension or however they call it does pg dumps already. I'm not sure for how long they're stored thought... For storing a yearly dump on our side, what about just an additional private `archive` repo? (cc @alicetragedy )

ramonh:
That'd be super cool, thanks @klappradla

klappradla:
Ok @ramonh & @carpodaster I updated the `full_time` and `part_time` scopes on the `Team` model to also include the legacy terms *(`voluntary` and `sponsored`)*. The `accepted` scope is then based on these two and all further occurences of explicitly using the contents of the `KINDS` constant.

**TL;DR** I implemented what I think @carpodaster suggested 😉 

There is one leftover though: the `recipient` model still make use of `Team::KINDS`. I did not really understand the code there and the purpose why the `KINDS`  are actually used 😹 
It for now does not matter since this particular piece of code operates on current data anyways. I'll have a follow up PR where I update this.

Aside ☝️ I also included some small drive-by-fixes in the Team model, e.g. removing some no longer working associations and the duplicated `selected` scope. It it's too many changes, try looking at the individual commit messages.

app/models/team.rb
```diff
@@ -45,9 +45,9 @@ class Team < ApplicationRecord
     where.not(id: Activity.where(kind: ['status_update', 'feed_entry']).where("created_at > ?", 26.hours.ago).pluck(:team_id))
   }
 
-  scope :accepted, -> { where(kind: %w(full_time part_time)) }
-  scope :full_time, -> { where(kind: 'full_time') }
-  scope :part_time, -> { where(kind: 'part_time') }
+  scope :full_time, -> { where(kind: %w(full_time sponsored)) }
+  scope :part_time, -> { where(kind: %w(part_time voluntary)) }
+  scope :accepted, -> { full_time.or(part_time) }
```
ramonh: This is super cool, did not know!

app/models/team.rb
```diff
@@ -16,24 +16,24 @@ class Team < ApplicationRecord
...
   end
   has_many :sources, dependent: :destroy
   has_many :activities, dependent: :destroy
-  has_one :last_activity, -> { order('id DESC') }, class_name: 'Activity'
   has_many :comments, as: :commentable
   has_many :status_updates, -> { where(kind: 'status_update') }, class_name: 'Activity'
   has_many :conference_attendances, dependent: :destroy
 
-  has_one :conference_preference, dependent: :destroy
-  has_many :conferences, through: :conference_preference
```
F3PiX: Double check: The  `has_many :conferences` is gone? We don't need that?

app/models/team.rb
```diff
@@ -45,9 +45,11 @@ class Team < ApplicationRecord
     where.not(id: Activity.where(kind: ['status_update', 'feed_entry']).where("created_at > ?", 26.hours.ago).pluck(:team_id))
   }
 
-  scope :accepted, -> { where(kind: %w(full_time part_time)) }
-  scope :full_time, -> { where(kind: 'full_time') }
-  scope :part_time, -> { where(kind: 'part_time') }
+  scope :full_time, -> { where(kind: %w(full_time sponsored)) }
+  scope :part_time, -> { where(kind: %w(part_time voluntary)) }
```
F3PiX: I like this solution 🙏

app/models/team.rb
```diff
@@ -16,24 +16,24 @@ class Team < ApplicationRecord
```
F3PiX: Now that you are at it: can we get rid of line 11 as well?

F3PiX:
O sorry!!! I was late to the party. So usefull to review merged PR's 😂

klappradla:
No worries @F3PiX 💚 
This PR is not yet merged, just another one referencing this 😉 And I'm of course always super happy about comments 🤗 _(especially since all the coding I do here is definitely **after** the cognitive heights of my day)_

app/models/team.rb
klappradla: ✅ no, we don't need it any more.

The `conference_preference` has a `first_conference` and a `second_conference`, so the `conferences` associations as it is defined here is no longer available on the datamodel.

app/models/team.rb
klappradla: Oh yes, I probably can.

I remember I wanted to search for what this could have been once - but then just forgot about it 🙈 

I'll clean this up.

klappradla:
Thank you so much for your comments @F3PiX 🙏 

Not sure if I'll find the time tonight, but probably until Thursday I can do some updates 🚢 

Very good point with the seeds! I actually forgot about them 🙈 I'll updated.
However, a more general thing about *the seeds*: I think they are in the conceptually wrong place and should instead go into a `db:populate` task. I'd update this as a follow up to this already sorta large PR - unless people disagree with this.
Background: `seeds` at least to me is the minimal data required for the app to actually run. So maybe something like a `season` because everything else depends on an initial season to be there (just as an example). The seeds are automatically run if one does a `db:setup`, also on production.
These things ☝️ make it the wrong place for "fake" data for development where one would frequently drop the database or even run the seeds multiple times.
I can get into more details about this, but that's the rough tl;dr reasoning behind this 😉 

For restructuring models: I in general agree that it would be a useful things to do. **However**, I personally think it is sort of pointless if one does not combine it with cleaning up the specs a bit, removing non-used methods and things like that. Therefore an in my opinion better approach is to do something like
> clean up one model at a time (including tests)

Not sure if ☝️ is good fit for newcomes to the codebase though 🤔

F3PiX:
@klappradla 

> (especially since all the coding I do here is definitely after the cognitive heights of my day) 

😉Yeah, and me reading the merge messages after the cognitive heights of my day....

I totally agree with you on the concept of a db:populate task vs seeds. Would you mind if I take that upon me? Or would you prefer to do it yourself?

Great point also about cleaning up the models, but yeah, that makes it unsuitable as beginner_friendly tasks.

klappradla:
😹 

And sure, go ahead with the `db:populate` thing @F3PiX 👍

klappradla:
Ok, rebased and removed the commented code in the `Team` model ✅

F3PiX:
Ace!

klappradla:
Rebased again and slimmed down the slightly overkill `by_season` scope _(at least for our current use case)_.

klappradla:
Does anyone want to do another review or should I merge this?

klappradla:
Well, I'll just merge then. It's too much effort to rebase all the time 😉

app/models/team.rb
```diff
@@ -45,47 +44,27 @@ class Team < ApplicationRecord
...
-      joins(:season).where('seasons.name': year_or_season)
-    when Season
-      where(season: year_or_season)
-    else
-      none
-    end
-  end
+  scope :full_time, -> { where(kind: %w(full_time sponsored)) }
+  scope :part_time, -> { where(kind: %w(part_time voluntary)) }
+  scope :accepted, -> { full_time.or(part_time) }
```
carpodaster: Nice! Didn't know about `or` (seems logical, just never thought about it)

PR #947, opened 2018-02-25
Update dependencies

klappradla:
Some minor updates to dependencies (only the ones without breaking changes affecting us), including Rails.


klappradla:
Oh and just for your information: I also started updating `font-awesome`, but it's touching way too many files, so I'll add it in a separate PR 😹

klappradla:
**update:** the efforts of updating `font-awesome` are super huge, especially since we also use an additional soft-deprecated way of using icons as pseudo elements... It would make more sense to to this update combined with a refactoring of legacy css parts...

I'll keep the branch around, but for now I don't think it makes much sense.

PR #948, opened 2018-02-26
Filter full/part time teams

ramonh:
Related issue #925

The orga dashboard of the teams app will now show full and part time teams and allow filtering thereof.

There was also a bug where the star icon wasn't being correctly shown.

screen shot 2018-02-26 at 09 52 49

### DIFF
--- a/app/controllers/organizers/teams_controller.rb
+++ b/app/controllers/organizers/teams_controller.rb
@@ -17,8 +17,6 @@ module Organizers
         @teams = Team.in_current_season.full_time.ordered
       elsif params[:filter] == 'part_time'
         @teams = Team.in_current_season.part_time.ordered
-      elsif params[:filter] != 'all'
-        @teams = Team.in_current_season.selected.ordered
       end
     end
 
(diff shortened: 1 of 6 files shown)


app/controllers/organizers/teams_controller.rb
```diff
@@ -12,7 +12,12 @@ def index
...
         @teams = Team.order(:kind, :name)
       end
-      if params[:filter] != 'all'
+
+      if params[:filter] == 'full_time'
+        @teams = Team.in_current_season.full_time.ordered
+      elsif params[:filter] == 'part_time'
+        @teams = Team.in_current_season.part_time.ordered
+      elsif params[:filter] != 'all'
         @teams = Team.in_current_season.selected.ordered
```
klappradla: Hm, why do we even have the `selected` scope around any more? Isn't this the same as `accepted`?

app/models/team.rb
```diff
@@ -46,6 +46,8 @@ class Team < ApplicationRecord
   }
 
   scope :accepted, -> { where(kind: %w(full_time part_time)) }
+  scope :full_time, -> { where(kind: 'full_time') }
```
klappradla: Can you add tests for these new scopes?

ramonh:
Thanks @klappradla !

Apologies, I'm having a hard time hitting what should be tested and what already is. 😭 

As for the legend, this is currently at the bottom of the list. What's perhaps missing is an indicator for when a team is not selected.

ramonh:
I've taken care of these, @klappradla thank you very much for the feedback! Could you please take another look?

klappradla:
All good 🎉

Issue #949, opened 2018-02-26
User profile: Own invisible teams are hidden, too


ramonh:
Oops! 

Some coaches aren't able to confirm their team membership, because the teams are hidden for everybody.

klappradla:
Closed via #950 

🙏 @ramonh

PR #950, opened 2018-02-26
Show all teams if user is same as current

ramonh:
Related issue #949 

This will show all teams if user is equals to signed in user

### DIFF
--- a/spec/helpers/users_helper_spec.rb
+++ b/spec/helpers/users_helper_spec.rb
@@ -1,0 +1,36 @@
+require 'rails_helper'
+require 'cancan/matchers'
+
+RSpec.describe UsersHelper, type: :helper do
+
+  describe "#teams_for?" do
+    let!(:user) { create :user }
+    let!(:invisible_team) { create :team, invisible: true, kind: nil }
+    let(:role_name) { 'student' }
+
+    before do
+      invisible_team.attributes = { roles_attributes: [{ name: role_name, user_id: user.id }] }
+      invisible_team.save!
+    end
+
+    context 'user signed in is viewing own teams' do
+      before do
+        allow(self).to receive(:current_user).and_return(user)
+      end
+
+      it 'returns all teams' do
+        expect(teams_for user).to eq([invisible_team])
+      end
+    end
+
+    context 'user not signed in is viewing teams' do
+      before do
+        allow(self).to receive(:current_user).and_return(nil)
+      end
+
+      it 'returns only visible teams' do
+        expect(teams_for user).to eq([])
+      end
+    end
+  end
+end
--- a/spec/helpers/users_helper_spec.rb
+++ b/spec/helpers/users_helper_spec.rb
@@ -1,16 +1,16 @@
 require 'rails_helper'
-require 'cancan/matchers'
 
 RSpec.describe UsersHelper, type: :helper do
 
-  describe "#teams_for?" do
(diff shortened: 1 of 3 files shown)


spec/helpers/users_helper_spec.rb
```diff
@@ -0,0 +1,36 @@
+require 'rails_helper'
+require 'cancan/matchers'
+
+RSpec.describe UsersHelper, type: :helper do
+
+  describe "#teams_for?" do
```
klappradla: Typo: the `? ` is extra.

spec/helpers/users_helper_spec.rb
```diff
@@ -0,0 +1,36 @@
+require 'rails_helper'
+require 'cancan/matchers'
```
klappradla: Why are you requiring these matchers here? I don't see them being necessary here...

spec/helpers/users_helper_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+
+RSpec.describe UsersHelper, type: :helper do
+
+  describe "#teams_for?" do
+    let!(:user) { create :user }
```
klappradla: No need for bang here.

spec/helpers/users_helper_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+    let!(:invisible_team) { create :team, invisible: true, kind: nil }
```
klappradla: Also no need for bang here.

spec/helpers/users_helper_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+    let(:role_name) { 'student' }
+
+    before do
+      invisible_team.attributes = { roles_attributes: [{ name: role_name, user_id: user.id }] }
```
klappradla: Personal preference, but I find this a bit too verbose...

How about just creating the role?
```ruby
before { create(:role, name: 'student', user: user, team: invisible_team) }
```

spec/helpers/users_helper_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+      invisible_team.save!
+    end
+
+    context 'user signed in is viewing own teams' do
+      before do
+        allow(self).to receive(:current_user).and_return(user)
+      end
+
+      it 'returns all teams' do
+        expect(teams_for user).to eq([invisible_team])
```
klappradla: Again personal preference, but I think it would be more convenient if you'd also add a visible team for the user which then could be expected in the latter example.

klappradla: Cosmetic thing, but I prefer using `contain_exactly` instead of `eq` and building a new array.
```ruby
.to contain_exactly(invisible_team)
```

spec/helpers/users_helper_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+      end
+    end
+
+    context 'user not signed in is viewing teams' do
+      before do
+        allow(self).to receive(:current_user).and_return(nil)
```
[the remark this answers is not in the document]
klappradla: Did not look into the link, but the approach looks valid to me 👍

ramonh:
Done! Thank you @klappradla, could you please take another look?

klappradla:
🥇

PR #951, opened 2018-02-26
Temporarily disable student submitted notification

ramonh:
There was an incident where students got their names on the opposite application forms. 

In the interest of privacy, we'll disable this feature until we can figure this one out.


ramonh:
Thanks for merging that so quickly @carpodaster !! ❤️

PR #956, opened 2018-02-28
Fix crash when exporting applications to csv

ramonh:
We were using nonexistent accessor methods 😊


carpodaster:
Nothing about this PR, but it got me thinking: should we override `method_missing` to attempt getting a value in `application_data`?

klappradla:
Nice idea @carpodaster 👍 

I just see the tiny downside of things then _silently_ failing without us noticing 🤔 We don't have any tests covering these views in the app, so the only way we can find out if they don't work is either by coincidence or via an error that's reported. And if it just silently fails and misbehaves, we'd miss that and it will never get fixed.
Just a thought ☝️

carpodaster:
@klappradla your argument about the silent fails is very convincing. Let's not do it.

PR #959, opened 2018-02-28
Tweak mentor comment text for transparency

ramonh:
As discussed with @alicetragedy, it makes sense to update this field to show that the selection committee will have access to these comments.


ramonh:
Thank you both for checking! <3

PR #962, opened 2018-03-07
Remove unused "Less than 40 hours" and "volunteer team" application flags

ramonh:
Since we're switching to having part-and full time teams, this flag is no longer needed.

### DIFF
--- a/app/models/application.rb
+++ b/app/models/application.rb
@@ -5,13 +5,7 @@ class Application < ApplicationRecord
   PROJECT_VISIBILITY_WEIGHT = ENV['PROJECT_VISIBILITY_WEIGHT'] || 2
   COACHING_COMPANY_WEIGHT = ENV['COACHING_COMPANY_WEIGHT'] || 2
   MENTOR_PICK_WEIGHT = ENV['MENTOR_PICK_WEIGHT'] || 2
-  FLAGS = [:remote_team,
-          :volunteering_team,
-          :selected,
-          :male_gender,
-          :zero_community,
-          :age_below_18,
-          :less_than_two_coaches]
+  FLAGS = Selection::Table::FLAGS + [:selected]
 
   belongs_to :application_draft
   belongs_to :team, inverse_of: :applications, counter_cache: true
--- a/app/models/application.rb
+++ b/app/models/application.rb
@@ -11,8 +11,7 @@ class Application < ApplicationRecord
           :male_gender,
           :zero_community,
           :age_below_18,
-          :less_than_two_coaches,
-          :less_than_40_hours_a_week]
+          :less_than_two_coaches]
 
   belongs_to :application_draft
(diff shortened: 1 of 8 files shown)


nerdyvaishali:
Thanks @ramonh for working on this. Looks good to me. 

I think removing volunteer flags was also in that list I made along with this stuff. Although I was under the impression that you did that change at lot of other parts of the reviewing system. All I can see at the moment is that the application table have column 'Volunteer (h per week)' which I'm not sure why is named as 'volunteer'. 🤔

On a side note, thanks @klappradla on accidentally clearing one of my confusion. I was also wondering why same constant is defined at two places. 🙂

ramonh:
Thanks for the feedback, all! I'm on it 👍

ramonh:
And done! Could you please take another look? ✌️

app/models/application.rb
```diff
@@ -5,14 +5,7 @@ class Application < ApplicationRecord
...
   MENTOR_PICK_WEIGHT = ENV['MENTOR_PICK_WEIGHT'] || 2
-  FLAGS = [:remote_team,
-          :volunteering_team,
-          :selected,
-          :male_gender,
-          :zero_community,
-          :age_below_18,
-          :less_than_two_coaches,
-          :less_than_40_hours_a_week]
+  FLAGS = Selection::Table::FLAGS + [:selected]
```
klappradla: Good job 👍 

To be honest, I fear that skipping the `selected` flag on the table was probably a mistake - I think it's fine to just use the exact same collection in both cases 🤞

app/views/reviewers/applications/index.html.slim
```diff
@@ -21,7 +21,7 @@ table#applications.table.table-striped.table-bordered.table-condensed.table-hove
       th Location
       th = link_to_ordered 'Coaching Company', :coaching_company
       th.text-center colspan="2" Money ($)
-      th.text-center colspan="2" Volunteer (h per week)
+      th.text-center colspan="2" Hours per week
```
klappradla: ✔️

app/models/application.rb
ramonh: I'm not sure. I'm all for merging the flags, and adding the `:selected` one to the selection table won't break anything, but I wonder if it makes sense, as selected teams would never show up during the selection process, wouldn't they?

klappradla: I don't remember the exact details of the selection process. But I'd say they indeed need to show up in the table as I don't see another way for reviewers to have an overview on which teams were selected.

The actual "reviews" are based on each reviewer's `todos` which is a second list view. The table shows all the teams with their points, review status etc. and is then the place where teams will be selected.

ramonh: That makes sense! I'll make this change. Thanks for the feedback

ramonh:
Should be good to go. Thanks for the feedback! I'll go ahead and merge :)

PR #963, opened 2018-03-07
Add pre-validation hook to ensure application flags are unique

ramonh:
Related issue https://github.com/rails-girls-summer-of-code/selection_process/issues/58

This runs a pre-validation hook on applications to ensure that the flags are unique.

The reason I went with `before_validation` is as described in the [Rails docs](http://api.rubyonrails.org/classes/ActiveRecord/Callbacks.html), that this is the step where attributes should be adapted.

### DIFF
--- a/app/models/application.rb
+++ b/app/models/application.rb
@@ -22,6 +22,8 @@ class Application < ApplicationRecord
 
   validates :team, :application_data, presence: true
 
+  before_validation :ensure_flags_unique
+
   scope :hidden, -> { where('applications.hidden IS NOT NULL and applications.hidden = ?', true) }
   scope :visible, -> { where('applications.hidden IS NULL or applications.hidden = ?', false) }
 
@@ -81,5 +83,9 @@ class Application < ApplicationRecord
 
   def project2
     Project.find_by(id: application_data['project2_id'])
+  end
+
+  def ensure_flags_unique
+    flags.uniq!
   end
 end
--- a/app/models/application.rb
+++ b/app/models/application.rb
@@ -22,7 +22,7 @@ class Application < ApplicationRecord
 
   validates :team, :application_data, presence: true
 
-  before_validation :ensure_flags_unique
+  before_validation :remove_duplicate_flags
 
   scope :hidden, -> { where('applications.hidden IS NOT NULL and applications.hidden = ?', true) }
   scope :visible, -> { where('applications.hidden IS NULL or applications.hidden = ?', false) }
@@ -85,7 +85,7 @@ class Application < ApplicationRecord
     Project.find_by(id: application_data['project2_id'])
   end
 
-  def ensure_flags_unique
+  def remove_duplicate_flags
     flags.uniq!
   end
 end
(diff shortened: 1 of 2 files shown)


app/models/application.rb
```diff
@@ -22,6 +22,8 @@ class Application < ApplicationRecord
 
   validates :team, :application_data, presence: true
 
+  before_validation :ensure_flags_unique
+
```
F3PiX: Since `flags` is an attribute on `Application`, wouldn't it be easier to add a validator instead of the before hook?  And back it up with a database constraint.  
If that is possible at all, I am not totally sure about how the flags are set. 

If that ^ is not possible, maybe add an error message to the :ensure_flags_unique method? Because this will fail silently, right?

app/models/application.rb
```diff
@@ -82,4 +84,8 @@ def project1
   def project2
     Project.find_by(id: application_data['project2_id'])
   end
+
+  def ensure_flags_unique
+    flags.uniq!
+  end
```
F3PiX: See other comment, about error message 💣

app/models/application.rb
ramonh: Thank you for the feedback!

Hm, I'm afraid I don't understand the question. What I'm doing here is cleaning up, not validating. Maybe the problem is in the name? Maybe `cleanup_flags` would be better?

What do you thinK?

F3PiX: LOL As I totally misunderstood this as cleaning up, that may be a better name :-)
(I don't have access to the issue you are linking.) 
So, is the goal accomplished when the cleaning up happens _only_  when you update or create a validation?

ramonh: Sweet! Thanks for confirming that this was a bad naming issue :) 

This is meant to be run before the record is validated. That is, the flags will automatically be cleaned up before the application is validated (and by extension, saved). 

Hope that makes sense!

spec/models/application_spec.rb
```diff
@@ -75,6 +75,13 @@
       subject.send(:"#{flag}=", '0')
       expect(subject.flags).not_to include(flag)
     end
+
+    it 'ensures flags are unique' do
+      flag = flags.sample.to_s
+      subject.flags = [flag, flag]
+      subject.save!
```
klappradla: Just a suggestion: maybe we can save some microseconds by not persisting the record? Something like:
```ruby
it 'ensures flags are unique' do
  subject.flags = %w(fun fun)
  expect { subject.validate }.to change(subject, :flags).from(%w(fun fun)).to(%w(fun))
end
```
But no biggie, the way it is put right now is also perfectly fine.

app/models/application.rb
```diff
@@ -82,4 +84,8 @@ def project1
   def project2
     Project.find_by(id: application_data['project2_id'])
   end
+
+  def remove_duplicate_flags
```
klappradla: Since we're not using this outside of the scope of the `Application` model, this should definitely be in the `private` section 👮

ramonh: My bad! Makes sense, will do.

spec/models/application_spec.rb
ramonh: Makes sense to me, good call!

klappradla:
⛵️💨

nerdyvaishali:
Thanks @ramonh for fixing this.

Issue #939, opened 2018-02-19
Show mentors' comments in review process


ramonh:
Comments left by mentors on applications should be visible in the review process

carpodaster:
@ramonh can this be closed now that #965 has been merged?

ramonh:
Indeedy! Thanks @carpodaster

PR #965, opened 2018-03-11
Show mentor comments in application review

ramonh:
Related issue #939 

This will grab all mentors' comments on an application and display them to the reviewer.

### DIFF
--- a/app/controllers/reviewers/applications_controller.rb
+++ b/app/controllers/reviewers/applications_controller.rb
@@ -28,7 +28,6 @@ module Reviewers
 
     def show
       @application = Application.includes(:team, :project, :comments).find(params[:id])
-      grab_mentor_comments
       @rating = @application.ratings.find_or_initialize_by(user: current_user)
       @breadcrumbs << ["Application ##{@application.id}", (self.class::PATH_PARENTS + [@application])]
     end
@@ -79,10 +78,6 @@ module Reviewers
         hide_flags: Selection::Table::FLAGS.select { |f| session[f] }
       }
       Selection::Table.new(applications: Application.rateable, options: options)
-    end
-
-    def grab_mentor_comments
-      @mentor_comments = Mentor::Comment.where(commentable_id: @application.id)
     end
   end
 end
--- a/app/controllers/reviewers/applications_controller.rb
+++ b/app/controllers/reviewers/applications_controller.rb
@@ -28,6 +28,7 @@ module Reviewers
 
     def show
       @application = Application.includes(:team, :project, :comments).find(params[:id])
+      grab_mentor_comments
       @rating = @application.ratings.find_or_initialize_by(user: current_user)
       @breadcrumbs << ["Application ##{@application.id}", (self.class::PATH_PARENTS + [@application])]
     end
@@ -78,6 +79,10 @@ module Reviewers
         hide_flags: Selection::Table::FLAGS.select { |f| session[f] }
       }
       Selection::Table.new(applications: Application.rateable, options: options)
+    end
+
+    def grab_mentor_comments
+      @mentor_comments = Mentor::Comment.where(commentable_id: @application.id)
     end
   end
 end
--- a/app/controllers/reviewers/applications_controller.rb
+++ b/app/controllers/reviewers/applications_controller.rb
@@ -70,7 +70,7 @@ module Reviewers
     end
 
     def mentor_comments
-      Mentor::Comment.where(commentable_id: @application.id).where('created_at > ?', Date.parse('12.03.2018'))
+      Mentor::Comment.where(commentable_id: @application.id).where('created_at > ?', Date.parse('2018.03.12'))
     end
 
     def persist_order
--- a/app/controllers/reviewers/applications_controller.rb
+++ b/app/controllers/reviewers/applications_controller.rb
@@ -5,6 +5,7 @@ module Reviewers
   class ApplicationsController < Reviewers::BaseController
     before_action :store_filters, only: :index
     before_action :persist_order, only: :index
+    helper_method :mentor_comments
     respond_to :html
 
     PATH_PARENTS = [:reviewers]
@@ -66,6 +67,10 @@ module Reviewers
       Selection::Table::FLAGS.each do |key|
         session[key] = params[:filter][key] == 'true' if params.dig(:filter, key)
       end
+    end
+
+    def mentor_comments
+      Mentor::Comment.where(commentable_id: @application.id).where('created_at > ?', Date.parse('12.03.2018'))
     end
 
     def persist_order
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -175,10 +175,6 @@ module ApplicationHelper
     link_to title, params.except('action', 'controller').permit!.merge(sort: column, direction: direction)
   end
 
-  def mentor_comments_for_application(application)
-    Mentor::Comment.where(commentable_id: application.id).where('created_at > ?', Date.parse('12.03.2018'))
-  end
-
   def user_for_comment(comment)
     if comment.user.nil?
       "Deleted user"
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -175,6 +175,10 @@ module ApplicationHelper
     link_to title, params.except('action', 'controller').permit!.merge(sort: column, direction: direction)
   end
 
+  def mentor_comments_for_application(application)
+    Mentor::Comment.where(commentable_id: application.id).where('created_at > ?', Date.parse('12.03.2018'))
+  end
+
   def user_for_comment(comment)
     if comment.user.nil?
       "Deleted user"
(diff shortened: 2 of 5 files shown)


app/controllers/reviewers/applications_controller.rb
```diff
@@ -79,5 +80,9 @@ def applications_table
       }
       Selection::Table.new(applications: Application.rateable, options: options)
     end
+
+    def grab_mentor_comments
```
carpodaster: I'd rather have this as a lazily loaded helper method. This would remove the somewhat procedural statement in line 31 and the call to an instance variable in the template view.

```ruby
# in the controller

…

private

def mentor_comments
  @mentor_comments ||= Mentor::Comment.where(commentable_id: @application.id)
end
helper_method :mentor_comments # or up top, don't know if we have a style guide for this /shrug
```

spec/controllers/reviewers/applications_controller_spec.rb
```diff
@@ -89,6 +89,17 @@
...
       before { sign_in user }
 
+      context 'given the application has mentor comments' do
+        let(:mentor) { create :mentor }
+        let!(:mentor_comment) { Mentor::Comment.create(user: user, commentable_id: application.id, text: 'something') }
+
+        before { get :show, params: { id: application } }
+
+        it 'assigns @mentor_comments' do
+          expect(assigns :mentor_comments).to contain_exactly(mentor_comment)
```
carpodaster: I know we check for `assigns` all over the place but this is discouraged in recent Rails versions. Can you regexp-match for a unique comment string in the `response.body` instead?

ramonh:
Good point, @carpodaster thanks for the review! I'll make the change to only show them after today.

I did in a previous PR (few weeks ago) update that message for the current season. In your opinion does it need further tweaking?

app/helpers/application_helper.rb
```diff
@@ -175,6 +175,10 @@ def sortable(column, title = nil)
     link_to title, params.except('action', 'controller').permit!.merge(sort: column, direction: direction)
   end
 
+  def mentor_comments_for_application(application)
```
carpodaster: Oh sorry, I should have been more clear about the "helper method" bit. I meant to leave it in the controller; basically making the private/protected method that you added available to the view using the `helper_method` macro. We already know the application instance so not having to add it as a functional argument improves readability imho.

spec/controllers/reviewers/applications_controller_spec.rb
```diff
@@ -95,8 +95,8 @@
 
         before { get :show, params: { id: application } }
 
-        it 'assigns @mentor_comments' do
-          expect(assigns :mentor_comments).to contain_exactly(mentor_comment)
+        it 'shows mentor comment' do
+          expect(response.body).to match(mentor_comment.text)
```
carpodaster: 💟 

One minor (optional) thing: can you change the `text` value for the record that's being created in line 94 to something more specific so that we don't accidentally match on a string that's present through other means?

app/helpers/application_helper.rb
```diff
@@ -175,6 +175,10 @@ def sortable(column, title = nil)
...
+    Mentor::Comment.where(commentable_id: application.id).where('created_at > ?', Date.parse('12.03.2018'))
```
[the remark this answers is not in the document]
ramonh: Makes sense! On it

app/helpers/application_helper.rb
ramonh: Ahh! Didn't know that was a thing. 😅 Thank you for the feedback!

spec/controllers/reviewers/applications_controller_spec.rb
ramonh: Sounds good! Will do.

ramonh:
Thank you for the feedback @carpodaster! I think I should have handled everything. Could you please take another 👀 ?

app/controllers/reviewers/applications_controller.rb
```diff
@@ -68,6 +69,10 @@ def store_filters
       end
     end
 
+    def mentor_comments
+      Mentor::Comment.where(commentable_id: @application.id).where('created_at > ?', Date.parse('2018.03.12'))
```
klappradla: Maybe I'm missing context, but what is the condition on `created_at` for? And why is it hardcoded to a certain value...?

app/views/reviewers/applications/_mentor_feedback.html.slim
```diff
@@ -15,3 +15,11 @@ p = show_or_na(project2_sign_off.try(:name))
 
 b = Application.data_label("mentor_fav_project2")
 p = heart_or_na(ad["mentor_fav_project2"])
+
+h3 Mentor comments
+- if comments.empty?
+  p n/a
+- else
+  - comments.each do |comment|
+    h5 = "#{ user_for_comment(comment) }, #{ l(comment.created_at, format: :short) }".html_safe
```
klappradla: No need for `.html_safe` magic here in my opinion, would:
```ruby
h5 #{ user_for_comment(comment) }, #{ l(comment.created_at, format: :short) }
```
do the same?

app/views/reviewers/applications/show.html.slim
```diff
@@ -46,7 +46,7 @@
   hr
 
   h2 Mentors' feedback
-  = render partial: 'mentor_feedback', locals: { application: @application }
+  = render partial: 'mentor_feedback', locals: { application: @application, comments: mentor_comments }
```
klappradla: Personal taste, but I really dislike the `helper_method` pattern _(even more than helper modules themselves)_. In my opinion, it makes the already fuzzy process of making things available in views even fuzzier, all that including enough magic to make it really hard to understand what's happening behind the scenes: when is the method invoked? It is an additional request?

I think the best way of staying sane with Rails not being explicit about these things is an instance variable in the controller and then `assigns.fetch('instance-var')` in the view to be absolutely sure both things speak the same language.

Just wanted to have this mentioned ☝️ no need to change it though 😉

app/controllers/reviewers/applications_controller.rb
```diff
@@ -5,6 +5,7 @@ module Reviewers
   class ApplicationsController < Reviewers::BaseController
     before_action :store_filters, only: :index
     before_action :persist_order, only: :index
+    helper_method :mentor_comments
```
klappradla: I personally really dislike the `helper_method` thing... Aside personal feelings: it would make sense to have this available only on the `show` action or at least actions where you'd also have an `@application` instance var available.

app/controllers/reviewers/applications_controller.rb
ramonh: Good question! This comes from @carpodaster 's review [here](https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/965#pullrequestreview-102945776) where it makes sense to only load recent mentor reviews, as previously these were private.

app/controllers/reviewers/applications_controller.rb
ramonh: Hmm, dunno if I'm reading this wrong, but it seems that there's no way to make this available to specific actions?
http://api.rubyonrails.org/classes/AbstractController/Helpers/ClassMethods.html

app/views/reviewers/applications/show.html.slim
ramonh: @klappradla thank you very much for your feedback here.

I'd originally gone with the way you suggested, as this was also present in other tests, but I'm fine with having it this way too. Since the taste seems to be more divisive as @carpodaster originally suggested, I'll be leaving it in for now, if no one minds. Maybe we should strive for consistency down the road, though.

ramonh:
Thank you both very much for the feedback! 

Since some of the actions here are to be thought of down the line, I'll go ahead and merge this for now and noted.

app/controllers/reviewers/applications_controller.rb
klappradla: Oh ok, then no worries 💚

app/views/reviewers/applications/show.html.slim
klappradla: Oh, no worries, feel free to go ahead with the helper 👍 

It's just my own personal preference that I hate them - they may actually make sense in some cases ✌️

app/controllers/reviewers/applications_controller.rb
klappradla: 😹 ah ok, now I get it, thanx!

app/views/reviewers/applications/show.html.slim
ramonh: Your personal preference is much appreciate it! I'd love to know why they seem like a bad idea? 

Frankly, if they're as good as the same as helper modules, I think it would be good to write these all consistently.

klappradla: They aren't equal to helper modules. I think they have a closes scope and are only available in views rendered via the controller they're defined in. Plus they can of course also be used in the controller itself.

The reason I dislike them: they blurry the view and hide where things come from 😉 If one assigns an instance variable in the controller it's a bit more obvious from the start where things in the view come from.

But as I said, no biggie, some people like this pattern and there may be obvious reasons to use it - such as the `current_user` helper.

And for view helper modules: I personally think, if it's so much more effort to make things useful for a view, adding a decorator or a view object is an approach that's easier to maintain on the long term then view helpers.
I vaguely recall that things have changed since Rails 5, but before, view helpers were available globally in all views 🙈

PR #968, opened 2018-03-20
Fix bug where valid mentor comments weren't shown

ramonh:
Some mentor comments weren't being shown because they were younger than the cutoff date. However, since some are younger than the given date, and we [fixed the transparency notice](#959) on 28.02.2018, we can show earlier ones.

### DIFF
--- a/app/controllers/reviewers/applications_controller.rb
+++ b/app/controllers/reviewers/applications_controller.rb
@@ -70,7 +70,7 @@ module Reviewers
     end
 
     def mentor_comments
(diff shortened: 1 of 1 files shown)


ramonh:
Since only an arbitrary date was changed from #965, I'll go ahead and merge this, if no one minds 🙈

klappradla:
😹 was already looking at it

ramonh:
Arg, sorry, I should'a given it a moment. 😭

PR #972, opened 2018-03-26
Update gems

klappradla:
Update rails-html-sanitizer and loofah again cc @ramonh


ramonh:
Works locally! 🤞 Thank you @klappradla !

Issue #975, opened 2018-03-27
Application table: “Flags” column shows wrong content


alicetragedy:
In the application table for reviewers, the column “Hours per Week” has been split into two sections: the hours per week selected by the applicants, as well as a section that should contain the explanation for these selected hours. However, the content for that section seems to have moved over to the “Flags” column and vice versa (meaning that application flags are shown under “Hours per Week”, and reasons for hours per week are shown in the “Flags” column) — see screenshot below.

![screenshot](https://user-images.githubusercontent.com/3143348/37972491-c88441d2-31d8-11e8-8a04-b4ff14b6e68b.jpg)

ramonh:
Oops! I just saw this 🙊  Thanks for noticing! I'll get on this.

klappradla:
Closed via #978

PR #978, opened 2018-04-04
Fix bug where columns were swapped around

ramonh:
Related issue #975 

This aims to fix the above issue by swapping the positions of the columns around, putting them back in their rightful place.


ramonh:
Thanks for merging this! I'm curious, what makes you cautious about it?

klappradla:
Just thought: maybe this can be done with "normal" linebreaks and without an `html_safe` thing. I feel like often times `html_safe` is used, there's usually a way to do it cleaner 😉

PR #764, opened 2017-06-22
[WIP] Document All The Things!

acnagy:
This PR is for re-writing the readme and relevant docs for the teams app - see https://github.com/rails-girls-summer-of-code/rgsoc-teams/issues/606 for the discussion

PR #984, opened 2018-04-25
Update README with the doc-improvements from Anna

klappradla:
This update / formats / all-things the README with the documentation improvements @acnagy started in #764 👷‍♀️ 

Additional changes:
- removed some super obsolete sections
- moved some things into the troubleshooting guide instead

**Please note**
I did all of this with basically only one side of my brain left - I may have forgotten or overseen some important parts (cc @acnagy)  😴 

**Further suggestions**
- fully remove the `mailtrap` part (also from the app)
- fully remove the `google places` 
- remove the Procfile setup

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,7 +150,7 @@ To access all functionality of the app, add yourself as an organizer. For this, 
 
 Once you're an `organizer`, you can:
 - add a season and switch between the phases of a season: http://localhost:3000/organizers/seasons
-- impersonate other users to test the system from a different perspective: http://localhost:3000/user _(this only works in development mode)_
+- impersonate other users to test the system from a different perspective: http://localhost:3000/community _(this only works in development mode)_
 
 💁‍ you are good to go now. Happy coding!
 
--- a/README.md
+++ b/README.md
@@ -156,37 +156,13 @@ Once you're an `organizer`, you can:
 
 ## Optional Setup
 
-**Note**: If you configure any of these optional dependencies, you need to run the project via the `Procfile` (e.g. using the [foreman](https://github.com/ddollar/foreman) tool).
-
-### Mailtrap
-
-By default emails are printed to the rails server log and never actually sent from your development environment. If you want to test sending emails to a real inbox, we recommend you use [Mailtrap](https://mailtrap.io) to avoid accidentally sending out emails to real users.
-
-Copy the `env-example` file and replace `InboxUsername` and `InboxPassword` with your Mailtrap inbox credentials:
-
-    cp .env-example .env
-
-Now, when running the `foreman` command **before** any command, this configuration will be loaded into the environment:
-
-    foreman run rails server
-
-    foreman run rails console
-
-💁 Not working? Check our **[Troubleshooting Guide](TROUBLESHOOTING.md)**.
-
 ### Google Places API
 
-To avoid accidentally exceeding the rate limit on [Google's Places API][google-places] (e.g. when heavily testing city-autocomplete fields) and thus blocking its usage for other RGSoC sites and devs:
+To avoid accidentally exceeding the rate limit on [Google's Places API](https://developers.google.com/places) (e.g. when heavily testing city-autocomplete fields) and thus blocking its usage for other RGSoC sites and devs:
 
-Get your own [API key](https://developers.google.com/places/javascript/), and copy the `.env-expample` file _(if you haven't already)_ to replace `YourGoogleMapsAPIKey` with it:
+Get your own [API key](https://developers.google.com/places/javascript/), and set the `GOOGLE_MAPS_API_KEY` environment variable *(or simply override the key in `config/initializers/google_maps.rb`)*:
 
-        cp .env-example .env
-
-Now, when running the `foreman` command **before** any command, this configuration will be loaded into the environment:
-
-    foreman run rails server
-
-    foreman run rails console
+    export GOOGLE_MAPS_API_KEY=<your-api-key>
 
 ## Testing
 
--- a/README.md
+++ b/README.md
@@ -136,19 +136,18 @@ To access all functionality of the app, add yourself as an organizer. For this, 
 
 1. In the browser 🖥: log in with your Github account
 1. In the console 📟
+    ```ruby
+    user = User.last
+    user.roles.create(name: 'organizer')
 
-  ```ruby
-  user = User.last
-  user.roles.create(name: 'organizer')
+    # alternative: find your user, e.g. by your Github handle
+    user = User.find_by(github_handle: <your-github-handle-here>)
 
-  # alternative: find your user, e.g. by your Github handle
-  user = User.find_by(github_handle: <your-github-handle-here>)
-
-  # You can assign yourself other roles the same way.
-  # If you've assigned yourself student AND organizer roles however,
-  # this may lead to undesired behavior of the app. Best remove the
-  # student role again.
-  ```
+    # You can assign yourself other roles the same way.
+    # If you've assigned yourself student AND organizer roles however,
+    # this may lead to undesired behavior of the app. Best remove the
+    # student role again.
+    ```
 1. Refresh the browser 🖥: you should see links to orga pages now
 
 Once you're an `organizer`, you can:
--- a/README.md
+++ b/README.md
@@ -59,12 +59,18 @@ sudo -u postgres createuser -P -s rgsoc
 #### Setup on macOS 🍏
 ```bash
 # Install required packages
-$ brew install ruby postgres chromedriver
+brew install ruby postgres
 # Make sure to follow the instructions printed on the screen for postgres
-$ gem install bundler
+
+# Install chromedriver
+brew tap caskroom/cask
+brew cask install chromedriver
+
+# Install bundler
+gem install bundler
 
 # Create database user rgsoc with password rgsoc
-$ createuser -P -s rgsoc
+createuser -P -s rgsoc
 Enter password for new role: rgsoc
 Enter it again: rgsoc
 ```
--- a/README.md
+++ b/README.md
@@ -74,8 +74,8 @@ gem install bundler
 
 # Create database user rgsoc with password rgsoc
 createuser -P -s rgsoc
-Enter password for new role: rgsoc
-Enter it again: rgsoc
+# Enter password for new role: rgsoc
+# Enter it again: rgsoc
 ```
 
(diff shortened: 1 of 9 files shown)


klappradla:
Awesome @pgaspar 💚 
I even thought about putting this into an `install.sh` to have the steps verified in CI to be informed about changes like this.
```
❯ brew search chromedriver
==> Searching local taps...
==> Searching taps on GitHub...
caskroom/cask/chromedriver
==> Searching blacklisted, migrated and deleted formulae...

If you meant "chromedriver" specifically:
It was migrated from homebrew/core to caskroom/cask.
You can access it again by running:
  brew tap caskroom/cask
```

README.md
```diff
@@ -5,77 +5,87 @@
...
 * Make sure to check our [Contribution Guide](CONTRIBUTING.md)
 
-## System Requirements
+## Getting Started
+
+In order to contribute, you need to run the Teams App locally on your computer. For setting it up, you'll need to have some tools installed on your system:
 
 * PostgreSQL 9.5 or newer
 * Ruby 2.5.1
 * [Chromedriver](https://sites.google.com/a/chromium.org/chromedriver/)
```
F3PiX: Rails may come in handy too ;-)
Shall we put the system requirements at the top of the file? I see that a lot, I like that. 
And keep the set up of those in the Installation sections.

klappradla: > Rails may come in handy too ;-)

☝️ why? The app needs a particular version _(that will be installed as a dependency when `bundle install`)_ - so I don't really see a point in people installing a potentially non-matching version first. Or did I get your point wrong?

For putting the requirements to the top of the file: this makes 💯 sense for libraries and other things that you'd run as a dependency. For the sake of readability in our case, I'd personally keep it close to the _"I want to setup / work with this app locally"_ to not scatter the information into different corners of the README.

F3PiX: I think it's good PR to show potentially new contributors that we are on a 5+ rails version, and from my experience, it saves time if the project shows that upfront in the Readme.
I see that it is not a 'system requirement'; maybe we can add a online 'stack' heading? Up to you.

README.md
```diff
@@ -84,87 +94,116 @@ Inside database.yml add username and password for development and test:
...
-- Once you are an `organizer`, you can add a season and switch between season's phases at
-http://localhost:3000/organizers/seasons in your browser.
-- While in development, you are also able to impersonate other users to easily test the system
-as someone else. Go to http://localhost:3000/users while logged in to do that.
-- You are good to go now. Happy coding!
+1. Refresh the browser 🖥: you should see links to orga pages now
+
+Once you're an `organizer`, you can:
+- add a season and switch between the phases of a season: http://localhost:3000/organizers/seasons
+- impersonate other users to test the system from a different perspective: http://localhost:3000/user _(this only works in development mode)_
```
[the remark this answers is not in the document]
klappradla: Thanx for pointing out 🙏

README.md
```diff
@@ -84,87 +94,116 @@ Inside database.yml add username and password for development and test:
...
+
+💁‍ you are good to go now. Happy coding!
+
+## Optional Setup
+
+**Note**: If you configure any of these optional dependencies, you need to run the project via the `Procfile` (e.g. using the [foreman](https://github.com/ddollar/foreman) tool).
+
+### Mailtrap
+
+To avoid accidentally sending out mails to real addresses you can create a free account at [Mailtrap](https://mailtrap.io) with an inbox to _trap_ emails sent from your development environment.
```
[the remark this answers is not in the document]
klappradla: Valid point 👍 

I think I already made a quick poll for who's actually using Mailtrap - none - and if we could just swap it for something like letter opener...

I'd personally like to get rid of the whole foreman setup. Will do another quick poll in Slack 😉

README.md
klappradla: Ok, got your point, thanx 🙏 
As every character we add to the README is actually code that needs to be maintained, I chose to however not add it 😉 The _stack_ section would need to be kept up to date and in sync with the rest of the README. Looking at the very first introductory section which was out of date for years shows that this is a hard thing to do 🙈

README.md
```diff
@@ -5,77 +5,87 @@
...
-$ sudo apt-get install postgresql libpq-dev libcurl3 libcurl3-gnutls libcurl4-openssl-dev postgresql-contrib-9.5 chromium-chromedriver
+sudo apt-get install postgresql libpq-dev libcurl3 libcurl3-gnutls libcurl4-openssl-dev postgresql-contrib-9.5 chromium-chromedriver
+
 # Create database user rgsoc with password rgsoc
-$ sudo -u postgres createuser -P -s rgsoc
-Enter password for new role: rgsoc
-Enter it again: rgsoc
+sudo -u postgres createuser -P -s rgsoc
+# Enter password for new role: rgsoc
+# Enter it again: rgsoc
```
pgaspar: Maybe we could add the same `#`s on the 🍏 instructions for consistency 👍

README.md
```diff
@@ -5,77 +5,87 @@
...
+brew install ruby postgres
 # Make sure to follow the instructions printed on the screen for postgres
-$ gem install bundler
+
+# Install chromedriver
+brew tap caskroom/cask
+brew cask install chromedriver
+
+# Install bundler
+gem install bundler
```
pgaspar: I noticed we're missing this step on the 🐧 instructions. Should we add it?

I'm sure this isn't a problem for people experienced in Rails, but I can see it stumbling beginners.

README.md
```diff
@@ -84,87 +94,92 @@ Inside database.yml add username and password for development and test:
...
-      remove the student role.

-    * Refresh the browser and you should see links for organizers.

-- Once you are an `organizer`, you can add a season and switch between season's phases at
-http://localhost:3000/organizers/seasons in your browser.
-- While in development, you are also able to impersonate other users to easily test the system
-as someone else. Go to http://localhost:3000/users while logged in to do that.
-- You are good to go now. Happy coding!
+1. Refresh the browser 🖥: you should see links to orga pages now
```
pgaspar: Interestingly, Github is rendering this `1.` as `1.` and not as `3.` as I would expect it at least. Maybe it's because of the code block above?

README.md
```diff
@@ -201,8 +216,7 @@ at http://teams.railsgirlssummerofcode.org.
 
 Append `-r staging` or `-r production` to any `heroku` command in order to specify the app.
 
-This app uses [camo](https://github.com/atmos/camo) to proxy insecure images in activity logs
-when `CAMO_HOST` and `CAMO_KEY` environment variables are set.
+This app uses [camo](https://github.com/atmos/camo) to proxy insecure images in activity logs when `CAMO_HOST` and `CAMO_KEY` environment variables are set.
```
pgaspar: I've never used camo and so I was a bit confused when I read this. Here's some things that popped into my mind when I initially saw it. Hopefully they'll help us come up with clearer copy:

- Is this in the right section?
- Do I need to set this up to deploy to heroku?
- What happens if I don't set these up?
- In what situations would I want to set up my own camo environmental variables?

README.md
klappradla: Well, for the sake of _"consistency"_ probably yes, you're right - but for the sake of _"simplicity"_ no I'd say 😉 

Installing `postgres` via homebrew automatically generates a password-less role with the same name as the user logged in into the system -> that's exactly what Rails defaults to for a database `username` and `password` if it's skipped in the `database.yml` file.

The `database.yml.example` file has exactly this configuration - so it would work outta the box without modifying it.

In my opinion ☝️ is just easier, therefore I did not add any `role` instructions.

For 🐧 tbh, they're also just there because someone who works with Linux once specified them like this. Maybe it's also obsolete there... 🤔

README.md
klappradla: Oh, thanx for pointing out 🙏

README.md
klappradla: Good point!

I was however just wondering: can we not just ditch the whole `role` thing? If one installs `postgres` via homebrew, it will automatically create a password-less role with the same name as the user on the system -> that's exactly what Rails defaults to as `username` and `password` if it's not specified explicitly in the `database.yml` file (as the default `database.ym file and our `database.yml.example` also do).

Not sure if the same works for 🐧 though 🤔

README.md
klappradla: Hm, you're right. It should be more obvious that this is something like an "infrastructure setup" similar to cron jobs on Heroku... It's of course important knowledge for anyone operating the app, but it's not really part of "deploying the app".

I moved it to its own section and changed the wording a bit - let's see if this makes things more obvious.

README.md
pgaspar: I think so! I tried it on my VM and it worked, but I needed to remove `host: localhost` for it to work. Otherwise it was requiring a password, even if was trying to create a database from scratch using `rails db:setup`.

This may give some context into what's going on: https://stackoverflow.com/questions/23375740/pgconnectionbad-fe-sendauth-no-password-supplied

It may be the case that by default on linux `localhost` is set to only allow md5 connections, while on macOS it doesn't. It seems to be the case, looking at my own configurations.

```
# On Ubuntu:
host    all             all             127.0.0.1/32            md5

# On macOS:
host    all             all             127.0.0.1/32            trust
```

Since our `database.yml.example` has `host: localhost`, people on linux may see some errors 😟

Here's what I was seeing when trying to create the DB:

```
pedro@ubuntu:~/code/rgsoc-teams$ bundle exec rails db:setup
fe_sendauth: no password supplied
Couldn't create database for {"adapter"=>"postgresql", "database"=>"rgsocteams_development_2", "host"=>"localhost"}
rails aborted!
PG::ConnectionBad: fe_sendauth: no password supplied
bin/rails:4:in `<main>'
Tasks: TOP => db:setup => db:schema:load_if_ruby => db:create
(See full trace by running task with --trace)
```

My settings were:

```
development:
  adapter: postgresql
  database: rgsocteams_development_2
  host: localhost
```

The same settings work on macOS.

README.md
pgaspar: Looks great - thanks!

README.md
klappradla: Oh, you're so right I totally forgot about this 🙈 

Removing the `host` key sounds more of a hack / coincidence than an actual setting, so I think we're closed to a default that people are used to if we just stick to creating a role with a password.

I'll update the section accordingly ✌️

klappradla:
Are there any more comments, or should I just merge this now?

klappradla:
Yay! Merging PRs from my phone 🐒 

Thanx a ton for your help @pgaspar 💚

acnagy:
thanks so much for doing this @klappradla -- this is awesome! 🎉

PR #987, opened 2018-05-01
Improve impersonation copy

pgaspar:
This PR includes a few small copy changes that will hopefully make the impersonation feature a bit easier to use.

These changes are based on my experience when I was trying out the feature after setting up the project again this week.

### Changes

Added some copy on how to stop impersonation to the initial flash message:

screen shot 2018-05-01 at 16 30 53

Changed the button you can use to stop impersonation from "Stop" to "Stop Impersonation":

screen shot 2018-05-01 at 16 31 03

This is how it looks on mobile:

screen shot 2018-05-01 at 16 31 16

I was confused about how to stop impersonating, hence these changes. I had to revisit the original PR and current code to figure out where I had put the button 😊

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -61,7 +61,7 @@ class UsersController < ApplicationController
 
   def impersonate
     impersonate_user(@user)
-    redirect_to community_path, notice: "Now impersonating #{@user.name}."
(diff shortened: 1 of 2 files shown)


carpodaster:
:confetti_ball: !

PR #988, opened 2018-05-01
Generate dasherized Github handles on User factory

pgaspar:
This is a very small change for something I noticed when looking at some seed data: I was seeing names strings like `rodney o'keefe` in the User's `github_handle` attribute.

The change I'm proposing will create handles that are closer to what Github allows, for example: `rodney-o-keefe`.

It seems the `underscore` method will not do exactly what one might expect it to do, as is described in this issue: https://github.com/rails/rails/issues/4519

Replacing it with `parameterize` will turn something like `Becki O'Hara Zí` into `becki-o-hara-zi`, which would be a valid Github handle (it seems Github doesn't allow underscores).

There may be a reason for the original `underscore` method that I may be missing here - please let me know 🐱

### DIFF
--- a/spec/factories/users.rb
+++ b/spec/factories/users.rb
@@ -1,6 +1,6 @@
 FactoryBot.define do
   factory :user, aliases: [:member] do
-    github_handle { FFaker::Name.name.parameterize }
+    github_handle { FFaker::InternetSE.user_name_variant_short }
     name     { FFaker::Name.name }
     email    { FFaker::Internet.email }
     location { FFaker::Address.city }
--- a/spec/factories/users.rb
+++ b/spec/factories/users.rb
@@ -1,6 +1,6 @@
 FactoryBot.define do
   factory :user, aliases: [:member] do
-    github_handle { FFaker::Name.name.underscore }
+    github_handle { FFaker::Name.name.parameterize }
     name     { FFaker::Name.name }
     email    { FFaker::Internet.email }
     location { FFaker::Address.city }


spec/factories/users.rb
```diff
@@ -1,6 +1,6 @@
 FactoryBot.define do
   factory :user, aliases: [:member] do
-    github_handle { FFaker::Name.name.underscore }
+    github_handle { FFaker::Name.name.parameterize }
```
klappradla: Makes 💯 sense as the smallest possible change to put things straight here, fully agree 👍 

I didn't even know the user factory looks like this...

However, for making things even simpler and less likely to cause flaky specs, I'd even go for something like
```ruby
github_handle { FFaker::Internet.user_name }
```
What do you think?

pgaspar: Thank you @klappradla!

I tried to find something suitable in FFaker, but nothing seemed to be perfect...

`FFaker::Internet.user_name`, for example generates all these:

```
jewell.towne
wilda
dorthea_reichert
```

From these, I think only wilda is a valid Github username, and it won't generate other valid usernames like `rails-girls`.

Closest I found was `FFaker::InternetSE.login_user_name` which seems to only return simple words or words with underscores, but even underscores seem to be invalid for Github ([see here](https://github.com/shinnn/github-username-regex) - this is likely not very reliable, but I couldn't find any username with a `_` after a quick search).

🤷‍♀️ wish we could use something directly from FFaker but I can't find anything that would match Github's usernames...

Unless we use `FFaker::InternetSE.user_name_variant_short` which only returns single words - that would work. Not sure if it's much clearer though - what do you think?

ramonh: Hey all! Hope you don't mind my throwing in an opinion. I think the `user_name_variant_short` is a good compromise.

pgaspar: It's much appreciated @ramonh! 🙌

I have the commit with the change ready to push. I'll give @klappradla a chance to chime in as well before I do, though.

klappradla: As said @pgaspar: I'd even merge it as it is, was just a cosmetic comment.

If you want, feel fee to update, otherwise just give me you "go" and I merge this.

pgaspar: Pushed! Thank you :shipit:

Issue #983, opened 2018-04-23
User confirmation: verify that a user can always resend the confirmation email


carpodaster:
This should already be in place, let's just be 101% sure.

Issue #981, opened 2018-04-23
User confirmation: Allow user to update their email address


carpodaster:
A user must be able to fix potential typos with their email address when they're still in an unconfirmed state.

In fact, they should be able to update all of their profile data.

ramonh:
If nobody minds, I'll go ahead and take care of this, get the balling rolling 👍

Issue #980, opened 2018-04-23
User confirmation: Remove the 30 days grace period 


carpodaster:
In order to remove the uncertainty of what a newly registered user can or cannot do, we should move potential errors with a user's registration closer to when it actually happened.

Issue #869, opened 2017-11-08
Locked out of account


LuisaAPF:
It's been a month since I last signed in to my account (locally). It happens that I forgot to confirm my email and that is an active option that forbids my access after 30 days. Is there a way to overcome this?

carpodaster:
@LuisaAPF agreed, if that happens in production, that does seem like a bug. To verify if I got it right: you encounter this problem on your development machine only right now?

If you don't have much data, the easiest way would probably to drop and re-create your development database:

    bundle exec rake db:reset

Let me know if that fixed the problem locally. Would you like to work on some form of account recovery to prevent this lock out on production? I'm happy to help of course.

LuisaAPF:
Hi @carpodaster, I could solve the problem locally. But is it really necessary to forbid the user from accessing her/his account if the email address is not confirmed? I think it would be better to allow access at least to the profile page, in case a change in email is needed. What do you think?

carpodaster:
@LuisaAPF yes, you are absolutely right, access to the profile page should be possible. Maybe the app should even auto-redirect to the profile page after a login with an unconfirmed account whose grace period has expired? I don't have a way of testing out myself right now: can you log in at all with an unconfirmed-expired account? And if so, are you shown a flash notification that prompts you to validate the email address (this should be shown during the grace period)

LuisaAPF:
When my account was expired, I wasn't able to login. I got a 401 unauthorized status. There was no notification asking me to validate my email address. But this wouldn't have solved the problem if my email was wrong or inactive. So I think we should give the user the chance to update the email before re-sending the confirmation request.

In my opinion, the best way would be to never block the access to an account. Instead, we should forbid a user with an unconfirmed email to have access to certain functionalities, like creating a team.

F3PiX:
Lately, I have been thinking about pulling the account-part out of the User model, and into a separate UserAccount. I expect that it would be easier to handle all devise-functions in the UserAccount functionality. Is that something to consider? (@carsten, @max)

F3PiX:
Sorry, I mean @carpodaster , @klappradla ^

klappradla:
I personally wouldn't consider it @F3PiX, don't really see the pay-off (but you can of course always give it a try 👍 ). Is there actually much business logic tied to "accounts"? Isn't it all in the devise gem? We're also only using very very few features of devise 😉 

[...]

PR #989, opened 2018-05-02
Allow infinite confirmation period

ramonh:
Related issues #869 #983 #981 #980

Relatively small, but what's worth considering here:
- All actions outside of updating the profile now require being confirmed
- The confirmation period has been set to indefinite (nil), but with the above limitation, we can only allow profile completion
- Navigating the teams app while unconfirmed offers to resend the confirmation email

### DIFF
--- a/config/initializers/devise.rb
+++ b/config/initializers/devise.rb
@@ -103,7 +103,7 @@ Devise.setup do |config|
   # able to access the website for two days without confirming his account,
   # access will be blocked just in the third day. Default is 0.days, meaning
   # the user cannot access the website without confirming his account.
-  config.allow_unconfirmed_access_for = 30.days
+  config.allow_unconfirmed_access_for = nil
 
   # A period that the user is allowed to confirm their account before their
   # token becomes invalid. For example, if set to 3.days, the user can confirm


app/models/ability.rb
```diff
@@ -93,7 +93,7 @@ def initialize(user)
   end
 
   def signed_in?(user)
```
klappradla: Given that change - or eventually even before - the method name actually does not make that much sense, right?

How about just switching it to `signed_up_and_confirmed?`

Would make then intention clearer.

app/models/ability.rb
```diff
@@ -93,7 +93,7 @@ def initialize(user)
...
-    user.persisted?
+    user.persisted? && user.confirmed?
```
carpodaster: I'm not a huge fan of devise (because it does so many magic things under the hood). I feel like an unpersisted user should never be confirmed. But since we inject the user object, we cannot guarantee that unless we overwrite `User#confirmed?` to `super && persisted?` Is it worth it? :thinking:  (sorry, just thinking aloud)

app/models/ability.rb
ramonh: I think that's a good point, @klappradla I'll take care of this. Thank you!

app/models/ability.rb
ramonh: Oh, hm, no idea. Good point. I should look into what `#persisted?` does...

ramonh:
Thanks for the feedback y'all! I'll add the tests for the updating

ramonh:
An excellent point, @pgaspar. I've added this to the rest of the abilities that require them.

Could you please take another look? 🙆

app/controllers/application_controller.rb
```diff
@@ -24,6 +24,10 @@ class ApplicationController < ActionController::Base
   # See the `pretender` gem.
   impersonates :user
 
+  def signed_in_and_confirmed?
+    signed_in? && current_user.try(:confirmed?)
```
klappradla: Does `current_user` not being `nil` not already tell the same as `signed_in?`

So I'd say
```ruby
def signed_in_and_confirmed?
  current_user&.confirmed?
end
```
Does the same thing but shorter.

app/models/ability.rb
```diff
@@ -23,77 +23,77 @@ def initialize(user)
...
+    can :read, :mailing if signed_in_and_confirmed?(user)
 
     # applications
-    can :create, :application_draft if user.student? && user.application_drafts.in_current_season.none?
+    can :create, :application_draft if user.student? && user.application_drafts.in_current_season.none? && signed_in_and_confirmed?(user)
   end
 
-  def signed_in?(user)
-    user.persisted?
+  def signed_in_and_confirmed?(user)
```
klappradla: Now with reading all the changes, I think my previous suggestion for renaming this method was actually kinda misleading - sorry 🙈 

For the context of the abilities files - "signed in" as in whether or not there's a session is actually not relevant at all. If there's no user signed in, the call to the respective ability would not have a user object...

So I'd say it's fine to just call
```ruby
user.confirmed?
```
In the respective places (same as `user.admin?`)

spec/controllers/users_controller_spec.rb
```diff
@@ -100,6 +100,11 @@
               put :update, params: { id: user.to_param, user: { name: 'Trung Le', email: 'newmail@example.com' } }
             }.to change { ActionMailer::Base.deliveries.count }.by(1)
           end
+
+          it "allows updating the requested user" do
```
klappradla: I think it would make sense to have a dedicated test with a user **that is not confirmed yet**.

spec/controllers/users_controller_spec.rb
```diff
@@ -100,6 +100,11 @@
...
+            expect_any_instance_of(User).to receive(:update_attributes).with(ActionController::Parameters.new({ name: 'Trung Le' }).permit(:name))
```
klappradla: Personal taste, but I see the usage of `any_instance` as a code smell 😉 
In this case, you could have just gone with
```ruby
expect(user).to receive(:update_attributes)
```

However, I also don't see the point in mocking the call here. I think it would make more sense to actually test the outcome, so that updating actually works.
I'd suggest something in the direction of this:
```ruby
subject(:request) { put :update, params: params }
let(:params) { { id: user.id, user: { name: 'whatever' } }

it 'updates the user record' do
  expect { request }.to change(user, :name).from('before').to('whatever)
  expect(response).to redirect_to user
  expect(flash[:notice]).to eq 'Successfully updated your profile'
end
```

So more in the direction of testing that users can really update their profile.

spec/models/ability_spec.rb
```diff
@@ -408,7 +408,7 @@
     let(:team) { create(:team) }
 
     it 'should be logged in' do
```
klappradla: Hm, what is this test supposed to test in the end?

As said in a previous comment, I don't really think that abilities need to have something to do with being signed in or not in our case...

app/views/application/_nav_right.html.slim
```diff
@@ -22,7 +22,7 @@ ul.nav.navbar-nav.navbar-right
       li class=active_if(community_path) = link_to 'Community', community_path
   li class=active_if(page_path('help')) = link_to 'Help', page_path('help')
 
-  - if signed_in?
+  - if signed_in_and_confirmed?
```
pgaspar: I was casually testing the branch with an unconfirmed user and noticed this condition was causing the Sign in button to read "Sign in" and not show me the menu even if I was already signed in. Maybe this check should really be `signed_in?` and return true even if the user is unconfirmed.

spec/models/ability_spec.rb
ramonh: I'm not sure I understand, sorry :(

Should I remove this test?

spec/controllers/users_controller_spec.rb
ramonh: I'm not sure I understood, but given that this is in the context of an unconfirmed user, this is what it does, right?

ramonh:
Thank you both @pgaspar and @klappradla for your feedback!

I've made the changes. What do you think?

app/views/application/_nav_right.html.slim
ramonh: Good catch, thank you!

app/controllers/application_controller.rb
carpodaster: Not sure if that's pedantic (feel free to ignore me), but relying on `signed_in?` == `current_user != nil` is digging into implementation details and lays a trap if we one day decide to e.g. have current_user always return a user-like object (like a `GuestUser` decorator).

spec/models/ability_spec.rb
klappradla: Yes, I'd suggest to remove it 👍

F3PiX:
What do y'all think of this? It is a different approach - not tested, just the broad overview.
We already have : https://github.com/rails-girls-summer-of-code/rgsoc-teams/blob/61bbaef1bc162c8b6f8553e48741544cd2c5211e/app/controllers/omniauth_callbacks_controller.rb#L2-L16
where we could add a `!user.confirmed?` to the conditions in line 6.

That way, all unconfirmed users go right to their profile page (if the grace period is set to 0 days.)

Then, we can do the authentication outside of `Ability`, by adding Devise's `authenticate_user!` to relevant controllers, to check if a user is signed_in. 
Big plus in my book: that means that we can separate Devise's authentication and access to the app, from Cancancan's authorisation and access to perform actions on things. 

As I said, this is a general idea, not tested yet. Is it worth exploring?

ramonh:
Ooh! Good idea, @F3PiX. 

My only question is, if we add this condition to the app, can we still restrict access to confirmed users?

Maybe I didn't understand how this should work...

spec/models/ability_spec.rb
ramonh: It's gone! Thanks @klappradla

F3PiX:
It works out that way. Unconfirmed users don't have access at all for actions that needs signing in, except for their profile. That is the authentication part, which manages access for _a_ user.  So that the authorisation can manage abilities _between_ users: some users can do this, other can do that.
Basically, it is not about changing the requirements, but separating concerns.

ramonh:
Thank you! Apologies if I missed something, so you would suggest replacing this PR with this call? Or adding it on top?

klappradla:
Sorry for coming back to this so late 🙈 

Your idea sounds / looks good @F3PiX! 

In terms of replacing things @ramonh: it looks a bit like `cancan` / the `Abilities` take care of authorization **and** authentication at the moment. The latter is not necessary, we have `Devise` for this.

So even though it was part of the `Abilities` file already, checking whether or not a user is `signed_in` in my opinion does not belong there. This check will be done by the controller even before asking for a user's abilities.

What do you think?

app/models/ability.rb
klappradla: A bit late, but I'm not sure if I understand your comment @carpodaster? What does the `Abilities` file have to do with Devise's magic?  
To me this looks more like an accidental mixing in things that Devise already does into role management... I for instance don't quite understand why `signed_in` is even a question within authorization - it's a Devise concern - and setting it equal to `persisted` is also wrong (persisted user can also be signed out...)

Maybe this thing just has the wrong name and we should simply use `user.persisted?` everywhere within the abilities where it makes sense? What do you think @carpodaster & @ramonh?

app/controllers/supervisors/base_controller.rb
```diff
@@ -8,7 +8,7 @@ class BaseController < ApplicationController
     protected
 
     def must_be_supervisor
-      unless signed_in? && current_user.roles.includes?('supervisor')
+      unless signed_in_and_confirmed? && current_user.roles.includes?('supervisor')
```
klappradla: Only vaguely related to the changes here, but thinking about @F3PiX idea on solving this problem, this line actually looks a bit weird to me: why should non-signed-in users get the same redirect as signed in users with the wrong role? I'd even say non-signed-in users shouldn't even know that there are such things as supervisors 😉 

Using the normal `authorize_user!` before action and then stripping this down to only
```ruby
def must_be_supervisor
  unless current_user.roles.includes?('supervisor')
    redirect_to root_path, alert: 'Sorry, the dashboard is for supervisors only'
end
```

Is this what you were referring to @F3PiX? It looks more correct in terms of separating concerns in my opinon.

spec/controllers/users_controller_spec.rb
```diff
@@ -100,6 +102,13 @@
               put :update, params: { id: user.to_param, user: { name: 'Trung Le', email: 'newmail@example.com' } }
             }.to change { ActionMailer::Base.deliveries.count }.by(1)
           end
+
+          it 'updates the user record' do
+            expect { request }.to change {user.reload.name}.from('before').to('after')
+            request
```
klappradla: This line is obsolete. We're already doing the request in the line above.

app/controllers/supervisors/base_controller.rb
F3PiX: 👍 Yes, that is exactly what I mean. 🎯  
`signed_in?` is obsolete because a `current_user` _is_ the signed_in user. 
And I think you mean to add the `before_filter: authenticate_user! `

ramonh:
Thanks for the feedback everyone! The fixes should be up already. So if I understood correctly, we'll reexamine the auth separations in a separate discussion, right?

F3PiX:
Re: approach and this PR:  
I would recommend to:
- revert the changes to the abilities, and open a new issue for removing all (existing) authentication logic from the Ability class
- implement the new approach in https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/989#issuecomment-387541425 in this PR

Because if it works, we get the confirm status covered at the start of the process, i.e at GH sign in, where it should be, it will be easier to identify and reason about other existing issues regarding the registration process.

ramonh:
Thanks for your feedback, @F3PiX!

I've done as suggested.

F3PiX:
Great @ramonh! The remaining logic with the 'signed_in_and_confirmed'  should be redundant now. Going through the oauth controller comes _instead_ of the `signed_in/confirmed checks you added.
And then we'll need to add the `before_filter :authenticate_user!`  in the relevant controllers.
So then we can see properly if 'the oauth way' passes the specs.

ramonh:
Thanks @klappradla! I'm not quite done yet, will let you know when that's done

app/controllers/application_drafts_controller.rb
```diff
@@ -134,7 +134,7 @@ def current_team
   end
 
   def open_draft
-    current_student.current_draft if signed_in?
+    current_student.current_draft if signed?
```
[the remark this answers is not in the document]
ramonh: Arg, typo! Sorry 💀

F3PiX:
@pgaspar Thanks so much! Probably be one of the other things we need to fix.

To determine which one, could you quickly check something for me? 
While unconfirmed, can you do in the console:
```
pedro = <Find your account>
pedro.unconfirmed_email
```
and if that returns an email address:
``` 
pedro.confirmed?
````
And post the outcome here?
(Because at the moment they can both be true. )
Also, it would be great if you could give one example of those things that you can d  unconfirmed. 

FYI: `authenticate_user!` checks both if the user is signed in and if the user is confirmed. So if we have the `authenticate_user!` at the right places, it doesn't help to add an extra check on the confirmation status.

**Comment by F3PiX** [comment]
> Another thing is that right now the edit page we redirect to has ?welcome=true which seems to hide the "You need to confirm your account" yellow banner, which might be confusing for the user. I don't know what the best approach is here, but maybe we should redirect to a very simple form explaining you need to confirm your email, which would also allow them to edit the email. This would be a simpler single-field form, not the standard edit form.
Nice, love that plan. Let's take it to a separate issue, because IICR it clashes with orga requirements. 

For now, it looks very helpful to set `welcome: false`.

File: app/controllers/application_drafts_controller.rb [hunk 12]
```diff
@@ -2,7 +2,7 @@
 class ApplicationDraftsController < ApplicationController
 
   before_action :checktime, only: [:new, :create, :update]
-  before_action :sign_in_required
+  before_action :authenticate_user!
```
F3PiX: 👏

pgaspar:
@F3PiX sure thing :+1:

`unconfirmed_email` is returning `nil` and `confirmed?` is returning `true`. But the thing is that the way I am making myself unconfirmed was by setting my user's `confirmed_at` to `nil`, so I'm not sure `unconfirmed_email` being nil is what happens for regular users (I can try to mimic that scenario though). Oh, and `email` is my personal email (not nil).

> FYI: authenticate_user! checks both if the user is signed in and if the user is confirmed. So if we have the authenticate_user! at the right places, it doesn't help to add an extra check on the confirmation status.

I was wondering about `authenticate_user!` myself earlier today and I assumed that could be the case, but it doesn't seem to be checking for confirmed users with our current setup. At least I tried a few controllers where we now have `authenticate_user!` calls (for example `CommentsController` and even `UsersController`) and I wasn't redirected by Devise.

I suspect the confirmation check only kicks in based on the configuration for `allow_unconfirmed_access_for`. Right now we have it at `nil`, which means we never block unconfirmed users from accessing the app. I tried it with `0.days` and wasn't even able to sign in, the result being:

screen shot 2018-05-12 at 21 08 02

F3PiX:
@pgaspar 🙇 thanks. Did you logged out and logged back in after setting yourself unconfirmed in the console (and confirmation period > 0)? Because currently, we only solve the 'edit your profile first' issue on log in. 

Re: the UsersController: that is an authorisation problem. I didn't have time to create an issue for it, but I am aware of that. I'll have to look into the CommentsController, that seems to be different, and probably a different authorisation problem.  
Also

pgaspar:
> @pgaspar 🙇 thanks. Did you logged out and logged back in after setting yourself unconfirmed in the console (and confirmation period > 0)? Because currently, we only solve the 'edit your profile first' issue on log in.

I did, yes.

The most important problem I'm seeing now is that after signing in as an unconfirmed user and being redirected to the edit profile page, I can still do whatever I want using the links in the menu since `authenticate_user!` doesn't seem to be taking into account if I'm confirmed or not.

Here's a small example of what I could do after accessing the projects list in the Orga section using my user menu from the edit user page, right after sign in:

untitled

As you can see by the yellow banner, I'm unconfirmed and can still comment on a project, even when the CommentsController has `before_action :authenticate_user!`. The gif is a bit fast, but in the end the comment was submitted and created. I double checked if I was indeed hitting the right controller, and it seems so:

```
Started POST "/comments" for 127.0.0.1 at 2018-05-12 23:14:15 +0100
Processing by CommentsController#create as HTML
```

If `authenticate_user!` cared if I was confirmed, it would have redirected me with the "You have to confirm your account before continuing." notice before the `create` action executed, right?

> Re: the UsersController: that is an authorisation problem.

About UsersController, as above, if `authenticate_user!` redirected unconfirmed users I wouldn't even be able to see my user's edit form, since right now we have a `before_action :authenticate_user!` on UsersController.

If others have some time to try it for themselves, it's easy to make your user unconfirmed. Just `user.update(confirmed_at: nil)` on the rails console 👍 🙇

F3PiX:
🤣 It's getting better all the time. 
I opened an issue to continue the discussion, so we can decide together on how to tackle all the related things. 
In the approach I proposed in this PR,  I intended to _only_ solve that a user who didn't get an email, is able to update his email address.
@pgaspar You are totally right that we can do that only if an unconfirmed user does _not_ have access to restricted areas. I'll push a PR today to provide a temp workaround for that. Sorry, I was too focussed on the email access. Give me a minute.

klappradla:
Hey @ramonh 👋 
What's the state of this? Any chance we can get this back up to speed and merged?

ramonh:
Oh thanks @klappradla I'll get that up and running!

ramonh:
Should be good to go! As discussed, I've removed the changes to make this exclusively to remove the confirmation period.

Not sure this is ready to be merged, however, as I'm not sure we still need to restrict access to unconfirmed users from pages other than editing their profile

klappradla:
Well, restricting access is a different topic, so I guess it would be fine to merge this one an take care of the restrictions in a different PR. Can you confirm @F3PiX ?

ramonh:
That makes sense to me!

F3PiX:
Confirm! I think that all the authorised access (= access for roles) is covered by the latest commits with authorities, but I didn't solve yet that people can ignore the Edit Profile page and proceed to different pages (access by confirmation state).

ramonh:
That's super cool. Thank you for confirming and for all your help! Time to go ahead and merge.

Issue #991, opened 2018-05-13
User Registration Overhaul: break down


F3PiX:
Continue the discussion in #989.  
My proposal would be to first clean up the most obvious things, start building feature tests that cover all the cases/user states and user roles. Then identify if there are underlying problems left. 

1) We need to give unconfirmed users access to update their email address, while denying access to restricted areas. That should be solved with point 2) and 3). 

2) Unconfirmed users having access to restricted pages.  We can't set `allow_unconfirmed_access_for` to 0 days YET, but I hope we can when we reach the end of this list in this issue. 
- [ ] TODO Quick fix /temporary solution to deny access for unconfirmed users. I'll push a PR today, so we can build #989 onto that. 

3) Users should be able to update their email address, even those who did not receive a confirmation email. That (and only that) is solved in #989 
 It means that we can't set `allow_unconfirmed_access_for` to 0 days YET.   

4) Separate the authentication from the authorisation. 
- [ ] TODO Update `abilities`, removing the authentication parts. #990 

5) Identify and solve omissions in the abilities. (UsersController, CommentsController) 
- [ ] TODO UsersController
I'll push a PR that fixes the ability settings on the Users Controller. So that we can double check if that solves the 'unconfirmed access' for that controller, with the new approach in #989 .
- [ ] TODO CommentsController
It looks like there is something missing in the authorising part of the polymorphic relations. Maybe best to see if the fixing the UsersController works before we tackle the comments.
- [ ] (Please add other issues here)

6) Write feature tests 
- [ ] TODO: add feature tests for all the user paths (+ the correct messages) 

7) Apart from all the Ability stuff, there is an underlying thingy with Devise and GH sign in and users created in the Teams form. Maybe things will be easier to maintain if we go Devise all the way.  OTOH, it could be that a little cleaning up, polishing and dusting off is all that is needed. Hard to tell at this point. 

8) @pgaspar suggestion: Change the flow for unconfirmed users without email address,  

> maybe we should redirect to a very simple form explaining you need to confirm your email, which would also allow them to edit the email. This would be a simpler single-field form, not the standard edit form.

Looks like a great plan. I'd prefer not to start with the latter, do the clean up first. 

Is this a plan? Are steps missing?

F3PiX:
@pgaspar @klappradla @ramonh @carpodaster Your opinions please 🙇

ramonh:
I think this is a good plan. Thank you @F3PiX.

[...]

PR #992, opened 2018-05-13
Temp Fix to restrict access for unconfirmed users

F3PiX:
Related issue #991 (Break Down step 2)

We need to restrict access for unconfirmed users but let them update their email address nevertheless. 
- Changed Ability class to catch unconfirmed users.
- Added a 'confirmed user' context for controller tests
- Updated the controller tests that failed; left others alone
- Extra: added first feature test, for guest user. 
Note: I would appreciate help with the feature tests.  This is a first attempt.
 

Follow up issues:
- [ ] check if we need the `can :read, :feed_entry`, or else remove its `can?` check in views. (Public activities shouldn't be restricted at all.)
- [ ] check controller tests and use either 'confirmed user' or 'unconfirmed user' contexts
- [ ] Fix pending specs and todo's in ability_spec

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -108,7 +108,6 @@ class Ability
       cannot :create, Project if !user.confirmed?
 
       # activities
-      # can :read, :feed_entry
       can :read, :mailing if signed_in?(user)
 
       # applications
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -1,6 +1,4 @@
 # frozen_string_literal: true
-# See the wiki for details:
-# https://github.com/ryanb/cancan/wiki/Defining-Abilities
 
 class Ability
   include CanCan::Ability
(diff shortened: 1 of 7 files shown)


spec/models/ability_spec.rb
```diff
@@ -69,6 +71,7 @@
               allow(user).to receive(:confirmed?).and_return(false)
             end
             it 'allows to see not hidden email address' do
+              pending "Fails. Unconfirmed user has no access"
```
F3PiX: ??? This spec is failing, but I am not sure if we even want it to pass for this conditions?

F3PiX:
@pgaspar @klappradla @ramonh Can you take a look please?

spec/models/ability_spec.rb
ramonh: Hmm. You're right. All an unconfirmed user should, in theory, be able to do is edit their email address.

app/models/ability.rb
```diff
@@ -10,102 +8,110 @@ def initialize(user)
 
     alias_action :create, :read, :update, :destroy, to: :crud
 
-    can :crud, User, id: user.id
-    can :crud, User if user.admin?
+    can :read, User
+    can :update, User, id: user.id
```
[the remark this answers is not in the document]
ramonh: I think the failing spec comes from here. Is there a reason we're removing admin access from `crud`'ing users?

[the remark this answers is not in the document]
F3PiX: Admin can still crud User, it has moved to line 38.

app/models/ability.rb
```diff
@@ -10,102 +8,110 @@ def initialize(user)
...
-  end
+      # todo add mailing controller and view for users in their namespace, where applicable
+      can :read, Mailing do |mailing|
+        mailing.recipient? user
+      end
 
-  def on_team_for_season?(user, season)
-    season && user.roles.student.joins(:team).pluck(:season_id).include?(season.id)
-  end
+      can :crud, :comments if user.admin?
```
pgaspar: This is not related to this PR specifically, and we should probably open a separate issue for this: it seems we're not testing this ability in the project comments (or anywhere I could find).

We used to have this in the teams show page, but it was removed meanwhile: https://github.com/rails-girls-summer-of-code/rgsoc-teams/blob/4fc8dfdec20a377fd3a1a500e2399aac7cea48d5/app/views/teams/show.html.slim#L17-L22

I've not used CanCanCan a lot in the past.. do we usually use the [controller helpers](https://github.com/CanCanCommunity/cancancan#3-controller-helpers) or just view helpers?

Should I create the issue or do we want to fix it in this PR?

pgaspar:
I like these changes 👍 I didn't do a lot of testing, but I think the changes make sense so far!

I'm not sure how thoroughly we use these abilities though, there seem to be some that are not checked anywhere - which means unconfirmed users can do _some_ things (I didn't test much other than the comments, but that's one that users can use even before confirmation).

Not sure if we should make an effort to see if thee abilities are used or not (in a separate issue, I assume) or if it's too much trouble.

app/models/ability.rb
F3PiX: We use both helpers ( PR #993 is all about the controller helpers :-) )
If it's okay with you, I'd prefer to keep the PR's as clean as possible. And collect issues for every step layed out in the Overhaul: Breakdown (#991) . So we can use that as a reference all the way.
The `crud comments issue` will be part of the CommentsController Overhaul (listed in nr 5 in 991).  Or afterwards, in the restructuring of the Ability. I'll add it. Thanks!!

pgaspar: Of course! Thank you for your patience - I'm still figuring out how we structure things! 😊

app/models/ability.rb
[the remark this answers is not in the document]
ramonh: My apologies! Dunno how I missed that...

F3PiX:
I don't have authority in CodeClimate to set the warning to 'wontfix'. Can someone disable the warning please? (@ramonh ?)

ramonh:
Oh! Sure! I thought it might've been a comment. I'll take a look, as this PR otherwise looks solid to me. Thanks @F3PiX

spec/models/ability_spec.rb
F3PiX: How about we keep it pending for now? I expect that it will solve itself in the upcoming Overhaul steps.

ramonh: Yes, I agree. Is there a way to mark this as a TODO? Or is the `pending` enough, do you reckon?

F3PiX: I am currently collecting all the todo's, and will add them when we have finished the discussion in #991 . Don't want to add to much detail to that.

ramonh: Fair point!

ramonh:
I think I did it 👍 Let's see!

carpodaster:
@F3PiX I marked the CC issue as wontfix

app/models/ability.rb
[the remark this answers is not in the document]
carpodaster: But the admin call is guarded behind a `if user.confirmed?`. I feel we should keep the current behaviour. If anything, we want admins to be able to fix a user's typo'd email address as a response to a support request ticket.

F3PiX:
@carpodaster THANKS! and sorry for the mess.
1) Moved the helper_methods back home. 🙈 
2) the `if user.confirmed?` checks if the admin user herself is confirmed. Which I expected always to be true...?
3) `can :crud, User if user.admin?` should check if an admin can :crud each and every @user. Because we are not passing in a restriction with the user ids. 
(As opposed to a user who can only crud their own record:
`can :crud, User, id: user.id`)
4) I added very verbose expectations, to double check if we are testing what we want to test, and the behaviour we want. I think the user.admin crudding users is covered now?
5) All this reveals two bigger problems: I didn't update the specs properly, my mistake. And the specs are confusing. 
6) My proposal: close this PR for now, first make sure that we can trust the abilities_specs before changing the behaviour. 
WDYT?

F3PiX:
This is covered now by #997 
Admin specs are covered too:
https://github.com/rails-girls-summer-of-code/rgsoc-teams/blob/9d685da7aeed9045bf9d7de40e973bc782d2eb66/spec/models/ability/admin_spec.rb#L14-L24
Closing this.

PR #993, opened 2018-05-13
Fix setting the @user in UsersController

F3PiX:
Related issue: User Registrations Overhaul 
 #991 .5 in Overhaul Break Down
Branch off of #992 

This is an example for the authorisation issues mentioned in # 991 5). 
Setting the @user happened in two before_actions:  both in Cancancan's `load_resource` method and in  `set_user` . The two before_actions were confusing, because some crud actions were called in both of them; so: sometimes with authorization, sometimes without. That caused some weird behaviour in the authorisations.  (Possibly the reason for some of the extra 'signed_in` checks in Ability.)
-  Now the @user is set only once in each controller action. 
-  Specs were flaky, because of a GitHub user name validation error in the Factory. Changed that to a sequence. 

@pgaspar
When logged in as an unconfirmed user, see the difference between the behaviour on the Teams pages compared to the new behaviour on the Users pages. 
This is what I meant here: https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/989#issuecomment-388581753

- [ ] Follow up: Find all the other controllers that have the same ambiguity and solve them.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -1,8 +1,6 @@
 # frozen_string_literal: true
 class UsersController < ApplicationController
-  before_action :set_user, only: [:show, :edit, :update, :destroy, :impersonate]
-
-  load_and_authorize_resource except: [:index, :show, :impersonate, :stop_impersonating]
+  load_and_authorize_resource except: [:impersonate, :stop_impersonating]
 
   def show
   end
@@ -60,6 +58,7 @@ class UsersController < ApplicationController
   end
 
   def impersonate
+    set_user
     impersonate_user(@user)
     redirect_to community_path, notice: "Now impersonating #{@user.name}. You can find the link to stop impersonation on the user menu."
   end
--- a/spec/factories/users.rb
+++ b/spec/factories/users.rb
@@ -1,6 +1,6 @@
 FactoryBot.define do
   factory :user, aliases: [:member] do
-    github_handle { FFaker::InternetSE.user_name_variant_short }
+    sequence(:github_handle) { FFaker::InternetSE.user_name_variant_short }
     name     { FFaker::Name.name }
     email    { FFaker::Internet.email }
     location { FFaker::Address.city }


pgaspar:
Hey @F3PiX!

Back in https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/989#issuecomment-388579791 I was referring specifically to what I expected `before_action :authenticate_user!` to do, since we were trying to figure out if that before action was blocking access to unconfirmed users or not.

That being said, I think I understand your changes in this PR, but I want to double check. Correct me if I'm wrong: index and show were not being authorized and were rendering without any ability check. Other than that, edit, update and destroy were setting `@user` twice because of the two before_actions. Am I right that these are the issue we're fixing here?

I tried checking how User pages (index, show, editing) compare with the Team pages (index, show, editing, creating) when it comes to confirmed and unconfirmed users and I couldn't see differences - Read actions were available for both confirmed and unconfirmed users while Create, Update (and Delete I assume) actions were only available for confirmed users, redirecting unconfirmed users with an authorization error. Both seemed to work similarly in this sense at least from my testing. I'm sorry if I'm missing the point you were trying to make! 😓

F3PiX:
@pgaspar Thanks for checking!! 
First:
> That being said, I think I understand your changes in this PR, but I want to double check. Correct me if I'm wrong: index and show were not being authorized and were rendering without any ability check. Other than that, edit, update and destroy were setting @user twice because of the two before_actions. Am I right that these are the issue we're fixing here?
That is exactly the point 🎯 

That you don't find differences in the UI is because all the extra authorisation calls in Ability.
I'd say: let's finish that first, because it will make all the other changes, including this one, super clear. 

Also, I should have given a bit more background info where to find the  changes. The UI doesn't change (because that was taken care with workarounds in Cancancan. ) (Except o joy, it is sooo much faster!)  
 So, for future reference:
The changed behaviour can be reveiled by adding a `byebug` at the start of the Ability initializer. 
Before this PR:  `/users/yourid/edit` will go multiple times through the authorisation process.   
After: same request will go directly to the edit page, without hitting authorisation - as it should! 
BUT I just checked this in my PR with the make over of Ability. 

TL;DR It is best to set this PR on hold untill after the make over.

PR #994, opened 2018-05-14
Revisit associations pt. 1

klappradla:
As a side-effect of the whole [GDPR party](https://en.wikipedia.org/wiki/General_Data_Protection_Regulation) we want to have a working _chain of associations_ for a user: deleting a user should a) be possible and b) also remove / unpersonalize all their associated data.

For starters I'm having a look at the current setup and document it in the specs. This PR looks at the first few models, documents their associations in the specs and of course sneaks in some drive-by-fixes here and there 🚜 

As a first **effect** we can also already delete `Activities` where the `team` no longer exists - we have a few of them in the production DB.

### DIFF
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -9,19 +9,9 @@ class Activity < ApplicationRecord
 
   delegate :students, to: :team
 
-  class << self
-    def with_kind(kind)
-      where(kind: kind)
-    end
-
-    def by_team(team_id)
-      where(team_id: team_id)
-    end
-
-    def ordered
-      order('published_at DESC, id DESC')
-    end
-  end
+  scope :with_kind, ->(kind) { where(kind: kind) }
+  scope :by_team, ->(team_id) { where(team_id: team_id) }
+  scope :ordered, -> { order('published_at DESC, id DESC') }
 
   def to_param
     "#{id}-#{title.to_s.parameterize}"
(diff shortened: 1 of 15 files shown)


ramonh:
Looks overall great to me, thank you @klappradla ! 👍 🎉

F3PiX:
:wave: 
That was a bit too quick for me. I have some questions.

ramonh:
Uh oh, sorry @F3PiX. I'll try and be less quick in the future

F3PiX:
@klappradla Super cool!
- With the GPDR in mind, do we still want the user to be able to destroy their own record? I mean: can we guarantee that we can clean up properly if the user account is destroyed by the user herself?
- I just invented a GPDR Rpsec matcher 😉     
I was thinking: just listing the associations doesn't guarantee we catched them all, right? It depends if we added them manually.
Isn't it neat if we check if we catched them all?
` expect(inflections_count).to eq (Model.inflections.keys.count)`
I also found a way to count the validation specs.

klappradla:
> can we guarantee that we can clean up properly if the user account is destroyed by the user herself?

☝️ that's the goal, yes _(needs a bit more work of course)_

> just listing the associations doesn't guarantee we catched them all, right? It depends if we added them manually.

👆 I fear I don't get that 🙈

F3PiX:
We want to 
> document their associations in the specs

And you did that by adding a validation for each of them. 
Now, if someone adds a new association, how can we be sure that it is added to the specs, so the documentation will always be acurate? By checking those validations against all the associations that exist on that model. 
` User.reflections.keys # => ["roles", "teams", "application_drafts", "applications", "todos"] `
Check! All User associations are documented; there is a validation for each of them.

PR #996, opened 2018-05-14
Fix code climate test coverage report

carpodaster:
idk if coverage reports in CC broke at some point or never worked at all. This should get coverage reports (back?) into CC.

**Edit**:

screen shot 2018-05-14 at 23 16 15

Doesn't show up in CC, yet. I reckon it won't before we merge this to master 🤔


ramonh:
Sounds good, looks good! 👍 Thank you @carpodaster

PR #997, opened 2018-05-15
Rearrange, not change the abilities in Ability

F3PiX:
Update:
Rearranged, not changed the abilities in Ability. 

Following this principles for Defining Abilities:
- [Arranging abilities per role, not per model](https://github.com/CanCanCommunity/cancancan/wiki/Defining-Abilities)
- [Give permissions, dont take them away](https://github.com/CanCanCommunity/cancancan/wiki/Defining-Abilities%3A-Best-Practices#give-permissions-dont-take-them-away)

The reason: it is easier to follow what a user can or cannot do. Less error prone and easier to reason about.
It is also 1000 times easier to write and read specs.

Follow up:
- [ ] In the console, I assigned my team_less account a student_role.  I added a conf. Is that even possible IRL? When hitting Conferences#create, it redirects to <my-team>-path. => UrlGenerationError. Conference not created. Create issue or irrelevant?
 
Questions about current abilities/implementation:

- [ ] A student can destroy their own team, which make sense. But what about an accepted team, mid-season? 
- [ ] A supervisor can read some extra details of the users; AFAIK it concerns email_addresses and team's blogpost_info. That makes sense for their own team, but not for users on other teams, right? It needs further investigation. ->  issue #1001 
- [ ] Note: Some 'ability-ish' feature are managed by methods in application_helper.rb (:can_see_private_info, :can_only_review_private-info.).


F3PiX:
@carpodaster Trying to unmess all things Ability from the previous PR. What do you think of this new setup? I have also made a start with the specs, but I don't want to throw too much stuf. LMK if you want to see them?

app/models/ability.rb
```diff
@@ -10,18 +8,50 @@ def initialize(user)
...
     alias_action :create, :read, :update, :destroy, to: :crud
 
+    # unconfirmed user
+    can :read, User
+    can :update, User, id: user.id
+    can :resend_confirmation_instruction, User, id: user.id
+    can :read_email, User, hide_email: false # view helper
+    can :read, Team
+    can :read, Project
+    can :read, :feed_entry
```
klappradla: First off: forgive my questions, I'm not familiar with CanCan - I'm always using Pundit... 

What does this syntax here mean? `:feed_entry` is a symbol, not a class / model... ❓

F3PiX: I <3 your questions. 
It is an :kind of an Activity, and used in lib/feed/item. I left it in here for now, because I am not changing abilities until they are thoroughly tested 😇

F3PiX:
Thanks @klappradla ! Let's not merge yet, I'll go rewrite the specs first. [Here](https://gist.github.com/F3PiX/84819d14af39e0d1003f2eb118a49736) is a taste of what they are going to look like. So excited about this! Look how neat it is going to be!

app/models/ability.rb
```diff
@@ -10,18 +8,50 @@ def initialize(user)
...
     can :resend_confirmation_instruction, User, id: user.id
-    can :resend_confirmation_instruction, User if user.admin?
+    can :read, :mailing  if signed_in?(user)
+    can :read, Mailing do |mailing|
+      mailing.recipient? user
+    end
+    can :create, Project if user.confirmed?
+
+    # current_student
+    can :crud, Conference if user.current_student?
```
pgaspar: Would we need to add `&& user.confirmed?` to all the checks below to make sure they are only available to confirmed users?

I like this organization - it seems clearer to me. Mixing "confirmed" and "unconfirmed" with user roles seems a bit odd to me, though.

Update: ok, things make a bit more sense when I read the "Give permissions, don't take them away" link you posted. Doing things like `return unless user.confirmed?` right before line 20 ensures that everything below is only for confirmed users.

F3PiX: Yes, @pgaspar , instead of all the `confirmed?` checks, we will have one return statement. 😍 
I agree it would be even neater to have no authentication check in here at all, but we'll probably going to need that until we went through everything in #991. 🍀 
Just a few steps away 🤣

F3PiX:
Wow! That was 💥 
I TRIED to make this as easy for you reviewers as I could. 😇 :
- The last commit is a clean file with some questions. The second to last has all kinds of annotations. They explain some of the cancancan quirks, so that should answer some of the questions you may have.
- There is more work to do, but I'd thought this makes for a reasonable pit stop. 
- Note that the abilities are not complete; I didn't add things, only rearranged what we already had. And all of those are covered by tests now. 🎉 
Questions about the current abilities, and if they should be changed, are listed in the PR description ^.
- There is some noise in the specs. I'd like to stub things, so that we are only testing the ability. Help is much appreciated!
 
Let me know if I can do anything else to make reviewing less hard. 
@klappradla @carpodaster @pgaspar @ramonh

spec/factories/users.rb
```diff
@@ -1,6 +1,6 @@
 FactoryBot.define do
   factory :user, aliases: [:member] do
-    github_handle { FFaker::InternetSE.user_name_variant_short }
+    github_handle { FFaker::InternetSE.unique.user_name_variant_short }
```
F3PiX: This fixes flickering specs because of failing uniqueness validation on gh handle.

spec/models/ability_spec.rb
```diff
@@ -1,437 +1,240 @@
...
 
-      describe 'she/he is not allowed to CRUD on someone else account' do
-        let(:other_user) { create(:user) }
-        it { expect(ability).not_to be_able_to(:show, other_user) }
+    it_behaves_like "same as logged in user"
+    it_behaves_like "can create a Project"
+    it_behaves_like "can see mailings list too"
+    it_behaves_like "can read mailings sent to them"
+    it_behaves_like "can comment now" # not implemented; false positives; needs work
+    it { expect(subject).to be_able_to([:join, :create], Team) }
```
klappradla: Cosmetic, but it's not necessary to write `subject` when using the short hand syntax. You can simply write:
```ruby
it { is_expected.to be_able_to %i(join create), Team }
```

Same for all further appearance 😉

spec/models/ability_spec.rb
```diff
@@ -1,437 +1,240 @@
...
-              expect(ability).to be_able_to(:read_email, user)
-            end
-          end
+      describe "Student" do
+        let(:student_team) { build(:team, :in_current_season) }
+        let(:second_team) { build(:team, :in_current_season) }
+        before do
+          allow(user).to receive(:current_student?).and_return(true)
         end
+        # TODO Question: is this ^ stub correct to test the ability only?
```
klappradla: Given that you know what `current_student?` actually does and that it's tested elsewhere: then yes ✅

spec/models/ability_spec.rb
```diff
@@ -1,437 +1,240 @@
...
-              allow(user).to receive(:admin?).and_return(false)
-              allow(user).to receive(:confirmed?).and_return(false)
-            end
-            it 'disallows to see not hidden email address' do
-              other_user.hide_email = false
-              expect(ability).not_to be_able_to(:read_email, other_user)
-            end
-          end
+        it 'cannot create a second team in current_season' do
+          pending 'fails; it is tested in team_spec.rb:33, and that passes'
```
klappradla: Well, I fear you have to actually write the teams to the DB to have a test like this...

spec/models/ability_spec.rb
```diff
@@ -1,437 +1,240 @@
...
+      describe "Supervisor" do
         let(:other_user) { create(:user) }
+        let(:other_team_user) { build_stubbed(:user) }
 
-        describe 'a user can only resend her/his own confirmation' do
-          it { expect(ability).to be_able_to(:resend_confirmation_instruction, user) }
-          it { expect(ability).not_to be_able_to(:resend_confirmation_instruction, other_user) }
+        before do
+          allow(user).to receive(:supervisor?).and_return(true)
+          allow(ability).to receive(:supervises?) { false } # Rspec complained: "stub default value first"
```
klappradla: I sorta understand why RSpec does this 😉 

I'd personally never use the `receive(:thing).with` when defining a mock (with `allow`). The matcher / syntax makes way more sense for **expecting message calls** in my opinion.
So within the test you'd have something like
```ruby
expect(model).to receive(:thing?).with(other_thing).and_return(false)
```
☝️ testing the "contract" so to say.

spec/models/ability_spec.rb
```diff
@@ -1,437 +1,240 @@
...
-  context 'ability' do
-    context 'when a user is connected' do
-      let!(:user) { create(:user) }
-      let!(:team) { create(:team) }
+RSpec.describe Ability, type: :model do
+  subject(:ability) { Ability.new(user) }
+  let(:user){ nil }
+  let(:other_user) { create(:user) }
+
+  describe "Guest User" do
```
klappradla: Personal preference and only a suggestion, bud I'd make all of these role-specific `describes` a `context`, such as
> context 'when the user is a guest' do

spec/support/shared_examples/abilities.rb
```diff
@@ -0,0 +1,95 @@
...
+
+
+# DETAILS, in order of appearance
+
+PUBLIC_INDEX_PAGES = [Activity, User, Team, Project, Conference].freeze
+
+# details for unrestricted
+shared_examples 'can view public pages' do
+  PUBLIC_INDEX_PAGES.each do |page|
+    it { expect(subject).to be_able_to(:read, page) }
```
klappradla: Same here: you can ditch the `subject` when using this syntax:
```ruby
it { is_expected.to be_able_to(:read, page) }
```

Same for the rest of this file of course.

spec/support/shared_examples/abilities.rb
```diff
@@ -0,0 +1,95 @@
+# collections of shared examples
```
klappradla: Using shared examples is definitely a valid approach ✔️ 
I'm personally not a huge fan of using them in this amount and spread across multiple files though.

Well, comes down to personal preference, but I'm just quickly dumping my brain here: From my experience, they are usually only maintainable if being used only a few times and defined in the same file - it's the only way how a human brain can still keep track of the implicit contracts between the files. Plus: updating the feature and respectively the specs behind it gets super complicated, especially when nesting them like this.
They don't even run faster 😉 

Alternative approaches here could for instance be:
- split the abilities specs in one file per role _(prefer that tbh)_
- test based on the requested entity rather then the role _(and maybe again split into muliple files)_

spec/models/ability_spec.rb
```diff
@@ -1,437 +1,240 @@
...
+    it_behaves_like "can view public pages"
+    it_behaves_like "can not modify things on public pages"
+    it_behaves_like "can not create new user"
+    it_behaves_like "can not comment"
+    it_behaves_like "has no access to other user's accounts"
+    it_behaves_like "can not read role restricted or owner restricted pages"
+    it "does not have an account" do
```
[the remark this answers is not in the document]
F3PiX: It is a pro memory. If you have to ask, it is not working, I'll delete it.

spec/models/ability_spec.rb
F3PiX: I had to look up the difference, but I still don't understand what it is happening. 
If I change the line to 
  ```expect(ability).to receive(:supervises?).with(other_user, user) { true }```
it fails with `received 0 times`
Do you understand what I am missing here?

spec/support/shared_examples/abilities.rb
F3PiX: I totally agree that is not an ideal solution. It crossed my mind to split the abilities themselves in different files, but your suggestion to split only the specs in different files sounds better. I'd rather save that for a next step, if that is okay. 

Moreover, the shared examples are mostly dealing with 'authentication stages', not roles. If everything works out with the Big Overhaul, we don't have to have that in Ability anymore... 🍀

spec/support/shared_examples/abilities.rb
```diff
@@ -0,0 +1,95 @@
...
+
+# logged in, unconfirmed:
+shared_examples "same as guest user" do
+  it_behaves_like "can view public pages"
+  it_behaves_like "can not modify things on public pages"
```
pgaspar: Small stylistic suggestion: maybe use `cannot` instead of `can not`? I think it makes it easier to parse out negatives and matches CanCanCan's `cannot`.

spec/models/ability_spec.rb
pgaspar: That ^ and we have this for confirmed users on line 27:

```ruby
can [:join, :create], Team
```

By the way, instead of persisting the teams in the DB maybe you can get away with stubbing the `user.teams` call.

spec/models/ability_spec.rb
klappradla: Oh sorry, I did not mean you should change it!

Expecting a method call only makes sense (and will only pass) within an example. So if your test want to check whether or not the specific message is passed to this object but you don't care about the inner logic and want to stub it, it may look like this:
```ruby
it 'is able to do something' do
  expect(ability).to receive(:supervise?).with(other_user, user).and_return(true)
  expect(ability).to be_able_to(:read, :users_info, other_user)
end
```

spec/support/shared_examples/abilities.rb
klappradla: Sure!

As said, your approach is totally valid 👍 

I'd personally have done it differently and just wanted to explain this in case you were actually looking for alternative approaches or did not think about certain gotchas that come with it.

So all good to stay with this.

spec/models/ability_spec.rb
F3PiX: OOOOH... and you actually said that: "within an example". Sorry.
NOW I understand the fail message "provide a stub first" 💃

spec/support/shared_examples/abilities.rb
F3PiX: I hate that syntax with a passion. 😅 There is no performance or functional difference, right?

spec/support/shared_examples/abilities.rb
F3PiX: Hmmm They are not always used the same as cancancan's `cannot`'s. That made it more confusing, imo, so I left them as is.

F3PiX:
@klappradla @pgaspar Thanks so much for your reviews! 🙇 Here is the next step.

- The problems with the specs turned out to be covering bigger problems with the implementation of cancancan in other places. With the Main Goal in mind - the Registration Overhaul - I reverted some of the changes to the previous state; they are working, and not needed for fixing the registration process. I did add specs for the desired behaviour though.
- With splitting the ability_spec file in separate spec files per role, it looks promising, but I don't know what to do now with the shared_examples. Should I just copy paste the examples into each file, and accept all the duplication? Or is there a better way?
- Bonus: the first registrations' feature test, for the guest user. 

Awaiting your reactions!

F3PiX:
Hi Ramon, we merge this one first, as soon as it is approved. And carry on from there.

F3PiX:
On hold! I found a way to remove one level of nesting in the shared_examples (inspired by Max' request tests) and I am adding issues for the TODO's.

spec/controllers/teams_controller_spec.rb
```diff
@@ -12,11 +12,11 @@
 
   let(:valid_attributes) { build(:team).attributes.merge(roles_attributes: [{ name: 'coach', github_handle: 'tobias' }]) }
 
-  before do
-    user.roles.create(name: 'student', team: team)
-  end
-
   describe "GET index" do
```
F3PiX: Now that the new ability rule prevents students from creating a second team, the specs can't assign two teams either. Move the one big `before` into the contexts.

spec/features/users/guest_user_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+    end
+
+    context 'in season' do
+      before do
+        Timecop.travel(summer_season)
+      end
+      after { Timecop.return }
+
+      it "can view the current season's accepted and selected projects" do
+        # project not visible on page. why?
```
F3PiX: Related to #1002?

app/controllers/application_controller.rb
```diff
@@ -33,7 +33,7 @@ def after_sign_in_path_for(user)
   end
 
   rescue_from CanCan::AccessDenied do |exception|
-    redirect_to root_url, alert: exception.message
+    redirect_back(fallback_location: root_path, alert: exception.message)
   end
```
F3PiX: 🏎 The 'not authorised' method will now show up on the page where the user tried to do something forbidden. We couldn't `redirect :back` before, because it would mess up if there wasn't a referer present.

F3PiX:
- Removed the nesting in the shared_examples
- The common examples of confirmed users are not repeated in every confirmed role anymore. 
- Solved issue #1003, because it was easier than to leave it alone. 
- Added the new Rails 5 redirect_back method, so that unauthorised users are getting the alert on the page where the forbidden action happened.

@klappradla @pgaspar

spec/features/users/guest_user_spec.rb
F3PiX: AHA! I had to use a `in_current_season` scope 🌞

spec/support/shared_examples/abilities.rb
klappradla: Hm, I tbh. don't know, but probably not 🤔 

I personally think the syntax reads way more natural than explictly saying "subject". Take for instance
> describe thing - it is expected to be such a thing

Since I haven't yet been on a project where linters would allow you syntax 🚔 🚒  - I'm probably also just used to the implicit one 😂

F3PiX:
Thanks all, for your reviews!! 💟

Issue #979, opened 2018-04-05
Add rake task to strip down production database dump


carpodaster:
**Why?**
We have a staging environment whose free database tier only supports up to 10,000 rows. We occasionally need to copy the current prod DB to staging in order to QA things.

We used to flush the Teams App's DB every year but since the introduction of the `Season` concept, we don't do that anymore. This is why the DB keeps growing.

**What?**
The list below shows how many records each ActiveRecord model has (as of April 2nd).

* The easiest way to solve this would be to delete data for old seasons, but ideally we also *keep some old data*
* The current season should be left untouched
* Accepted teams shouldn't be purged
* Bonus: sanitize the data (user data, team names, application data, comments) so that we may use the DB dump as a developer's snapshot

```ruby
ApplicationRecord
  .descendants
  .map { |klass| { klass.to_s => klass.count } }
  .sort_by(&:values)

# [{"Season"=>4}, {"ConferenceAttendance"=>22}, 
# {"Note"=>31}, {"Mailing"=>40}, {"Conference"=>103}, {"ConferencePreference"=>112}, 
# {"Maintainership"=>153}, {"Source"=>161}, {"Project"=>196}, {"Mentor::Comment"=>229}, 
# {"Application"=>542}, {"ApplicationDraft"=>699}, {"Team"=>782}, {"Todo"=>1158}, 
# {"Rating"=>2067}, {"Comment"=>2507}, {"Role"=>3150}, {"Submission"=>3187}, 
# {"Activity"=>3564}, {"User"=>3661}]
```

klappradla:
Just a thought 🤔 : in regards of what just happened at Travis CI _(and what previously also happened at Gitlab)_ we should probably prevent this from accidentally being run in the `production`environment 😉 

I'd therefore suggest to add a 4th environment: `staging` (don't know why we do not have it already)_ that is basically a clone of `production` and run the staging app in this environment.

carpodaster:
I'm not a fan of extra envs; it's also not advised by Heroku, but I'm also not totally against it :)

I've come across [this Twitter thread](https://twitter.com/schneems/status/981573494403080192) that has some retrospective on AR meta data about the env its DB is configured for. We could test if that actually works.

Or we could abort if there is a `DATABASE_URL` set in dev or test envs.

[...]

PR #1006, opened 2018-05-25
Spring cleaning users

F3PiX:
Related issue #991
 
On the way to separating authentication from authorisation: 
- Straighten out the authorisation in User: now we authorise all user actions, so that the abilities can take care what is shown to who (for example: we want a supervisor to be able to see email addresses of team members in the show view, even if the team member choose to hide their email from regular users). 
- Added a feature test for the sign in process for unconfirmed users.

Also:
- UX: Added a client side validation for the city field, which already has a server side validation
- Database: Added migration to remove users without a github_handle.  
  Why? For accounts imported from GitHUb, the github_id is not nil. So if the id is nil, it is a fake 
  account or an input error. Related issue #979 
  @klappradla I never used a migration for removing data before. I copied one of your examples, can you double double check please?
- FIX:  only confirmed users can read email addresses in users index and show

ALERT 
- In Dev env, running the migration will remove the seeded users from your local database. If you want some of them to persist, give them a fake `github_id` < 100. 
- Run `rails db:migrate` and `rails db:seed` 
  
Background info:
- Cancancan's load_and_authorize_resource will load the resource, unless it is already set. That means: we don't need the separate `before_action: set_resource` for authorised actions.  (Note that my previous description about this behaviour where not 100% accurate 😇 )

Follow up:
- [ ] Update the hint in the form, and tell the user that a supervisor and admin will be able to see email, whether marked hidden or not. Check first if this statement is and should be true.
- [ ] Add message to slack dev channel that all user-seeds will be removed with the latest migration.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -11,13 +11,6 @@ class ApplicationController < ActionController::Base
     session[:redirect_to] = redirect_to if redirect_to.present?
   end
 
-  # workaround fix for cancan on rails4 - https://github.com/ryanb/cancan/issues/835
-  before_action do
-    resource = controller_path.singularize.gsub('/', '_').to_sym
-    method = "#{resource}_params"
-    params[resource] &&= send(method) if respond_to?(method, true)
-  end
-
   before_action :set_timezone
 
   # Allow users to impersonate other uses in a non-production environment.
--- a/app/views/community/index.html.slim
+++ b/app/views/community/index.html.slim
@@ -16,7 +16,8 @@ table.users.table.table-striped.table-bordered.table-condensed.table-hover
       th
       th = sortable :name
       th = sortable :interested_in
-      th = 'Email' if current_user&.confirmed?
+      - if current_user&.confirmed?
+        th = 'Email'
       th = sortable :github
       th = sortable :team
       th = sortable :country
@@ -30,7 +31,8 @@ table.users.table.table-striped.table-bordered.table-condensed.table-hover
         td.image = avatar_url(user, size: 40)
         td = link_to user.name_or_handle, user
         td = user.interested_in.map { |key| User::INTERESTS[key].presence }.compact.join(', ')
-        td = mail_to user.email if can?(:read_email, user)
+        - if current_user&.confirmed?
+          td = mail_to user.email if can?(:read_email, user)
         td = link_to user.github_handle, "https://github.com/#{user.github_handle}"
         td = link_to_user_roles(user)
         td = user.country
--- a/app/views/community/index.html.slim
+++ b/app/views/community/index.html.slim
@@ -16,8 +16,7 @@ table.users.table.table-striped.table-bordered.table-condensed.table-hover
       th
       th = sortable :name
       th = sortable :interested_in
-      - if current_user&.confirmed?
-        th Email
+      th = 'Email' if current_user&.confirmed?
       th = sortable :github
       th = sortable :team
       th = sortable :country
--- a/app/views/community/index.html.slim
+++ b/app/views/community/index.html.slim
@@ -31,7 +31,7 @@ table.users.table.table-striped.table-bordered.table-condensed.table-hover
         td.image = avatar_url(user, size: 40)
         td = link_to user.name_or_handle, user
         td = user.interested_in.map { |key| User::INTERESTS[key].presence }.compact.join(', ')
-        td = mail_to user.email if can?(:read, user.email)
+        td = mail_to user.email if can?(:read_email, user)
         td = link_to user.github_handle, "https://github.com/#{user.github_handle}"
         td = link_to_user_roles(user)
         td = user.country
--- a/app/views/community/index.html.slim
+++ b/app/views/community/index.html.slim
@@ -16,7 +16,7 @@ table.users.table.table-striped.table-bordered.table-condensed.table-hover
       th
       th = sortable :name
       th = sortable :interested_in
-      - if signed_in?
+      - if current_user&.confirmed?
         th Email
       th = sortable :github
       th = sortable :team
@@ -31,8 +31,7 @@ table.users.table.table-striped.table-bordered.table-condensed.table-hover
         td.image = avatar_url(user, size: 40)
         td = link_to user.name_or_handle, user
         td = user.interested_in.map { |key| User::INTERESTS[key].presence }.compact.join(', ')
-        - if signed_in?
-          td = mail_to user.email unless user.hide_email?
+        td = mail_to user.email if can?(:read, user.email)
         td = link_to user.github_handle, "https://github.com/#{user.github_handle}"
         td = link_to_user_roles(user)
         td = user.country
--- a/db/migrate/20180524190818_destroy_non_existing_git_hub_users.rb
+++ b/db/migrate/20180524190818_destroy_non_existing_git_hub_users.rb
@@ -1,0 +1,5 @@
+class DestroyNonExistingGitHubUsers < ActiveRecord::Migration[5.1]
+  def up
+    execute 'DELETE FROM users WHERE github_id IS NULL'
+  end
+end
--- a/db/migrate/20180524190818_destroy_non_existing_git_hub_users.rb
+++ b/db/migrate/20180524190818_destroy_non_existing_git_hub_users.rb
@@ -1,11 +1,8 @@
 class DestroyNonExistingGitHubUsers < ActiveRecord::Migration[5.1]
   def up
     unless ENV['RAILS_ENV'] == 'development'
-      User.find_each do |user|
-        if user.github_id == nil
-          puts "removing user #{user.id} from database"
-          user.destroy
-        end
+      User.where(github_id: nil).destroy_all.each do |user|
+        puts "removed user #{user.id} from database"
       end
     end
   end
--- a/db/migrate/20180524190818_destroy_non_existing_git_hub_users.rb
+++ b/db/migrate/20180524190818_destroy_non_existing_git_hub_users.rb
@@ -1,5 +1,12 @@
 class DestroyNonExistingGitHubUsers < ActiveRecord::Migration[5.1]
   def up
-    execute 'DELETE FROM users WHERE github_id IS NULL'
+    unless ENV['RAILS_ENV'] == 'development'
+      User.find_each do |user|
+        if user.github_id == nil
+          puts "removing user #{user.id} from database"
+          user.destroy
+        end
+      end
+    end
   end
 end
--- a/db/migrate/20180524190818_destroy_non_existing_git_hub_users.rb
+++ b/db/migrate/20180524190818_destroy_non_existing_git_hub_users.rb
@@ -1,6 +1,14 @@
 class DestroyNonExistingGitHubUsers < ActiveRecord::Migration[5.1]
+
   def up
-    unless ENV['RAILS_ENV'] == 'development'
+    return if ENV['RAILS_ENV'] == 'test'
+    if ENV['RAILS_ENV'] == 'development'
+      puts "Updating users with github_id ...."
+      users = User.all.select { |user| user.github_id.nil? }
+      users.each_with_index do |user, index|
+        user.update!(github_id: index + 10)
+      end
+    else
       User.where(github_id: nil).destroy_all.each do |user|
         puts "removed user #{user.id} from database"
       end
--- a/spec/controllers/community_controller_spec.rb
+++ b/spec/controllers/community_controller_spec.rb
@@ -35,8 +35,10 @@ RSpec.describe CommunityController, type: :controller do
         user = create(:student, hide_email: false)
         user_opted_out = create(:user, hide_email: true)
         get :index
-        expect(response.body).to include user.email
-        expect(response.body).not_to include user_opted_out.email
+        # expect(response.body).to include user.email
+        # This expectation ^ fails, but should this be tested in a controller test in
+        # the first place?
+        expect(response.body).not_to include user_opted_out.email # false negative; also passes with hide_email: false
       end
 
       it 'shows user impersonation links when in development' do
--- a/spec/controllers/community_controller_spec.rb
+++ b/spec/controllers/community_controller_spec.rb
@@ -35,9 +35,8 @@ RSpec.describe CommunityController, type: :controller do
         user = create(:student, hide_email: false)
         user_opted_out = create(:user, hide_email: true)
         get :index
-        # expect(response.body).to include user.email
-        # This expectation ^ fails, but should this be tested in a controller test in
-        # the first place?
+        expect(response.body).to include user.email
+        # Should this be tested in a controller test?
         expect(response.body).not_to include user_opted_out.email # false negative; also passes with hide_email: false
       end
 
--- a/spec/controllers/community_controller_spec.rb
+++ b/spec/controllers/community_controller_spec.rb
@@ -36,7 +36,6 @@ RSpec.describe CommunityController, type: :controller do
         user_opted_out = create(:user, hide_email: true)
         get :index
         expect(response.body).to include user.email
-        # Should this be tested in a controller test?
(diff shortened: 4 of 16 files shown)


F3PiX:
This is ready for review 🙇

app/controllers/application_controller.rb
```diff
@@ -11,13 +11,6 @@ class ApplicationController < ActionController::Base
     session[:redirect_to] = redirect_to if redirect_to.present?
   end
 
-  # workaround fix for cancan on rails4 - https://github.com/ryanb/cancan/issues/835
-  before_action do
-    resource = controller_path.singularize.gsub('/', '_').to_sym
-    method = "#{resource}_params"
-    params[resource] &&= send(method) if respond_to?(method, true)
-  end
-
```
ramonh: 🙌

app/views/community/index.html.slim
```diff
@@ -31,8 +31,7 @@ table.users.table.table-striped.table-bordered.table-condensed.table-hover
         td.image = avatar_url(user, size: 40)
         td = link_to user.name_or_handle, user
         td = user.interested_in.map { |key| User::INTERESTS[key].presence }.compact.join(', ')
-        - if signed_in?
-          td = mail_to user.email unless user.hide_email?
+        td = mail_to user.email if can?(:read, user.email)
```
pgaspar: I think there's a bug here: this keeps the `td` even if the `can?` method returns false, which breaks the layout a bit. (Try it with a logged out or unconfirmed user.)

I think you need to change this back to be a 2-line conditional.

app/views/community/index.html.slim
```diff
@@ -16,7 +16,7 @@ table.users.table.table-striped.table-bordered.table-condensed.table-hover
       th
       th = sortable :name
       th = sortable :interested_in
-      - if signed_in?
+      - if current_user&.confirmed?
```
pgaspar: I think we should make this check be the same `can?` check as you have below. What do you think?

db/migrate/20180524190818_destroy_non_existing_git_hub_users.rb
```diff
@@ -0,0 +1,5 @@
+class DestroyNonExistingGitHubUsers < ActiveRecord::Migration[5.1]
+  def up
+    execute 'DELETE FROM users WHERE github_id IS NULL'
```
pgaspar: Some considerations 🤔:

First, I think we may want to prevent this from running on development to avoid the issue you pointed out on the PR description. I worry this may cause unexpected behavior for existing developers who update their local copy of the app.

Wrapping the `execute` statement on a `unless Rails.env.development?` should do it, I think 👍

Second point: do these users have any relationships with other models in the database? If they do, maybe we should consider doing this using Ruby so the destroy callbacks run. `User.where(github_id: nil).destroy_all`. I don't have access to production data to see if this is the case or not.

spec/controllers/community_controller_spec.rb
```diff
@@ -35,8 +35,10 @@
         user = create(:student, hide_email: false)
         user_opted_out = create(:user, hide_email: true)
         get :index
-        expect(response.body).to include user.email
-        expect(response.body).not_to include user_opted_out.email
+        # expect(response.body).to include user.email
+        # This expectation ^ fails, but should this be tested in a controller test in
+        # the first place?
```
pgaspar: I'm confused 😕 What changed in this PR to make this expectation start failing?

pgaspar: Ah, I think this really is an issue with your changes:

```ruby
# Shouldn't this:
can?(:read, user.email)

# Be this:
can?(:read_email, user)
```

app/views/community/index.html.slim
pgaspar: Also, as pointed out below, I think the check should be `can?(:read_email, user)`.

spec/models/ability/guest_spec.rb
```diff
@@ -15,6 +15,8 @@
 
     it_behaves_like 'has access to public features'
 
+    it { expect(subject).not_to be_able_to(:read, other_user.email) }
```
pgaspar: Heads up I believe this should be `be_able_to(:read_email, other_user)`. Same below.

F3PiX: I _thought_ this is testing the desired outcome of the `read_email` rule? No?

app/views/community/index.html.slim
F3PiX: 🤔To use the `can?` helper , I'd need access to the single user, as I do have below. I'd think this hack is better than alternatives, like adding an extra @user ivar? Or...?
Neat it is not. 😢

app/views/community/index.html.slim
F3PiX: Oiii, thanks for catching this. Turns out I was solving the wrong problem. Fixing the ability rules also fixed the layout _and_ what I was trying to do with the new ability rule.

spec/controllers/community_controller_spec.rb
F3PiX: Yes, it should! See other comment. 💃

app/views/community/index.html.slim
pgaspar: Hmm, good point. I wonder what people usually do in these circumstances - we are kinda duplicating authorization knowledge here.

Also this seems to be slightly different from what we have specified in the Ability class: from what I understand, confirmed users can: `can :read_email, User, hide_email: false`, can't they? (I may be missing something here! I'm not too familiar with CanCan).

db/migrate/20180524190818_destroy_non_existing_git_hub_users.rb
F3PiX: 1) I like the ruby destroy better anyway. 
2) Are you sure about not running the migration in dev env? I'd think that we want the user objects in the local database not to be different from the test and production databases?

pgaspar: Then maybe we could generate `github_id`s for users without one in development to make them be compatible without deleting them?

db/migrate/20180524190818_destroy_non_existing_git_hub_users.rb
```diff
@@ -0,0 +1,12 @@
+class DestroyNonExistingGitHubUsers < ActiveRecord::Migration[5.1]
+  def up
+    unless ENV['RAILS_ENV'] == 'development'
+      User.find_each do |user|
+        if user.github_id == nil
+          puts "removing user #{user.id} from database"
+          user.destroy
```
pgaspar: This may not matter much, depending on the number of records we have, but I think this version leads to better performance:

```ruby
User.where(github_id: nil).destroy_all.each do |user|
  puts "removed user #{user.id} from database"
end
```

`destroy_all` returns the records that were removed, so we can use those to print stuff.

PS: did a quick one-run test and for 5000 users this method ran in ~23 seconds while the inline `user.destroy` method ran in ~41 seconds. That being said, this probably doesn't matter at all since we don't have a lot of records being destroyed! 😅

spec/controllers/community_controller_spec.rb
```diff
@@ -36,7 +36,8 @@
         user_opted_out = create(:user, hide_email: true)
         get :index
         expect(response.body).to include user.email
-        expect(response.body).not_to include user_opted_out.email
+        # Should this be tested in a controller test?
+        expect(response.body).not_to include user_opted_out.email # false negative; also passes with hide_email: false
```
pgaspar: @F3PiX I may be missing something here, but I _think_ this may not be a false negative. Changing `user_opted_out` to have `hide_email: false` will make the test fail.

spec/features/users/guest_user_spec.rb
```diff
@@ -2,8 +2,8 @@
 
 RSpec.describe 'Guest User', type: :feature do
 
-  let!(:activity)      { create(:status_update, :published, team: team1) }
-  let!(:other_user)    { create(:user) }
+  let!(:activity) { create(:status_update, :published, team: team1) }
+  let(:other_user) { create(:user, hide_email: false) }
```
pgaspar: Isn't this change unnecessary? Removing it does not make the spec fail (I assume `hide_email` defaults to `false`).

Are you doing it for clarity? 👍 if you are.

spec/features/users/unconfirmed_sign_in_user_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+      email.set("newemail@example.com")
+      find_button("Save").click
+      expect(page).to have_content("updated")
+      expect(page).to have_content("Please click on the link in the email we just sent to confirm your email address. Haven't received it? Click here to resend the email.")
+      expect(page).to have_content("About me")
+    end
+
+    it 'nothing happens when clicking on resend link' do
+      find_link('Click here to resend the email.').click
+      # todo: expect_to happen: something visible
```
pgaspar: Maybe check that an email was enqueued here or something?

spec/models/ability/guest_spec.rb
pgaspar: I'm not experienced with CanCan, but looking at how the rule is defined, using `read_email` instead of `read`, I would assume you need to use `read_email` here as well. Also, the rule definition does not receive an email, right? It receives a User.

But it should be testable. I made these changes and the spec fails:

screen shot 2018-05-28 at 13 24 58

Using this makes it pass:

```ruby
it { expect(subject).to be_able_to(:read_email, other_user) }
```

app/views/community/index.html.slim
```diff
@@ -31,8 +31,7 @@ table.users.table.table-striped.table-bordered.table-condensed.table-hover
         td.image = avatar_url(user, size: 40)
         td = link_to user.name_or_handle, user
         td = user.interested_in.map { |key| User::INTERESTS[key].presence }.compact.join(', ')
-        - if signed_in?
-          td = mail_to user.email unless user.hide_email?
+        td = mail_to user.email if can?(:read_email, user)
```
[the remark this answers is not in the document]
F3PiX: 🙈 I thought I had it right. Sorry. 
I didn't miss your suggestion, I just found that it gave problems when a user as their email hidden. Look:

two_line_if_one_hidden_email
(first user hides email).

I will look again!!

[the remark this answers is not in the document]
pgaspar: Ah, that's a great point - I'm sorry I missed that!

I haven't tested this, but maybe something like this would work:

```slim
- if can?(:read_email, nil)
  td = mail_to user.email if can?(:read_email, user)
```

The idea being that the first check will be `false` for unconfirmed / logged out users, while being true for other users. It's not the prettiest thing every, but it may solve the issue without us needing to explicitly check if the user is confirmed (which duplicates knowledge that unconfirmed/logged out users should not see emails).

app/views/community/index.html.slim
F3PiX: That didn't work for me, not as a twoliner, and not as a oneliner :-( 
See next commits.

spec/features/users/unconfirmed_sign_in_user_spec.rb
F3PiX: I mean there is no visual feedback to the user.

pgaspar: Yep, I understood that part 😅 I was just saying that maybe we should also check that an email was enqueued, because that's what happens, right? Or maybe it isn't? I'm a bit confused by the spec name `nothing happens when clicking on resend link`.

F3PiX:
@pgaspar Sorry for keeping you busy 🙏 
What I did: I made it so that the column title is hiding now in sync with the email's visibility. That means, we have a stray mini column for guests and unconfirmed users. 
I rather left it as is for now, because both the authentication is not finished, and the ability rule :read_email is not renewed. The thing is: I couldn't leave it alone, because the email addresses got exposed to guest users. My priority was to prevent that. 
We already have an issue for the :read_email rule (I'll update that one with our acquired knowledge and questions).

spec/models/ability/guest_spec.rb
F3PiX: You are right. I remember what I did here. The thing is, the read_email rule is not yet renewed yet. So it is hard to test it (and use it in the community view) the right way. 
The one thing I wanted to guard is that a guest user can not read  emails. I am 98%  sure that the test is testing the right thing here. 😇 
See my general comment for a proposal on how to handle this.

F3PiX:
For confirmed users it looks like this now:
schermafbeelding 2018-05-28 om 20 32 22

For unconfirmed and guests like this 💩 
schermafbeelding 2018-05-28 om 20 31 48

pgaspar:
Thank you @F3PiX 👍 And thanks for your patience.

If you're willing to try one last solution, this should get rid of the empty column for guests:

```slim
- if current_user&.confirmed?
  th Email

/ ...

- if current_user&.confirmed?
  td = mail_to user.email if can?(:read_email, user)
```

Did a quick test and it seemed to work (and it should work, looking at the code).

spec/models/ability/guest_spec.rb
pgaspar: Still think this is a false positive @F3PiX 😟 I wasn't able to make `be_able_to(:read, other_user.email)` return true even for cases where it should return true, as I showed above.

This is the same error you did initially, where you were testing things in the view as `can?(:read, user.email)` and it didn't work and we changed it to `can?(:read_email, user)`. It makes sense that the same logic would apply here in the expectation.

F3PiX: O yes. So sorry. You are right.

db/migrate/20180524190818_destroy_non_existing_git_hub_users.rb
F3PiX: I intended to add a 'how-to' to the readme or troubleshooting guide. Or should I add it to the migration? Add a part to add gh_ids if env == dev?

Sorry, I missed this comment yesterday.

F3PiX:
@pgaspar I updated the specs; now they do test the `read_email` rule. Thanks.
One thing left to do: 
[ ] add github_ids to users in dev env 
Should I add it to the migration? Add a separate migration? WDYT?

pgaspar:
Great work @F3PiX 👍

As for the `github_ids`, I think both options are fine, so whatever you prefer 😸 There is probably a way to do this with SQL directly using Random() or something, but you can do it with Ruby as well...

db/migrate/20180524190818_destroy_non_existing_git_hub_users.rb
```diff
@@ -1,6 +1,14 @@
...
+
   def up
-    unless ENV['RAILS_ENV'] == 'development'
+    return if ENV['RAILS_ENV'] == 'test'
+    if ENV['RAILS_ENV'] == 'development'
+      puts "Updating users with github_id ...."
+      users = User.all.select { |user| user.github_id.nil? }
+      users.each_with_index do |user, index|
+        user.update!(github_id: index + 10)
+      end
```
pgaspar: @F3PiX `select` is doing this in ruby which is slow 😟 `users = User.where(github_id: nil)` does the same thing but in the database, which is fast!

`users.each_with_index` is not great for memory because it loads every record without a github id at the same time into memory.

Here's an alternative:

```ruby
User.where(github_id: nil).find_each(batch_size: 200).with_index do |user, index|
  user.update!(github_id: index + 10)
end
```

You can find a similar example in the docs here: http://api.rubyonrails.org/classes/ActiveRecord/Batches.html#method-i-find_each

F3PiX:
@pgaspar I got distracted before I submitted my comment.
I was going to say:
- I put it in the same migration, so that the nil records don't get deleted first :-)
- It will never be more than ca 30 records, so I didn't care about performance.

Do you want me to fix it anyway? I'll do that tomorrow then!  Either way, I will stick to the `where`'s in the future.

F3PiX:
Thanks so much @pgaspar for having my back! 🙇 Much appreciated!

Issue #982, opened 2018-04-23
User registration: ask user to provide email if it cannot be retrieved from Github


carpodaster:
Users can mark their email address private on their Github profile, making it unavailable for API clients like the one retrieving user data by github handle when a user registers in our site.

If the registration process cannot complete with a valid email address, redirect the user to a form and ask them to provide one.

**Note**: this may already be implement. Investigate first.

ramonh:
I've done some research and come to some conclusions:

1. You cannot Oauth on a fresh Github account without a verified email address:
![screen shot 2018-05-02 at 13 32 28](https://user-images.githubusercontent.com/656318/39521167-b1d7b648-4e0d-11e8-9c0e-8c5aac0d1e9b.png)

2. Marking an email address as private still allows this to be imported by the Oauth app

3. Creating a new account takes the new user to complete their profile.

Given this information, what do you think? I'd almost say that, given the above, we can leave it as is.

klappradla:
Oh wow @ramonh 🙇 

That's ace! Means the problem of not being able to retrieve an email address actually does not exist at all?

When and under which circumstances was this then originally reported @carpodaster ?

F3PiX:
When I start a new team, I can add team members that don't exist: `partyppooper`, or that have a hidden email address: Sandi Metz `skmetz`.
In both cases we store a user with `email address: nil`

I know I did something like this for real, when I didn't know Roos yet and I wanted to see what the application form was like. So, it's not a hypothetical case. 
Do we (need to) get rid of the fake users?

ramonh:
Ahh! Good point, @F3PiX. I overlooked adding team members. Maybe we should reopen this, huh?

F3PiX:
Well, I don't know by heart what happens when a user added by team members sign in with Gh the first time. (See conv in Slack, I am working on sketching the flow.)
Maybe it is a different issue alltogether - related to cleaning up the database?

F3PiX:
Definitively reopen
Reproduce:
- Make yourself a team member on the website
- Add a fake name, for a non existing user (like partypppooper) 
- On the teampage, click on the newly create team members name. => 404 
It is not a valid github name, but is added to our database and as a team member. 

In console, check out User.last. See that non of the github fields are populated, except for the fake handle. 

- [ ] We use  `find_or_create_by_github_handle` to create a new user. I'd like to investigate if we could change that to `github_id`, to catch invalid entries. 
TBD later, adding it to the list for #991

PR #967, opened 2018-03-12
[WIP] Correct student names in confirmation email

cesswairimu:
Restore email sending code
So as to investigate the bug and find solution

Related issue #955

PR #1007, opened 2018-05-26
Devise overhaul

F3PiX:
Related issue #991 #982
Users who need to update their profile should be redirected to the Edit page on sign in. 

This PR is about straightening out the Devise workflow for signing in and redirection.
The reasoning behind this: 
- Either users are confirmed, or they need to update their profile and thus be redirected to Edit page. 
So, use only `user.confirmed?` to decide where they need to be redirected.
- Confirmed users are redirected back to the page where they were when they signed in.  

Plus:
-  Moved the 'deliver_later` method to the user model, so all devise notifications are delivered_later now
- Moved the `after_sign_in_path` to the OAuth controller, because it is the only sign in in we use. 
- Updated texts on help page
- Added help text to the Edit Profile page, updated messages. 
- Updated specs 

ToDo / leftovers
- [ ] One of the specs in applications_draft_controllers needs investigation; made it pending for now
- [ ] One important Devise issue left: we are sending confirmation mails over and over again to unconfirmed users (in users#update) when they change any profile field.  There are several options that we can discuss after this PR is agreed upon.
- [ ] After landing on the Edit Profile page, a user can still navigate away without submitting the form.
- [ ] The Welcome message and the hint under the email field is party duplicated. Needs finetuning later. And the hint needs design love. 
- [ ] The email hint  should only show when the user changed the value in the email field (JS :on_change)
- [ ] A confirmed user who changed their email address should have the yellow confirm-your-email notice after submitting the form. Currently, they don't see that. (Their status is still `confirmed`)   
- [ ] The country + city fields are required, but not validated as a combination. So, Amsterdam on Antartica is valid input. (TBH I think the UX on first sign in would be better if only the email address would be required.)

### DIFF
--- a/app/controllers/omniauth_callbacks_controller.rb
+++ b/app/controllers/omniauth_callbacks_controller.rb
@@ -3,15 +3,21 @@ class OmniauthCallbacksController < Devise::OmniauthCallbacksController
   def github
     user = User.find_or_create_for_github_oauth(request.env['omniauth.auth'])
 
-    if user.just_created? || !user.valid? || user.github_import || !user.confirmed? || user.email.nil? || user.github_id.nil?
+
+    # if user.just_created? || !user.valid? || user.github_import || !user.confirmed? || user.email.nil? || user.github_id.nil?
+    # They ^ have one thing in common: they can not be confirmed yet
+    if !user.confirmed?
       sign_in user
+      # flash[:notice] = I18n.t 'devise.omniauth_callbacks.success', kind: 'Github' # add this notice for clarity
       redirect_to edit_user_path(user, welcome: true)
-    elsif user.persisted?
+    else
+    # elsif user.persisted? # all users persists
       flash[:notice] = I18n.t 'devise.omniauth_callbacks.success', kind: 'Github'
       sign_in_and_redirect user, event: :authentication
-    else
-      session['devise.github_data'] = request.env['omniauth.auth']
-      redirect_to new_user_registration_url  # this route doesn't go anywhere (we don't use Devise registerable module)
+    # nothing else
+      # else
+    #   session['devise.github_data'] = request.env['omniauth.auth']
+    #   redirect_to new_user_registration_url  # this route doesn't go anywhere (we don't use Devise registerable module)
     end
   end
 end
--- a/app/controllers/omniauth_callbacks_controller.rb
+++ b/app/controllers/omniauth_callbacks_controller.rb
@@ -5,4 +5,13 @@ class OmniauthCallbacksController < Devise::OmniauthCallbacksController
     flash[:notice] = I18n.t 'devise.omniauth_callbacks.success', kind: 'Github'
     sign_in_and_redirect user, event: :authentication
   end
+
+  def after_sign_in_path_for(user)
+    if user.confirmed?
+      request.env['omniauth.origin'] || root_path
+    else
+      edit_user_path(user, welcome: true)
+    end
+  end
+
 end
--- a/app/controllers/omniauth_callbacks_controller.rb
+++ b/app/controllers/omniauth_callbacks_controller.rb
@@ -2,7 +2,7 @@
 class OmniauthCallbacksController < Devise::OmniauthCallbacksController
   def github
     user = User.find_or_create_for_github_oauth(request.env['omniauth.auth'])
-    flash[:notice] = I18n.t 'devise.omniauth_callbacks.success', kind: 'Github'
+    flash[:notice] = I18n.t 'devise.omniauth_callbacks.success', kind: 'GitHub'
     sign_in_and_redirect user, event: :authentication
   end
 
--- a/app/controllers/omniauth_callbacks_controller.rb
+++ b/app/controllers/omniauth_callbacks_controller.rb
@@ -3,7 +3,7 @@ class OmniauthCallbacksController < Devise::OmniauthCallbacksController
   def github
     user = User.find_or_create_for_github_oauth(request.env['omniauth.auth'])
 
-    if user.just_created? || !user.valid? || user.github_import
+    if user.just_created? || !user.valid? || user.github_import || !user.confirmed? || user.email.nil? || user.github_id.nil?
       sign_in user
       redirect_to edit_user_path(user, welcome: true)
     elsif user.persisted?
@@ -11,7 +11,7 @@ class OmniauthCallbacksController < Devise::OmniauthCallbacksController
       sign_in_and_redirect user, event: :authentication
     else
       session['devise.github_data'] = request.env['omniauth.auth']
-      redirect_to new_user_registration_url
+      redirect_to new_user_registration_url  # this route doesn't go anywhere (we don't use Devise registerable module)
     end
   end
 end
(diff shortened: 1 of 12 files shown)


app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -2,16 +2,7 @@
...
-    if user.just_created? || !user.valid? || user.github_import
-      sign_in user
-      redirect_to edit_user_path(user, welcome: true)
-    elsif user.persisted?
-      flash[:notice] = I18n.t 'devise.omniauth_callbacks.success', kind: 'Github'
-      sign_in_and_redirect user, event: :authentication
-    else
-      session['devise.github_data'] = request.env['omniauth.auth']
-      redirect_to new_user_registration_url
-    end
```
F3PiX: All our users are either persisted on :create, or not confirmed.  So I can't think of a situation where the third clause is triggered? All the tests pass without this clause, and I can't think of a reason we need it.

F3PiX:
blocked until #1009 is merged

F3PiX:
blocked

F3PiX:
Before (current master):
- Sign in, unconfirmed:
unconfirmedsigninmaster
  - Go to Profile -> edit, change email address. 
     The profile show page shows the old email address, looks like nothing happened

After:
- Sign in, unconfirmed:
schermafbeelding 2018-06-01 om 14 54 32

F3PiX:
This is ready for review 💃

app/views/users/edit.html.slim
```diff
@@ -1,8 +1,8 @@
 - if params[:welcome]
   h1 Welcome!
 
-  p Please complete your profile:
-
+  p Please <strong>check your email address</strong> and fill all required fields to complete signing up.
```
ramonh: Should we notify users that updating their email address will require confirming, or will updating the email address reset the `confirmed` status?

spec/features/users/unconfirmed_sign_in_user_spec.rb
```diff
@@ -49,7 +49,7 @@
       email.set("newemail@example.com")
       find_button("Save").click
       expect(page).to have_content("updated")
-      expect(page).to have_content("Please click on the link in the email we just sent to confirm your email address. Haven't received it? Click here to resend the email.")
+      expect(page).to have_content resend_info
```
ramonh: Great cleanup!

app/views/users/edit.html.slim
```diff
@@ -34,6 +34,13 @@
 
   = f.input :name, wrapper_html: { class: 'col-md-6' }
   = f.input :email, required: true, wrapper_html: { class: 'col-md-6' }
+  - unless current_user.confirmed?
+    .help-block
```
pgaspar: I wonder if this help text should be more visible. Looking at the screenshot you shared it may kinda get lost in the rest of the info in the page.

Sorry, I know this isn't very actionable. Mentioning it in case someone has a better suggestion.

spec/controllers/users_controller_spec.rb
```diff
@@ -89,16 +89,21 @@
...
-            }.to change { ActionMailer::Base.deliveries.count }.by(1)
+            }.to have_enqueued_job.on_queue('mailers')
+            expect(enqueued_jobs.size).to eq 1
           end
 
           it "sends only one confirmation email if the user isn't confirmed yet and the email was changed" do
             expect {
               put :update, params: { id: user.to_param, user: { name: 'Trung Le', email: 'newmail@example.com' } }
-            }.to change { ActionMailer::Base.deliveries.count }.by(1)
+            # }.to change { ActionMailer::Base.deliveries.count }.by(1)
```
pgaspar: Maybe we can remove this now? 😊

spec/controllers/application_drafts_controller_spec.rb
```diff
@@ -233,6 +235,8 @@
           end
 
           it 'sends 1 mail to orga' do
+            pending "It changes by 3 instead of 1"
+            expect(enqueued_jobs.size).to eq 0
```
pgaspar: Can we maybe fix this spec in this PR? Or isn't it valid any more?

app/views/users/edit.html.slim
F3PiX: 👍 The notification should be in the (yellow) notice that should appear after they submit the form. Updating the email address will not reset the `confirmed` status, but will set the `unconfirmed_email`.

Currently the notification only appears for _un_confirmed users, so I'll fix that! Thanks!

F3PiX: No wait, that cannot be solved yet. See my general comment below.

app/views/users/edit.html.slim
F3PiX: I added a todo to give the help text some design love.

spec/controllers/users_controller_spec.rb
F3PiX: 😊

spec/controllers/application_drafts_controller_spec.rb
F3PiX: I added it to the todo list. TBH I am very unfamiliar with this corner of the app, and I'd rather not dive into it at the moment, because my head is already filled with all things Devise. 
Also, I hoped it would go away all by itself with PR #967 https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/967/files#diff-dc4cf03c7b58194272d6cea95a76d35e 😇

app/views/users/edit.html.slim
ramonh: Hmm, good point, @pgaspar.

How about using [highlighting](https://getbootstrap.com/docs/3.3/css/#marked-text)? Wdyt?

spec/controllers/application_drafts_controller_spec.rb
pgaspar: Oh, is this failing on `master`? Sorry, I always assume that if a spec breaks on a PR it's because of something the PR changed (that's why I thought it would be related to this work).

app/views/users/edit.html.slim
pgaspar: I think it would help 😸

spec/controllers/application_drafts_controller_spec.rb
F3PiX: No you are totally right, it is failing in this PR 😊 , and not in master. I looked again, but I can't find why.
What do you think, can I leave it alone and see if it will be solved in #967 ?

pgaspar: Ok, I looked into what `enqueued_jobs` had in it and it seems the 2 extra emails are confirmation emails for both students. As far as I understand the change in #967 is adding 2 new emails for the submitters (hence the change they did in the specs), so it's a different thing and won't fix the error we're seeing here (it will just say it expected the size to change by 3 but it changed by 5).

One way to solve this in the spec is to use `let!` when creating the draft, from what I tested.

Note: I'm a bit lost as to where in this PR we started sending the confirmation emails. Is it the change to `send_devise_notification`?

F3PiX: Ooo that makes totally sense!!! We already sent confirmation emails, we just delivered them `now`, because the `deliver_later` implementation was only accessible for users created through github.  
Yes, that is the `send_devise_notification` moved to the user model in https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/1007/commits/5d2158b585e6176bece601eeaea0f230a18a1dec
I suspected it has to do with the create_users method, but couldn't find the `let!` solution. Thank you!!

F3PiX:
@pgaspar Bang! in place, specs green. Thanks! 
One more look please?

PR #1009, opened 2018-05-26
Hotfix: Broken routes to Notepad on Supervisor Dashboard 

F3PiX:
The Supervisor Dashboard didn't work anymore, because the routes to the notepad and the comments were broken (after the namespace makeover in #893).
Fixed the route reference for comments, and reverted the removal of the separate `patch` notes route.
This may not be the optimal solution, but now the dashboard is working again. Ready to roll for the Summer! 🌞

### DIFF
--- a/app/views/supervisors/dashboard/index.html.slim
+++ b/app/views/supervisors/dashboard/index.html.slim
@@ -27,7 +27,7 @@ div.row
                 td
                   '= Red alert
           div.col-sm-6 class=("col-lg-4 padding" if supervised_teams.count > 2)
-            = simple_form_for Note.notepad(current_user), url: supervisor_notes_path do |f|
+            / = simple_form_for Note.notepad(current_user), url: supervisors_notes_path do |f|
               == render 'supervisors/notes/form', f: f
 
 div.row
@@ -43,7 +43,7 @@ div.row
             p = "You last checked on them #{ time_ago_in_words(t.comments.last.created_at) } ago."
           - else
             p You didn't check on them yet.
-          = simple_form_for Comment.new(commentable: t), url: supervisor_comments_path do |f|
+          = simple_form_for Comment.new(commentable: t), url: supervisors_comments_path do |f|
             == render 'supervisors/comments/form', f: f
             - if t.comments.any?
               p = render t.comments.recent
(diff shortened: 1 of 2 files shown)


config/routes.rb
```diff
@@ -101,10 +101,10 @@
     end
   end
 
+  patch 'supervisors/notes', to: 'supervisors/notes#update'
```
ramonh: I'm curious, what let do needing this change? I'm wondering only because the former looked nicer 😄. However, please feel free to disregard this as the rest looks great.

F3PiX: > The Supervisor Dashboard didn't work anymore, because the routes to the notepad and the comments were broken (after the namespace makeover in #893).

The dashboard is not reachable. Reproduce: assign your self a supervisor role, click on the link to supervisor dashboard -> 💥

pgaspar: I think another solution to this is to change `resources` to `resource`, which may be a bit more rails-y: `resource :notes, only: [:update]`.

Update: as I was thinking about this, I pulled the branch and tried it out. I noticed the error that the regular `resources :notes, only: [:update]` was getting me is that `supervisors_note_path` requires a note ID, which makes sense, since we would be updating an existing record.

Replacing line 30 with this form, makes it work:

```slim
= simple_form_for [:supervisors, Note.notepad(current_user)] do |f|
```

It uses the `note_path`, but namespaced to `supervisors`.

On the other hand, if we really want to make it so no ID is ever passed, because we will always be editing the same Note, I'd suggest using `resource`, since I think it [matches our intention here](http://guides.rubyonrails.org/routing.html#singular-resources). In that case, maybe it should be `resource :note` and not `noteS`.

app/views/supervisors/dashboard/index.html.slim
```diff
@@ -43,7 +43,7 @@ div.row
             p = "You last checked on them #{ time_ago_in_words(t.comments.last.created_at) } ago."
           - else
             p You didn't check on them yet.
-          = simple_form_for Comment.new(commentable: t), url: supervisor_comments_path do |f|
+          = simple_form_for Comment.new(commentable: t), url: supervisors_comments_path do |f|
```
pgaspar: @F3PiX I think you also need to change `supervisor_comments_path` to `supervisors_comments_path` on line 50.

config/routes.rb
F3PiX: 😍

F3PiX:
@pgaspar Back to you 🐝 Works like a charm! So much better!! Thanks!

ramonh:
🎉

PR #1014, opened 2018-06-02
Fix issue populating Team dropdown

pgaspar:
This PR is the continuation of the work started by @simransinghal and @ArushiSinghal on #941 👍
Related issue #934 

**Problem:** the teams dropdown on the Activities page becomes empty between seasons.

screen shot 2018-06-02 at 17 24 45

**Solution** (from #934):

1. During the current season, the dropdown is populated with the teams that are accepted in the current season
2. Until the current season starts, the dropdown is populated with the teams that were accepted in the previous season

**PS:** let's please add the original contributors as co-authors when we merge 🙌

### DIFF
--- a/app/controllers/activities_controller.rb
+++ b/app/controllers/activities_controller.rb
@@ -28,7 +28,7 @@ class ActivitiesController < ApplicationController
   end
 
   def teams
-    Team.accepted.in_nearest_season.order(:name)
+    @teams ||= Team.accepted.in_nearest_season.order(:name)
   end
   helper_method :teams
 
--- a/app/controllers/activities_controller.rb
+++ b/app/controllers/activities_controller.rb
@@ -28,14 +28,9 @@ class ActivitiesController < ApplicationController
   end
 
   def teams
-    Team.in_current_season.accepted.order(:name)
+    (Team.in_current_season.accepted.presence || Team.in_previous_season.accepted).order(:name)
   end
   helper_method :teams
-
-  def last_season_teams
-    Team.select {|t| t.season_id == Season.current.id + 1 && (t.kind == 'sponsored' || t.kind == 'voluntary' || t.kind == 'full_time' || t.kind == 'part_time')}
-  end
-  helper_method :last_season_teams
 
   def normalize_params
     params[:kind] = 'all' if params[:kind].blank?
--- a/app/controllers/activities_controller.rb
+++ b/app/controllers/activities_controller.rb
@@ -28,7 +28,7 @@ class ActivitiesController < ApplicationController
   end
 
   def teams
-    (Team.in_current_season.accepted.presence || Team.in_previous_season.accepted).order(:name)
+    Team.accepted.in_nearest_season.order(:name)
   end
   helper_method :teams
 
--- a/app/controllers/activities_controller.rb
+++ b/app/controllers/activities_controller.rb
@@ -32,6 +32,11 @@ class ActivitiesController < ApplicationController
   end
   helper_method :teams
 
+  def last_season_teams
+    Team.select {|t| t.season_id == Season.current.id + 1 && (t.kind == 'sponsored' || t.kind == 'voluntary' || t.kind == 'full_time' || t.kind == 'part_time')}
+  end
+  helper_method :last_season_teams
+
   def normalize_params
     params[:kind] = 'all' if params[:kind].blank?
   end
--- a/app/models/team.rb
+++ b/app/models/team.rb
@@ -51,6 +51,7 @@ class Team < ApplicationRecord
   scope :in_current_season, -> { where(season: Season.current) }
   scope :in_previous_season, -> { by_season(Date.today.year - 1) }
   scope :by_season, ->(year) { joins(:season).where(seasons: { name: year }) }
+  scope :in_nearest_season, -> { in_current_season.presence || in_previous_season }
 
   class << self
     def ordered(sort = {})
--- a/app/models/team.rb
+++ b/app/models/team.rb
@@ -49,6 +49,7 @@ class Team < ApplicationRecord
   scope :accepted, -> { full_time.or(part_time) }
   scope :visible, -> { where.not(invisible: true).or(accepted) }
   scope :in_current_season, -> { where(season: Season.current) }
+  scope :in_previous_season, -> { by_season(Date.today.year - 1) }
   scope :by_season, ->(year) { joins(:season).where(seasons: { name: year }) }
 
   class << self
--- a/spec/models/team_spec.rb
+++ b/spec/models/team_spec.rb
@@ -349,6 +349,29 @@ RSpec.describe Team, type: :model do
       end
     end
 
+    describe '.in_nearest_season' do
+      let(:current_season) { create(:season, name: Date.today.year) }
+      let!(:current_teams) { create_list(:team, 2, season: current_season) }
+      let(:prev_season) { create(:season, name: Date.today.year - 1) }
+      let!(:prev_teams) { create_list(:team, 2, season: prev_season) }
+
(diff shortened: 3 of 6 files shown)


app/controllers/activities_controller.rb
```diff
@@ -28,7 +28,7 @@ def set_activities
   end
 
   def teams
-    Team.in_current_season.accepted.order(:name)
+    (Team.in_current_season.accepted.presence || Team.in_previous_season.accepted).order(:name)
```
F3PiX: Cool!
Do you think we could make this even simpler as `Team.accepted.in_nearest_season` , where `in_nearest_season` is a new scope for in current season OR in previous season?
(Note, we do have an index teams_on_season, not on teams_on_kind. Maybe worth to add that if you want to go with ^)

app/models/team.rb
```diff
@@ -49,6 +49,7 @@ class Team < ApplicationRecord
   scope :accepted, -> { full_time.or(part_time) }
   scope :visible, -> { where.not(invisible: true).or(accepted) }
   scope :in_current_season, -> { where(season: Season.current) }
+  scope :in_previous_season, -> { by_season(Date.today.year - 1) }
```
F3PiX: 👍

pgaspar:
I noticed something when testing this - we're ordering the dropdown by team `name`. The result isn't perfect, I think:

screen shot 2018-06-02 at 23 35 35

I committed a change to order by team `number` instead. Thoughts?

screen shot 2018-06-02 at 23 35 08

**Update:** Actually, scratch this. The order only looks weird in the seeded database. Production data doesn't use the "Team #{number} - ..." format. I'm removing my commit.

app/controllers/activities_controller.rb
pgaspar: @F3PiX I tried adding the index on `kind` and was trying to use ActiveRecord's `explain` to see if the index was being used or not. I couldn't get it to be used on `Team.accepted`, only when filtering specific kinds.

screen shot 2018-06-03 at 00 46 22

screen shot 2018-06-03 at 00 46 47

(I had to create 3000 teams for the index to be used, by the way. Postgres was not using it even with 200 teams.)

Here's `Team.in_nearest_season.accepted`:

Without `kind` index:

screen shot 2018-06-03 at 00 55 22

With the `kind` index:

screen shot 2018-06-03 at 00 53 54

I only had 2 seasons for this experiment - not sure if that matters too much (I don't think so).

So, I don't know if it's worth it. But I can commit it if you think it makes sense! 👍

pgaspar:
@F3PiX this up for review!

app/controllers/activities_controller.rb
F3PiX: > I couldn't get it to be used on Team.accepted, only when filtering specific kinds.

Oooo. I didn't think of that. Is that because `accepted` is not a field in the database? I don't know enough about this, but maybe a compound index on season and kind would work with accepted? Or is that the same difference? I really don't know. Interesting!

> So, I don't know if it's worth it.  

Well, if it doesn't use the index with the `accepted` scope.... no. 
I didn't expect a lot better on this particular query, but we mostly use `teams` as accepted teams, so I'd expect a better performance over all. 
But again, not if the index is not being used.

Anyway, no reason to keep this PR on hold over this.  
Thank you so much for the clarification!

F3PiX:
@pgaspar 
See the comment in the code about the indexing. 🥇 

I like your `order by :id` much better than order by :name,  🎉 , so: +1 !

spec/models/team_spec.rb
```diff
@@ -349,6 +349,29 @@
...
+
+      context 'when current season has no teams' do
+        before { current_teams.each(&:destroy) }
+
+        it 'returns last season\'s teams' do
+          expect(returned_teams).to match_array(prev_teams)
+        end
+      end
+    end
+
```
F3PiX: 😍

app/controllers/activities_controller.rb
```diff
@@ -28,7 +28,7 @@ def set_activities
   end
 
   def teams
-    Team.in_current_season.accepted.order(:name)
+    Team.accepted.in_nearest_season.order(:name)
```
F3PiX: Afterburner: should we memoize this?

pgaspar:
@F3PiX I added this update to the order comment:

> **Update:** Actually, scratch this. The order only looks weird in the seeded database. Production data doesn't use the "Team #{number} - ..." format. I'm removing my commit.

I don't have access to production data to try things in the rails console - I based this on these team names: https://teams.railsgirlssummerofcode.org/teams?year=2016

What do you think? With names like this it kinda makes sense to me to order by name.

app/controllers/activities_controller.rb
pgaspar: > Is that because `accepted` is not a field in the database?

I think it's because the SQL query is an includes query:

```SQL
SELECT "teams".* FROM "teams" WHERE ("teams"."kind" IN ('full_time', 'sponsored') OR "teams"."kind" IN ('part_time', 'voluntary'))
```

I tried the query without the `OR` and it also did not use the index.

screen shot 2018-06-03 at 13 43 03

I think it may be because the result of this query in my local database is the entire teams table - apparently I don't have companies that are not accepted 😛 Yep, I think that's why. Here's me trying an includes query with values that don't exist:

screen shot 2018-06-03 at 13 49 42

As you can see it uses the index. Here's more context I found [here](https://robots.thoughtbot.com/why-postgres-wont-always-use-an-index):

> Given a large enough number of requested rows (such as all the posts with an id greater than 10), it is actually more efficient to go fetch all those rows from the table directly rather than to go through the extra step of asking the index for the position of those rows and next fetch those specific rows from the table.
> If the percentage of the rows is smaller than 5-10% of all rows in the table the benefit of using the information stored in the index outweighs the additional intermediate step.

I don't know how our production data look like, but I assume the vast majority of our teams are accepted, so it's likely the index will never be used for the `accepted` scope. But I'm no expert! It was fun exploring this though 😸 Thanks.

Didn't see changes with a compound kind, season_id index. Didn't explore much further though.

F3PiX: Very interesting stuff! Thanks for the background info! BTW, based on that info and as not all applying teams are accepted, I think the case for an index is getting stronger!

F3PiX:
@pgaspar I dm'ed you some production data on Slack. 
You are totally right! The seeds use a sequence to ensure unique names. I totally forgot about that. Yay for order by name ;-)

app/controllers/activities_controller.rb
pgaspar: Yep, given only 7.5% of teams in production are accepted, I think this makes sense 🚀

pgaspar: Yep, given only a fraction of teams in production are accepted, I think this makes sense 🚀

F3PiX:
Supernice!! I sent you a 'before' benchmark, I'll run one after as well.

F3PiX:
PS On production, there is one team display_name that is veeeeery long, and that makes the dropdown veeeeery wide. Should we use the Team's `name` instead of the `display_name`? Could you take a look please?

pgaspar:
I don't know if that's a great solution - it's helpful to see that these teams are from last year with the `[2017]` bit, which is added by `display_name`.

Looking at the code, it seems to use `Team#project_name` which seems to be a cached version of the actual project's name. Maybe we could edit that in production for this team record and make it shorter?

Maybe we should make this a new issue, what do you think?

F3PiX:
I agree that `display_name` adds sugar that we don't want to dismiss. And that we should create a new issue. 

If this happens more often than with this team only, we should consider to have something like a short format for the project_name, or split upcoming projects names in a title and a subtitle. We can discuss it in a new issue 👍  .

PR #1015, opened 2018-06-05
Seeds and mailer preview

F3PiX:
Related issue #

Dev happiness! 
- Update seeds with some unconfirmed users
- Update user factory's email field to `@example.com` (in case we'll want to send emails in the upcoming review apps)
- Add letter opener gem for supereasy previews of emails.

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -53,5 +53,4 @@ Rails.application.configure do
   # with letter_opener
   config.action_mailer.delivery_method = :letter_opener
   config.action_mailer.perform_deliveries = true
-  config.active_job.queue_adapter = :inline
 end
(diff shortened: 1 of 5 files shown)


config/environments/development.rb
```diff
@@ -48,7 +43,15 @@
...
+  # Action Mailer
+  config.action_mailer.perform_caching = false
   config.action_mailer.default_url_options = { host: 'localhost:3000' }
   config.action_mailer.raise_delivery_errors = false
   config.action_mailer.preview_path = "#{Rails.root}/lib/mailer_previews"
+
+  # with letter_opener
+  config.action_mailer.delivery_method = :letter_opener
+  config.action_mailer.perform_deliveries = true
+  config.active_job.queue_adapter = :inline
```
klappradla: I'd suggest to not do that 😉 It's just too far from reality if the real adapter is only used in production.

Depending on the adapter we're using the difference until an email is visible in the browser is only microseconds anyways. With Sidekiq it would be serializing to Redis and deserializing again which is fast enough.

PR #1017, opened 2018-06-06
Flick the switch on belongs_to pt.1

klappradla:
This is part one of enabling the new defaults introduced with Rails 5: `belongs_to` associations are required by default now.

This PR flicks the switch to enable this new default on adjusts all models and their tests accordingly to this new behavior (while the app stays exactly the same).

Based on this, I'll prepare more PRs for:
- remove unused belongs_to associations (e.g. `applications` <-> `projects`)
- make certain `belongs_to` associations actually required (the hard part 😉 )

If the context for the current changes is not super obvious in some places, it may be easier to look at the individual commits (with their messages)...

### DIFF
--- a/app/models/maintainership.rb
+++ b/app/models/maintainership.rb
@@ -1,8 +1,8 @@
 # frozen_string_literal: true
 
 class Maintainership < ApplicationRecord
-  belongs_to :project
-  belongs_to :user
+  belongs_to :project, optional: true
+  belongs_to :user, optional: true
 
   validates :project, :user, presence: true
   validates :project_id, uniqueness: { scope: :user_id }
--- a/app/models/maintainership.rb
+++ b/app/models/maintainership.rb
@@ -1,9 +1,7 @@
 # frozen_string_literal: true
-
 class Maintainership < ApplicationRecord
-  belongs_to :project, optional: true
-  belongs_to :user, optional: true
+  belongs_to :project
+  belongs_to :user
 
-  validates :project, :user, presence: true
   validates :project_id, uniqueness: { scope: :user_id }
 end
--- a/app/models/team.rb
+++ b/app/models/team.rb
@@ -2,7 +2,7 @@
 class Team < ApplicationRecord
   include ProfilesHelper, HasSeason
 
-  attr_accessor :checked
+  KINDS = %w(full_time part_time)
 
   belongs_to :project, optional: true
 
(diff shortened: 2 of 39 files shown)


app/models/maintainership.rb
```diff
@@ -1,9 +1,7 @@
 # frozen_string_literal: true
-
 class Maintainership < ApplicationRecord
   belongs_to :project
   belongs_to :user
 
-  validates :project, :user, presence: true
```
ramonh: 👍

config/initializers/new_framework_defaults_5_1.rb
```diff
@@ -10,8 +10,7 @@
 Rails.application.config.action_view.form_with_generates_remote_forms = false
 
 # The belongs_to associations are required by default in Rails 5
-# we cannot opt into this right now, but it would be a goal for the future
-Rails.application.config.active_record.belongs_to_required_by_default = false
+Rails.application.config.active_record.belongs_to_required_by_default = true
```
ramonh: 🎉

klappradla:
For your concern about forms for nested resources @F3PiX: you're right, it's all a bit messy and definitely not well tested... But as you said yourself:
> As far as I can read from the code that is still working as it should

This PR does not include any changes on the associations yet, it's basically just changing the syntax. So with the validations staying the same, I also expect the forms to behave the same as before 🤔 

Aside that, thanx for your comments!

config/initializers/new_framework_defaults_5_1.rb
ramonh: Did not know that, nice catch!

klappradla: Only if we actually enable the 5.1 defaults 😉 But I guess we can be save to do that.

app/models/team.rb
```diff
@@ -2,20 +2,9 @@
...
-  validates :name, presence: true, uniqueness: true
-  validate :disallow_multiple_student_roles
-  validate :disallow_duplicate_members
-  validate :limit_number_of_students
-  validate :limit_number_of_coaches
-
   attr_accessor :checked
 
-  belongs_to :project
+  belongs_to :project, optional: true
```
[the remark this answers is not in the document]
klappradla: Ok, I looked through some of them and shoved around some macros, etc. But being suuuper strict on this and having validation before or after scopes sounds a bit like bikshedding to me. So when things where grouped and made sense, I mostly kept it 😉

But I'll have a look at all of the models again anyways when checking whether they should require associations.

PR #1018, opened 2018-06-17
Add GDPR-friendly opt-in checkboxes for users

ramonh:
As discussed with the core organizing team, we're adding several opt-in options for users.


carpodaster:
Do we also need to track when the consent has been given? If see, we should add corresponding *_at datetime fields.

ramonh:
Thank you @carpodaster! We do indeed need to track this

db/migrate/20180617074552_add_communication_opt_in_to_users.rb
```diff
@@ -1,10 +1,16 @@
 class AddCommunicationOptInToUsers < ActiveRecord::Migration[5.1]
   def change
     add_column :users, :opted_in_newsletter, :boolean, default: false
+    add_column :users, :opted_in_newsletter_at, :datetime
```
klappradla: I think what @carpodaster meant was: have a `datetime` column **instead** of the `boolean` one, not both.

The previous boolean value for `false` would then just be `NULL` with the `datetime`column.

Same as e.g. you'd have `user.confirmed?`  just check for a value on `user.confirmed_at`, instead of both being separate columns on the DB (which you'd have to keep in sync which is a way harder problem 😉 )

app/controllers/users_controller.rb
```diff
@@ -92,6 +92,32 @@ def teams
...
+      :opted_in_newsletter_at,
+      :opted_in_announcements_at,
+      :opted_in_marketing_announcements_at,
+      :opted_in_surveys_at,
+      :opted_in_sponsorships_at,
+      :opted_in_applications_open_at
+    ).to_h.map do |opt_in, value|
+      next if opted_in?(opt_in_selected?(value), opt_in)
+      [opt_in, opted_in_time(value)]
+    end.to_h
```
ramonh: This is not pretty at all D: 

Any ideas how I could improve this, please?

ramonh:
Nearly there! Would appreciate some feedback, please. @carpodaster @klappradla

carpodaster:
@ramonh I've started a feedback but had to run before I could finish it. Will devote some more time tomorrow.

app/controllers/users_controller.rb
klappradla: Sorry for coming back to this so late @ramonh 🙇 

I don't know the exact circumstances for this feature, but I'll just quickly dump my brain here on how I'd have implemented it if it would have been in this very second 😉 (disclaimer: this is all just written out of memory, copy-pasting will probably 💥)

You could have a GDPR model concern (bad name, but just for an example):
```ruby
module GDPR
  extend ActiveSupport::Concern

  included do
    def opted_in_newsletter_at=(value)
      super(nil) unless value # should be set to nil for falsey values
      return value if value && self.opted_in_newsletter # don't do anything if already set
      super(Time.current)
    end

    # other methods similar
  end
end
```

If you then include this into the user model, you can just use ordinary checkboxes in the form and pass the params through as they are, without some `to_h` etc. magic in the controller.

One should probably dry this up, think about edge-cases, etc. But just as an example on how one could do it 😉

ramonh: Wow, this is a cool idea! Thank you @klappradla! I'll look into this

pgaspar: Hello! 👋

What I've done in the past is something similar to what this gem does: https://github.com/SonderCollection/timestamp_as_boolean

So, something along the lines of:

```ruby
  def opted_in_newsletter
    opted_in_newsletter_at.present?
  end
  alias opted_in_newsletter? opted_in_newsletter

  def opted_in_newsletter=(checkbox_value)
    accepted = ActiveRecord::Type::Boolean.new.cast(checkbox_value)
    if accepted
      self.opted_in_newsletter_at ||= Time.now
    else
      self.opted_in_newsletter_at = nil
    end
  end
```

Then you could use `opted_in_newsletter` in the checkboxes.

There may be other more known gems that do this - this one was what I found after a quick google search. I like the idea of making this a concern 👍 And it should be DRY'd too.

This is just to give you more options, if you need them - I like @klappradla's solution too 😃

ramonh: Cool! Thanks for the tip, @pgaspar. 

I'll weigh out both options. 

Thanks!

ramonh:
Hey everybody, thank you once again for all your feedback!

I've gone ahead and implemented the feature according to @pgaspar's solution. 

Could somebody please have a look?

app/models/user.rb
```diff
@@ -87,6 +87,25 @@ def github_import=(import)
     @github_import = import
   end
 
+  %i(opted_in_newsletter opted_in_announcements opted_in_marketing_announcements opted_in_surveys opted_in_sponsorships opted_in_applications_open).each do |attribute|
+    define_method attribute.to_sym do
+      send("#{attribute}_at").present?
```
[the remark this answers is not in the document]
klappradla: Aside that, if you're just interested in whether or not the attribute has been set on the model, ActiveRecord gives you what you want to do for free as far as I can remember 🤔 

You can append a `?` to every attribute and have it report whether the attribute is NULL or not, like this:
```ruby
model.opted_in_thing_at?
```

app/models/user.rb
```diff
@@ -87,6 +87,25 @@ def github_import=(import)
...
+    end
+
+    define_method "#{attribute}?".to_sym do
```
klappradla: From what I remember, this is fine to be a string, no need for a `to_sym` (same for the other methods).

app/views/users/edit.html.slim
```diff
@@ -57,6 +57,43 @@
   = f.input :company_name, label: 'Company name'
   = f.input :company_info, label: 'Company info', hint: 'Where is the office, how much space can you provide, etc.'
   = f.input :bio, label: 'About you', hint: 'You can use basic <a href="http://en.wikipedia.org/wiki/Markdown" target="_blank">Markdown</a> here.'.html_safe
+  h3.page-header Communication
+  .checkbox
+    label
+      = f.check_box :opted_in_newsletter, checked: @user.opted_in_newsletter.present?
```
klappradla: I fear I don't really understand this approach 🙈 
Didn't you define `opted_in_newsletter` on the `user` model just before? So why do you need to duplicate the internals of the method again in the view? I thought the whole point in defining all those methods was also to be able to have the form be simple as:
```ruby
= f.check_box :opted_in_newsletter
```
☝️ no manually setting it to `checked`.

Or is this just a leftover before refacting it to the final approach you chose?

The same question of course applies to all the other fields as well.

app/models/user.rb
```diff
@@ -87,6 +87,25 @@ def github_import=(import)
...
+      send(attribute)
+    end
+
+    define_method "#{attribute}=".to_sym do |value|
```
klappradla: Please also add model-level specs for this 🙏 
If you decide to put it into a concern, you could also put the specs into _"not-the-user-model-specs"_ so things stay not to polluted there either 😉

spec/controllers/users_controller_spec.rb
```diff
@@ -120,34 +120,78 @@
...
 
-      describe "with invalid params" do
-        it "assigns the user as @user" do
-          allow_any_instance_of(User).to receive(:save).and_return(false)
-          put :update, params: { id: user.to_param, user: { name: 'invalid value' } }
-          expect(assigns(:user)).to eq(user)
-        end
+        context "communication opt-in" do
+          before do
+            Timecop.freeze(Time.new(2002, 10, 31))
```
klappradla: Personal preference, but I'd suggest to just save the `Time.now(etc)` thing into a variable and reuse it later. Just to make sure you're actually talking about the same thing:
```ruby
let(:now) { Time.now(2002, 10, 31) }

before { Timecop.freeze(now) }

after { Timecop.return)

# some thing

it 'some spec' do
  expect(user.attribute).to eq(now)
end
```

spec/controllers/users_controller_spec.rb
```diff
@@ -120,34 +120,78 @@
...
+            it_behaves_like 'tracks opt-out', :opted_in_sponsorships
+            it_behaves_like 'tracks opt-out', :opted_in_applications_open
+          end
         end
       end
+    end
 
-      context "another user's profile" do
-        let!(:another_user) { create(:user) }
+    describe "with invalid params" do
```
klappradla: Minor thing, but this is clearly a `context` I'd say 😉

spec/controllers/users_controller_spec.rb
```diff
@@ -120,34 +120,78 @@
...
+      it "assigns the user as @user" do
+        allow_any_instance_of(User).to receive(:save).and_return(false)
+        put :update, params: { id: user.to_param, user: { name: 'invalid value' } }
+        expect(assigns(:user)).to eq(user)
```
klappradla: Testing against `assigns` is deprecated since Rails 5 and also no longer the way people want to test Rails applications 😿 Therefore, since the app will probably be maintained for some more time, let's not add new things and base them on already deprecated API.

However, I also don't understand the point of the test anyways I think 🙈 What do you want to test? Is there any **outcome** of the request that you're interested in? Maybe the page shows the name of the user, or the page shows the form again, etc? If so, please test this.
If this is to make sure that the controller assigns the right user: just remove the test and ensure this quality of the controller by another test that actually checks the outcome of a request 😉

spec/controllers/users_controller_spec.rb
```diff
@@ -120,34 +120,78 @@
...
+      end
 
-        it "does not update the requested user" do
-          expect_any_instance_of(User).not_to receive(:update_attributes)
-          put :update, params: { id: another_user.to_param, user: { name: 'Trung Le' } }
-        end
+      it "re-renders the 'edit' template" do
+        allow_any_instance_of(User).to receive(:save).and_return(false)
```
klappradla: Personal preference, but I'd suggest to not use `allow_any_instance_of` for cases where it's easy to just have _a real world scenario_. So if you just update the params to result in a validation error, things are even tested **better**, since you don't have to know the internals of the controller, such as if it calls `save` or `save!` _(for which this test would fail btw.). So maybe just leave the name blank or something like this and get rid of the `allow_any_instance` bit.

spec/controllers/users_controller_spec.rb
```diff
@@ -120,34 +120,78 @@
...
 
-        it "redirects the user to the homepage" do
-          put :update, params: { id: another_user.to_param, user: valid_attributes }
-          expect(response).to redirect_to(root_url)
-        end
+    context "another user's profile" do
+      let!(:another_user) { create(:user) }
+
+      it "does not update the requested user" do
+        expect_any_instance_of(User).not_to receive(:update_attributes)
```
klappradla: Same here: let's not use these `any_instance` things 😉 
You could for instance instead just check for the changes:
```ruby
expect { request }.not_to change(other_user, :name)
```

Plus you could probably also check for the response being a 404 or something.

spec/controllers/users_controller_spec.rb
```diff
@@ -120,34 +120,78 @@
...
+        put :update, params: { id: another_user.to_param, user: { name: 'Trung Le' } }
+      end
+
+      it "redirects the user to the homepage" do
+        put :update, params: { id: another_user.to_param, user: valid_attributes }
+        expect(response).to redirect_to(root_url)
```
klappradla: I would move this part into the example above and also add a test to check that the response body contains the string from the notice message (something like "you're not allowed here").
On their own, these expectations don't provide that much value.

spec/controllers/users_controller_spec.rb
ramonh: Whoa. No idea how this got in. Wonder if I rebased something wrong....

ramonh:
THank you so much for the detailed feedback!

I'll get right to it 💪

spec/controllers/users_controller_spec.rb
ramonh: There's a few points here that are showing up as new changes, when frankly I haven't touched these lines. Would you recommend removing them all around?

ramonh:
Took some work, but I think it's good to go!

@klappradla Could you take another look, please? Thank you so much for the feedback!

app/controllers/users_controller.rb
```diff
@@ -103,6 +103,7 @@ def user_params
       :is_company, :company_name, :company_info,
       :application_about, :application_motivation, :application_gender_identification, :application_age,
       :application_coding_level, :application_community_engagement, :application_language_learning_period,
+      :opted_in_newsletter, :opted_in_announcements, :opted_in_marketing_announcements, :opted_in_surveys, :opted_in_sponsorships, :opted_in_applications_open,
```
klappradla: Just a suggestion: to keep the two lists in sync, why not just pull them directly from where they are defined?
```ruby
*EmailPreferences::ATTRIBUTES
```

app/models/user.rb
```diff
@@ -2,6 +2,7 @@
 require 'github/user'
 
 class User < ApplicationRecord
+  include EmailPreferences
```
klappradla: Let's leave a blank like after this.

spec/controllers/users_controller_spec.rb
```diff
@@ -85,6 +85,50 @@
 
     context "its own profile" do
       describe "with valid params" do
+        context "communication opt-in" do
+          let(:now) { Time.new(2002, 10, 31) }
+          before do
```
klappradla: Only syntactic things, but I'd suggest to use the one line syntax here for these:
```ruby
before { Timecoop.freeze(now) }

after { Timecop.return }
```
Just because the file probably is already super huge...

app/models/concerns/email_preferences.rb
```diff
@@ -0,0 +1,33 @@
...
+    :opted_in_announcements,
+    :opted_in_marketing_announcements,
+    :opted_in_surveys,
+    :opted_in_sponsorships,
+    :opted_in_applications_open
+  ]
+
+  ATTRIBUTES.each do |attribute|
+    define_method attribute do
+      send("#{attribute}_at?")
```
klappradla: Please use `public_send` here.

app/models/concerns/email_preferences.rb
```diff
@@ -0,0 +1,33 @@
...
+    end
+
+    define_method "#{attribute}?" do
+      send(attribute)
```
klappradla: Same here: this can be `public_send`

spec/controllers/users_controller_spec.rb
```diff
@@ -85,6 +85,50 @@
...
+          end
+
+          after do
+            Timecop.return
+          end
+
+          shared_examples_for 'tracks opt-in time' do |attribute|
+            it "sets #{attribute} to the current time" do
+              put :update, params: { id: user.to_param, user: { attribute => '1' } }
+              expect(user.reload.send "#{attribute}_at").to eq(now)
```
klappradla: Let's use `public_send` here.

spec/controllers/users_controller_spec.rb
```diff
@@ -85,6 +85,50 @@
...
+
+          context 'opting out' do
+            shared_examples_for 'tracks opt-out' do |attribute|
+              before do
+                user.update_attribute attribute, Time.new(2002, 10, 31)
+              end
+
+              it "sets #{attribute} to nil" do
+                put :update, params: { id: user.to_param, user: { attribute => '0' } }
+                expect(user.reload.send "#{attribute}_at").to eq(nil)
```
klappradla: Let's use `public_send` here.

spec/features/users/opt_in_spec.rb
```diff
@@ -0,0 +1,41 @@
+require 'rails_helper'
+
+RSpec.describe 'Email opt in', type: :feature do
+  let(:user)      { create(:user) }
```
klappradla: Spacegame 👾

spec/features/users/opt_in_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  context 'signed in' do
+    before { sign_in user }
+    after { sign_out user }
```
klappradla: That's not necessary, there is no state persisted.

spec/features/users/opt_in_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+
+      it 'is opted out of all email preferences by default' do
+        expect(page.find('#user_opted_in_newsletter')).to_not be_checked
+        expect(page.find('#user_opted_in_announcements')).to_not be_checked
+        expect(page.find('#user_opted_in_marketing_announcements')).to_not be_checked
+        expect(page.find('#user_opted_in_surveys')).to_not be_checked
+        expect(page.find('#user_opted_in_sponsorships')).to_not be_checked
+        expect(page.find('#user_opted_in_applications_open')).to_not be_checked
+      end
+
```
klappradla: Please remove this extra white line.

spec/features/users/opt_in_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+  context 'signed in' do
+    before { sign_in user }
+    after { sign_out user }
+
+    context 'in the user edit page' do
+      before { visit edit_user_path(user) }
+
+      shared_examples 'opting in' do |optin|
+        it 'allows opting in' do
```
klappradla: I may also be too paranoid here, but these tests look a bit too fine granular to me. They only check both extremes: no boxes are being checked and **a single one is checked**. For the latter scenario, the tests would also still be green if checking a single one would also automatically check all of the others 🤔 

So maybe a good trade of would just be a single long example instead of the shared one? I could for instance look like this:
```ruby
it 'allows opting in and out of email preferences' do
  # check that by default none are checked

  # check one or two and submit

  # navigate back to the page and check that the one (or two) are checked and all others aren't

  # opt out of one of the checked ones again

  # navigate back and check that the one you opted out is no longer checked
end
```

ramonh:
On it! Thanks @klappradla

ramonh:
Phew! Should be good to go now. Thanks for the feedback! Could you please take another look @klappradla ?

ramonh:
Thanks again @klappradla! I think I got them all. Wdyt?

klappradla:
💯 will have a look tonight @ramonh (cannot squeeze in the time right now 😞 )

ramonh:
No stress, @klappradla! I really appreciate you doing this in your spare time :v:

klappradla:
Awesome job @ramonh 😻 

Let's ship this ⛵️

ramonh:
YAY! Thank you @klappradla

PR #1022, opened 2018-07-06
[#892] Migrate `team.project_name` to `team.project_id`

### LINKED ISSUE #892: Migrate `team.project_name` to `team.project_id`
Related to #816 Add team to projects page

#816 restored the association between team and the project they are working on.
There is one task left: running the [rake task](  desc '2017-08-15: Change project_name to project_id') to migrate  the old `team.project_name` to the new `team.project_id`.

I am having cold feet for my first encounter with the production database. Who wants to pair with me?

klappradla:
This is supposed to fix #892 : replaying `team#project_name` with an actual _(already existing but not used)_ association between `project` and `team`. 

This includes:
- migrating existing data
- enabling orga to choose the project for a team
  _(one can only a project from the same season as the team)_
- ⚠️ no longer allow teams to change their project
  _(not sure if this was allowed on purpose before - it does not really make sense to me)_
- some new settings for our headless Chrome driver 🤡 
  _(we have to set the window size to something closer to real-world -> otherwise bootstrap will hide certain elements and make expectations harder..)_

I'm pretty sure this will also require some more changes to where certain links are displayed, etc. But for now I think this simply migrates the existing behavior to a new datamodel.

One potential 💣 I found on the way: the `HasSeason` concern does not **require** things to have a season... However, I think in most cases this should actually be enforced in order to not break half of the queries we're doing.
☝️ If anyone knows reason why this was not required - please tell me 🙏

### DIFF
--- a/app/controllers/teams_controller.rb
+++ b/app/controllers/teams_controller.rb
@@ -87,7 +87,6 @@ class TeamsController < ApplicationController
       :name, :twitter_handle, :github_handle, :description, :post_info, :event_id,
       :checked, :'starts_on(1i)', :'starts_on(2i)', :'starts_on(3i)',
       :'finishes_on(1i)', :'finishes_on(2i)', :'finishes_on(3i)', :invisible,
-      :project_name,
       roles_attributes: role_attributes_list,
       conference_preference_attributes: [:id, :terms_of_ticket, :terms_of_travel, :first_conference_id, :second_conference_id, :lightning_talk, :comment, :_destroy],
       sources_attributes: [:id, :kind, :url, :_destroy]
--- a/app/views/teams/_form.html.slim
+++ b/app/views/teams/_form.html.slim
@@ -39,7 +39,6 @@
     = f.link_to_add 'Add another member', :roles, class: 'btn btn-primary col-sm-offset-3'
 
   - if f.object.accepted?
-    = f.input :project_name, placeholder: 'Name of your project', label: 'Project Name'
     = f.input :github_handle, placeholder: '', hint: 'Should your team have an organization'
     = f.input :twitter_handle, placeholder: '@', hint: 'Should your team have one'
     = f.input :description, label: 'Introduction', placeholder: 'Tell the world about your team (eg. about the project you will be working on)', input_html: { rows: 10 }, hint: "This is just for your public team profile. You can use basic #{link_to 'Markdown', "https://en.wikipedia.org/wiki/Markdown", target: :blank} here.".html_safe
--- a/app/views/teams/_form.html.slim
+++ b/app/views/teams/_form.html.slim
@@ -39,6 +39,7 @@
     = f.link_to_add 'Add another member', :roles, class: 'btn btn-primary col-sm-offset-3'
 
   - if f.object.accepted?
+    = f.association :project, readonly: true, include_blank: false
     = f.input :github_handle, placeholder: '', hint: 'Should your team have an organization'
     = f.input :twitter_handle, placeholder: '@', hint: 'Should your team have one'
     = f.input :description, label: 'Introduction', placeholder: 'Tell the world about your team (eg. about the project you will be working on)', input_html: { rows: 10 }, hint: "This is just for your public team profile. You can use basic #{link_to 'Markdown', "https://en.wikipedia.org/wiki/Markdown", target: :blank} here.".html_safe
--- a/app/views/teams/_form.html.slim
+++ b/app/views/teams/_form.html.slim
@@ -39,7 +39,7 @@
     = f.link_to_add 'Add another member', :roles, class: 'btn btn-primary col-sm-offset-3'
 
   - if f.object.accepted?
-    = f.association :project, readonly: true, include_blank: false
+    = f.association :project, disabled: true, include_blank: false
     = f.input :github_handle, placeholder: '', hint: 'Should your team have an organization'
     = f.input :twitter_handle, placeholder: '@', hint: 'Should your team have one'
     = f.input :description, label: 'Introduction', placeholder: 'Tell the world about your team (eg. about the project you will be working on)', input_html: { rows: 10 }, hint: "This is just for your public team profile. You can use basic #{link_to 'Markdown', "https://en.wikipedia.org/wiki/Markdown", target: :blank} here.".html_safe
--- a/spec/features/organizers/teams/edit_spec.rb
+++ b/spec/features/organizers/teams/edit_spec.rb
@@ -1,0 +1,46 @@
+require 'rails_helper'
+
+RSpec.describe 'Editing Teams', type: :feature do
+  let(:user)            { create(:organizer) }
+  let(:season)          { Season.current }
+  let(:other_season)    { create(:season) }
+  let(:team)            { create(:team, name: 'Team Yay', season: season) }
+  let(:selection_phase) { season.applications_close_at + 1.day }
+  let!(:exercism)       { create(:project, :accepted, name: 'Exercism', season: season) }
+
+  before do
+    create(:project, :accepted, name: 'Some Old Project', season: other_season)
+    create(:project, name: 'Non-Accepted', season: season)
+    create(:project, :accepted, name: 'Open Source Dropbox', season: season)
+
+    Timecop.travel(selection_phase)
+  end
+
+  after { Timecop.return }
+
+  it 'allows to set the project for a team (from the same season)' do
+    login_as user
+    visit edit_organizers_team_path(team)
+
+    expect(page).not_to have_content('Some Old Project')
+    expect(page).not_to have_content('Non-Accepted')
+    expect(team.project).to be_nil
+
+    select 'Exercism', from: 'Project'
+    fill_in 'URL', with: 'https://example.com'
+    click_on 'Save'
+
+    expect(page).to have_content('Team was successfully updated')
+    expect(page).to have_content('https://example.com')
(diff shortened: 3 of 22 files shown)


spec/features/users/guest_user_spec.rb
```diff
@@ -1,16 +1,14 @@
...
 RSpec.describe 'Guest User', type: :feature do
-
-  let!(:activity) { create(:status_update, :published, team: team1) }
-  let(:other_user) { create(:user) }
-  let!(:project)       { create(:project, :in_current_season, :accepted, submitter: other_user) }
-  let!(:team1)         { create(:team, :in_current_season, name: 'Cheesy forever', project_name: project.name, project_id: project.id) }
+  let!(:activity)       { create(:status_update, :published, team: team1) }
+  let(:other_user)      { create(:user) }
+  let!(:project)        { create(:project, :in_current_season, :accepted, submitter: other_user) }
+  let!(:team1)          { create(:team, :in_current_season, name: 'Cheesy forever', project: project, project_id: project.id) }
```
F3PiX: Do we need the project_id here?

spec/models/project_spec.rb
```diff
@@ -12,13 +12,46 @@
...
 
-  context 'with validations' do
+  describe 'validations' do
     it { is_expected.to validate_presence_of(:name) }
     it { is_expected.to validate_presence_of(:submitter) }
     it { is_expected.to validate_presence_of(:mentor_email) }
   end
 
-  context 'with callbacks' do
+  describe 'scopes' do
```
F3PiX: 😍 If only I could write specs like you do...

spec/models/team_spec.rb
```diff
@@ -422,42 +422,46 @@
...
+      it 'uses only the project name if the team name is missing' do
+        expect(display_name).to eq('Team Sinatra')
       end
     end
 
-    it 'will not prepend "Team" if already in place' do
-      subject.name = "Team Three Headed Monkey"
-      expect(subject.display_name).to eql "Team Three Headed Monkey"
+    context 'when from a past season' do
+      let(:team)   { build(:team, name: 'Cheesy', season: season, project: project) }
```
F3PiX: 🧀

spec/models/team_spec.rb
```diff
@@ -422,42 +422,46 @@
...
-        expect(subject.display_name).to eql 'Team Sinatra'
-      end
+    let(:team)    { build(:team, :in_current_season, name: 'Blue') }
+    let(:project) { build(:project, name: 'Sinatra') }
 
-      it 'returns "Team Blue" if name given' do
-        subject.name = 'Blue'
-        expect(subject.display_name).to eql 'Team Blue (Sinatra)'
-      end
+    it 'returns "Team" and the name' do
```
F3PiX: I think it would be even more clear if you put this spec together with the Team Team name spec (on line 434) in one context together. 
Also, I like it when specs start with the 'complete happy path', and the special cases unfolding afterwards. In this case: the display_name with team name, current season and project name. I think it would be an easier read here.

app/views/teams/_form.html.slim
```diff
@@ -39,7 +39,6 @@
     = f.link_to_add 'Add another member', :roles, class: 'btn btn-primary col-sm-offset-3'
 
   - if f.object.accepted?
-    = f.input :project_name, placeholder: 'Name of your project', label: 'Project Name'
```
F3PiX: If you like, you can still show the project name here, with something like: `f.object.project`
(So that the user doesn't need to go and find the correct Project Name, when they are telling something about the project in the `description field`)

klappradla:
Thanx for your comments @F3PiX 🙏 - I hope I can find the time to look into this tomorrow 🤓 

And **HUGE THANKS** to pointing out the problem with the review apps 💯 _(I totally forgot about them)_. I guess the "original" problem is that we once purged the migrations directory and Heroku by default tries to run them all. So even if I fix the one you're mentioning it will crash for the next one that is e.g. operating on a table that should have been created in a previous one 🙈 

Not sure if I can really fix this now... I guess it should be somehow configurable at Heroku 🤔 I'll have a look.

app/controllers/teams_controller.rb
```diff
@@ -87,7 +87,6 @@ def team_params
       :name, :twitter_handle, :github_handle, :description, :post_info, :event_id,
       :checked, :'starts_on(1i)', :'starts_on(2i)', :'starts_on(3i)',
       :'finishes_on(1i)', :'finishes_on(2i)', :'finishes_on(3i)', :invisible,
-      :project_name,
```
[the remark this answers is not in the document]
klappradla: Exactly ✅ 

I don't know how the `project_id` param (and the field in the form) actually ended up in the `TeamsController`. It would basically allow teams to change the project they're working on once they have been selected 😹

spec/features/users/guest_user_spec.rb
klappradla: Or course not, thanx! 🙏

spec/models/team_spec.rb
klappradla: Ok. Do you mean putting together as in merging both into one example, or just positioning them next to each other?

app/views/teams/_form.html.slim
klappradla: Yes, thought about this already.

spec/models/team_spec.rb
F3PiX: I meant positioning them next to each other, within one context, like  `context "with a team name"`

app/controllers/teams_controller.rb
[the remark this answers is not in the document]
F3PiX: LOL Probably I am to blame for that, when I changed `project_name` to `project_id` 🙈

F3PiX:
Closing and reopening to rebuilt the review app

klappradla:
For the review apps thing: you can also just delete and rebuild in the Heroku UI. I've seen the no permission error but didn't have a direction yet on how to fix that. Will hopefully have more ideas on that tonight 💡

F3PiX:
I wouldn't mind finishing this PR first, and solve the review app stuff in a new PR.

spec/models/team_spec.rb
klappradla: Ok. The `team` model has a presence validation on the `name` attribute, so a context like _"without a name"_ seems obsolete anyways. So I'll just remove that one.

klappradla:
Ok @F3PiX, I removed all the commits related to review apps (I didn't fix them anyways) and quickly threw in the `project` into the teams edit form as a simple readonly field.

It's only displayed for team who already got accepted.

Looks like this for teams who have been assigned to a project 👇 
screen shot 2018-07-10 at 00 03 10

And just to double check that it does not break if teams are accepted and **don't have a project** 👇 
screen shot 2018-07-10 at 00 00 55
_(it just does not display anything)_

app/views/teams/_form.html.slim
F3PiX: The project doesn't need to be submitted, right? `readonly` does submit, `disabled` doesn't. Also, `disabled` will make sure the field is skipped when navigating with tabbing. Looks like a more appropriate choice here. What do you think?

spec/features/organizers/teams/edit_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+  let(:other_season)    { create(:season) }
+  let(:team)            { create(:team, name: 'Team Yay', season: season) }
+  let(:selection_phase) { season.applications_close_at + 1.day }
+  let!(:exercism)       { create(:project, :accepted, name: 'Exercism', season: season) }
+
+  before do
+    create(:project, :accepted, name: 'Some Old Project', season: other_season)
+    create(:project, name: 'Non-Accepted', season: season)
+    create(:project, :accepted, name: 'Open Source Dropbox', season: season)
+
```
F3PiX: Not a review but an inquiry, to get better at writing specs: I'd think that you can use `build_stubbed` here instead of `create`? Just ignore me if that's incorrect.

spec/models/team_spec.rb
F3PiX: 🤔  I think this is about two tests about how a team name is being handled for display: the handling will be different for  "Team Cheesy" or "Cheesy".  
That's two valid inputs for the display method, so I thought it makes sense to have them both AND put them in one context together. No?

klappradla: I may still be missing the point, sorry 😹 🙈 

I have both tests in place, but I don't understand why I should put them in the same context. They have a different setup (they need two different teams) - so there's no point to put them in the same context. Or do you mean something different?

spec/features/organizers/teams/edit_spec.rb
klappradla: It's super rare that one needs or can use `build_stubbed` - I personally wouldn't even include it into `FactoryBot` if I were the author 😉 

Here, since we are in a "feature" spec which uses the app _"end to end"_ the controller is pulling stuff out of the database so things actually need to be persisted - therefore the `create`.

F3PiX: O thanks for the great explanation! And I'll stop looking at uses for build_stubbed starting NOW.

klappradla: Haha, well it probably has its usecases somehow. But I usually could archive anything I wanted with `build` alone.

klappradla:
Thanx for you comments @F3PiX 🙏

PR #1023, opened 2018-07-09
Setup postdeploy script for review apps

ramonh:
Adapted from [this](https://devcenter.heroku.com/articles/github-integration-review-apps#the-postdeploy-script), this should fix the crashes from review apps

### DIFF
--- a/app.json
+++ b/app.json
@@ -1,6 +1,7 @@
 {
   "name": "rgsoc-teams",
   "scripts": {
+    "postdeploy": "bundle exec rake db:schema:load db:seed"
   },
   "env": {
     "AWS_ACCESS_KEY_ID": {


klappradla:
It's unfortunately still not working... 😿 I already tried the same over at #1022 ...

I think there's more wrong with the overall setup of the postgres addon.

ramonh:
Hmm, that's a shame :( I'll continue to investigate. Thanks for confirming!

ramonh:
Hey @klappradla good news! I found [here](https://devcenter.heroku.com/articles/github-integration-review-apps#the-postdeploy-script). That the `postdeploy` script is only run once, and that to rerun it I had to close and re-open the PR. 

Lo and behold, it seems that it works!

F3PiX:
Yay!! Almost there. The OAuth url is not pointed at the review apps, so we can't log in yet.
schermafbeelding 2018-07-29 om 22 43 10

ramonh:
Hey all, after lots of poking around and trying to understand, the underlying issue is with OAuth itself.

Heroku is currently basing the review apps on the `rgsoc-teams-staging` app, which has its own OAuth callback.

We'd need to set up one callback URL per review app, which is pretty tricky to set up automatically. See the following link for some ideas:
https://twitter.com/searls/status/1012343568072105984

One idea is to set up a master domain for Oauth? I'm not sure.

Any thoughts on this? I'm a bit stuck.

klappradla:
Oh, that's tough 😞 

Just quickly dumping my brain about why we want the review apps, what we hope they'd solve and where they eventually fall short:
- we want a quick way to click and see UI changes / if the app still works ✅
  Should be possible with the review apps. However, having enough feature tests in place should also give us some safety here.
- we want an easier way to test drive PRs via the UI ❓
  Review apps may only partially help here. It may still be a large effort to get the review app in the right state (have the right data, phase, user, etc) in order to be able to test a particular scenario. So I have my doubts that it's going to be that helpful here.
- we want to spare the minutes of checking out the PR locally ❓
  Clicking through the review app, logging in, etc. may actually be close to the same effort to checking out a PR locally and using it with an existing DB.

Looking at this list ☝️ I'd say it's not worth the effort getting the review apps to work. The time may be better spent in better (feature-) test coverage of the app. Automated tests may be able to give us more guarantees not not break things on the long run, whereas the review apps only give us some hints for a particular moment and are again based on how thoroughly someone clicks through them.

ramonh:
Hey @klappradla thank you so much for your insight.

I'm highly inclined to agree on this. It seems that we'd have to manually add a URL to our OAuth settings which only the core organization team have access to. Bugging them every time we create a pull request seems very tedious.

I'd like to get others' opinions here as well as in the dev sync tonight, please!

Issue #895, opened 2017-12-14
Finetuning team selector on teams page


F3PiX:
Thanks to @mbrien12, we do have a nicely formatted dropdown selector #877 🎉 

Follow up: 
- display the years in order; 
- do not display the `succeeding` year in the dropdown list, unless the succeeding year is the current year (see convo here);
- Change the text on the button to 'Select by year'

mbrien12:
Can I have a go this week? ✋

F3PiX:
It's yours! We'll add you to the contributors, so we can assign the issue to you in the GH list.)
Thanks :-)))

mbrien12:
Will take a look over the weekend 👍

mbrien12:
Apologies haven't had time to look at this yet and now travelling over the Christmas break. Am back 10th of Jan so feel free to assign to somebody else over that time, otherwise can look at when I get back in the new year

F3PiX:
Thanks for the heads up and have a nice trip!

F3PiX:
This issue is available for new contributors.

agpiermarini:
Hi, @F3PiX. I could take a look at some of these issues if you would like.

F3PiX:
Hey @agpiermarini That would be supercool! Welcome to the Teams App 💟 
You can find the set up in the readme, and don't hesitate to ask for help here!! 
Also, I updated the description so that you can find the previous convo about the `succeeding year` showing up in the list. And reordered the steps. 
I'd recommend to start with the first step and take it from there! 
Have fun!

PR #1026, opened 2018-07-14
Fine-tune year selector on teams page (Issue #895)

agpiermarini:
Related issue #895 

Addresses the following items from the above-referenced issue:
- [x]   Change the text on the button to 'Select by year'
- [x]   Display the years in order
  - Add `order(:name)` clause to `Season.all_years` query
- [x]   Do not display the succeeding year in the dropdown list, unless the succeeding year is the current year
  - Add `where` clause to `Season.all_years` to filter out succeeding years that are not equal to the current year. `Season.all_years` is only invoked in the teams index, so it seemed more appropriate to revise the method versus create a new method. 
  - Add objects to `Season.all_years` test in  season_spec to ensure filtering functionality works. Update text in `it` block to more accurately describe what the method achieves.

**NOTE:** Tests were run locally. All are passing with the exception of `1) Season.transition? on New Year's should not be transition (#./spec/models/season_spec.rb:208)`, which I can confirm was failing prior to the above changes.

### DIFF
--- a/app/models/season.rb
+++ b/app/models/season.rb
@@ -39,7 +39,7 @@ class Season < ApplicationRecord
     end
 
     def all_years
-      pluck(:name)
+      where("name <= ?", Date.today.year.to_s).pluck(:name)    
     end
   end
 
--- a/app/models/season.rb
+++ b/app/models/season.rb
@@ -39,7 +39,7 @@ class Season < ApplicationRecord
     end
 
     def all_years
-      where("name <= ?", Date.today.year.to_s).pluck(:name)    
+      where("name <= ?", Date.today.year.to_s).order(:name).pluck(:name)    
     end
   end
 
--- a/spec/models/season_spec.rb
+++ b/spec/models/season_spec.rb
@@ -233,9 +233,11 @@ RSpec.describe Season, type: :model do
   end
 
   describe '.all_years' do
-    it 'return array of years' do
+    it 'returns array of years that includes succeeding year only if succeeding year is equal to current year' do
+      next_year = (Date.today.year + 1).to_s
       create :season, name: '2015'
       create :season, name: '2016'
+      create :season, name: next_year
       expect(Season.all_years).to match_array ['2015', '2016']
     end
   end
--- a/app/views/teams/index.html.slim
+++ b/app/views/teams/index.html.slim
@@ -8,7 +8,7 @@ nav.actions
       | Select by Year
       span.caret<
     ul.dropdown-menu role="menu"
-      - Season.all_years.sort.each do |year|
+      - Season.all_years.each do |year|
         li
          a data-behaviour="switch-teams-year" data-value=year =year
 
--- a/app/views/teams/index.html.slim
+++ b/app/views/teams/index.html.slim
@@ -5,10 +5,10 @@ nav.actions
     = 'Sign up to create your own team!' unless current_user
   div.pull-right.dropdown
     button.btn.btn-default.dropdown-toggle data-toggle="dropdown"
-      | Past Teams
+      | Select by Year
       span.caret<
     ul.dropdown-menu role="menu"
-      - Season.all_years.each do |year|
+      - Season.all_years.sort.each do |year|
         li
          a data-behaviour="switch-teams-year" data-value=year =year
 


spec/models/season_spec.rb
```diff
@@ -233,9 +233,11 @@
   end
 
   describe '.all_years' do
-    it 'return array of years' do
+    it 'returns array of years that includes succeeding year only if succeeding year is equal to current year' do
```
klappradla: Just a suggestion: it could make sense to also add a season for the _"current"_ year here - as this is something the query explicitly takes care of. Right now, I could go and change the `<=` to `<` without the tests breaking 😉

app/models/season.rb
```diff
@@ -39,7 +39,7 @@ def projects_proposable?
     end
 
     def all_years
-      pluck(:name)
+      where("name <= ?", Date.today.year.to_s).order(:name).pluck(:name)
```
F3PiX: Until Dec  31st of the current year, it will return all the years that have teams. As it should! So that's great. 
But on Jan 1st (and all the way until we actually have teams _in_ the _upcoming_ season), it will incorrectly show that upcoming as a choice. 
For example: the current season is '2018'. Next season is "2019". On 2019 Jan 1st this method will return 2019 as well, because `"2019" <= "2019"` is true.  But there are no teams for 2019, so '2019' should not show up in the selection. 

So, your solution works absolutely to include the current year 🎉 . The caveat is that it sometimes _shouldn't_ return the current year.

Are you up to the challenge to try to solve this? 
(Also, this means that the method name may not be the best choice...)
FYI: the readme mentions the Season Switch, that you can use to fake that we are in the 'proposal phase' of a season, and in that phase there are no teams for that season yet.

spec/models/season_spec.rb
F3PiX: Also: see my comment at the `all_year` method. You may want to add a spec for a transition season. There are some examples that use the different phases of a season, maybe you can use them for inspiration.

agpiermarini: Good catch! I will implement this.

app/models/season.rb
F3PiX: We have something like an `active` season. Out of my head, I'd say that's your best bet. Or at least it should give you the information you need. 👍

agpiermarini: Thanks!

agpiermarini:
@F3PiX @klappradla 

I pushed some updates to address your requested changes. Thanks again for the great feedback.

app/models/season.rb
```diff
@@ -38,8 +38,8 @@ def projects_proposable?
         season.project_proposals_close_at
     end
 
-    def all_years
-      pluck(:name)
+    def active_and_previous_years
+      where("acceptance_notification_at <= ?", Time.now.utc).order(:name).pluck(:name)
```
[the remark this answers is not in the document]
F3PiX: Hey @agpiermarini Great!! I love your solution using the general `acceptance_notification` date!!

One more thing, and sorry I missed that in the previous review. The `order` will return the season names as `oldest season first`, right? And it really should return the latest season first, = `:desc`.
Would you please add that? Either leave it in this query, or put it in a named scope `ordered_by_name` or something.

spec/models/season_spec.rb
```diff
@@ -232,11 +232,27 @@
...
+    end
+
+    it 'returns current year only if current season is active season' do
+      before_acceptance_notification = Time.utc(2018, 5, 1, 23, 59, 58)
+      after_acceptance_notification = Time.utc(2018, 5, 2, 0, 0, 0)
+      create :season, name: '2015'
+      create :season, name: '2016'
+      create :season, name: '2018'
+
+      Timecop.freeze(before_acceptance_notification)
```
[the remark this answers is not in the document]
F3PiX: Bonus points if the specs checks the right order of the years (see my previous comment).

F3PiX:
@agpiermarini Almost there!!

app/models/season.rb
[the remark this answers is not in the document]
agpiermarini: Absolutely. I'm inclined to leave it in the query because the reverse chronological ordering does not seem to be used in any other Season model queries, and it may be a bit more readable to have everything in one place. With that said, if your preference would be to scope the ordering operation, I'd be happy to do so.

spec/models/season_spec.rb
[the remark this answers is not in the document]
agpiermarini: @klappradla We should definitely avoid that! I went with the block syntax because my understanding is that to place the the two Timecop functions in before hooks, we would need to break out the spec into two separate specs (so as to have control of when the time is set in relation to the two assertions). On the hook approach: would it suffice to simply add an `after` hook with the `Timecop.return`? I'm happy to revisit this if your preference is to avoid the block syntax.

@F3PiX I will revise the spec to test for exact array ordering.

agpiermarini:
@F3PiX @klappradla I have pushed updates addressing each of your comments, and revised the original PR message.

agpiermarini:
@F3PiX Perhaps I misunderstood @klappradla's comment about `Time.current` being a safer way to test that condition over `Time.now.utc`. Was the comment specifically in reference to the preferred approach for ActiveRecord queries? 

I understand the difference to be that `.now` will default to the server's timezone, whereas `.current` will default to the time zone set in the Rails environment. My reason for switching all of the references in the Season model was to keep things consistent, as in reading up on the difference, I did not see anything that would imply it behaving differently depending on the method in which it is invoked.

F3PiX:
@agpiermarini I see.  I'd prefer to just change it in the method you are changing, and not everywhere in the Season model. But @klappradla 's vote in this is leading :-)
Although, in this particular case, with the config in this app, I doubt if it even makes a difference, TBH. So yeah, Max, your vote please :-)

klappradla:
🙈 sorry for causing confusion @F3PiX & @agpiermarini 🙉 

Just revert a7867c342715a9e48b32c10b0250ffc31ff39b69 and everything should be fine and we're good to go ✌️

agpiermarini:
Will do!

agpiermarini:
@F3PiX  @klappradla As requested, I reverted to Time.now.utc in the Season model. I'll keep an eye out for any additional feedback. Thanks for all your help so far

klappradla:
Looks good to me 👍 
Sorry for the extra hassles with `Time.now` 🙈

Feel free to merge once you've seen the changes @F3PiX 👋

agpiermarini:
Excellent! And no hassle at all. Thanks for the opportunity!

Issue #945, opened 2018-02-25
Remove "Time now" column from community page users' listing


AnkitaKhurana:
The ‘Time Now’ column of a few community members is not fetching any results in the Community page.

![image](https://user-images.githubusercontent.com/13145896/36641820-1ec087c0-1a5c-11e8-9a61-f8f715e81eb2.png)

AnkitaKhurana:
I would like to work on this issue.

klappradla:
Hey @AnkitaKhurana ✌️ 

Thanx for reporting 🙏 

The question here is: how could this be "fixed"? The column is based on the `timezone` attribute of a user record which is right now non-mandatory. See the user form for this 👇 
![artboard 1](https://user-images.githubusercontent.com/3491815/36644195-00d78146-1a57-11e8-9c41-cc4a17e02c3d.png)

I'm personally strongly against making more attributes of the user profile *mandatory* unless they **really** have a benefit in making the program easier to run / easier for the orga team or something similar. Otherwise, it just raises the bar for new users creating their accounts, e.g. potential coaches, which would be sad if we'd miss them.

Right now, the `country` is a mandatory attribute. Guessing a user's timezone based on her country won't work though. Plus: eventually people want to change their timezone when they're travelling.. 🤔 

I would be inclined to close this issue as something we cannot "fix" (or maybe don't need to fix 😉 ). Or what would be your take on this? Did you already have an idea in mind how to solve this?

klappradla:
**Edit:**
We just came up with an even better "fix™" for this issue @AnkitaKhurana 💡 

Let's just fully get rid of the `Time now` timezone column on the community table. It's not a super relevant information in this content and the table is already way to huge anyways.

Would you be interested in implementing this change?

carpodaster:
I agree that removing the column is the way to go, seeing how few people seem to fill out that info (and that's ok, it's not required). We can, however, add an HTML title attribute hover to the location cell that holds the user's current local time (if their timezone is set)

ramonh:
Hey @AnkitaKhurana, thanks again for reporting this issue! 

Are you still interested in handling this?

agpiermarini:
Hi @ramonh, @carpodaster and @klappradla. From the above conversation, it appears that this is still an open item. I am a new contributor and think that this would be a great first issue. Would you like me to try to take care of it over the next day or two?

klappradla:
Closed via #1027

PR #1027, opened 2018-07-14
Remove 'Time now' and 'Conferences' columns from Community user listings table (#945)

agpiermarini:
Related issue #945

Addresses the following items in the above-referenced issue:

- [x] Removes 'Time now' column from user listings table
- [x] Deletes previously commented-out 'Conferences' column from the user listings table

Additionally, as per the comments below, I have deleted the `time_for_user` helper method and associated spec, as it is no longer referenced in the application.

screen shot 2018-07-14 at 11 31 58 am

### DIFF
--- a/app/views/community/index.html.slim
+++ b/app/views/community/index.html.slim
@@ -22,8 +22,6 @@ table.users.table.table-striped.table-bordered.table-condensed.table-hover
       th = sortable :team
       th = sortable :country
       th = sortable :location, User.human_attribute_name(:location)
-      th Time now
-      / th Conferences
 
   tbody
     - @users.each do |user|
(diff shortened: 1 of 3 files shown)


app/views/community/index.html.slim
```diff
@@ -37,8 +35,6 @@ table.users.table.table-striped.table-bordered.table-condensed.table-hover
         td = link_to_user_roles(user)
         td = user.country
         td = user.location
-        td = time_for_user(user)
```
klappradla: I think™ with this column no longer in place we may even be able to fully get rid of the `time_for_user` view helper. Can you please double check that?

agpiermarini: Indeed, it was only referenced in this table. I will remove the helper method and associated spec.

agpiermarini:
@klappradla I pushed some updates addressing your comment, and updated the PR message accordingly.

klappradla:
Awesome @agpiermarini 💚

PR #1028, opened 2018-07-15
Delete Comments with User

klappradla:
This is another quick _"fix"_ vaguely related to GDPR things: remove a user's comments when they delete their account. While that, I made both associations (`user` and `commentable`) required on the `comment` model. Comments without these two things are useless for us.

I also realized that the `seeds.rb` file somehow hangs forever for me... I simplified it a bit without breaking changes.

While looking at the comments table, I also realized that the whole `Mentor::Comment` thing is eventually too much of hack and may not be fully compatible with Rails 5 🙈. I'll probably look into refactoring this again next... 😉

### DIFF
--- a/db/migrate/20180714153306_remove_comments_without_user.rb
+++ b/db/migrate/20180714153306_remove_comments_without_user.rb
@@ -1,0 +1,9 @@
+class RemoveCommentsWithoutUser < ActiveRecord::Migration[5.1]
+  def up
+    execute <<~SQL
+      DELETE
+      FROM comments
+      WHERE user_id NOT IN (SELECT id FROM users)
+    SQL
+  end
+end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -176,12 +176,10 @@ module ApplicationHelper
   end
 
   def user_for_comment(comment)
-    if comment.user.nil?
-      "Deleted user"
-    elsif comment.user == current_user
-      "You"
+    if comment.user == current_user
+      'You'
     elsif comment.user.admin?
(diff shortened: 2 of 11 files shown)


db/migrate/20180714153306_remove_comments_without_user.rb
```diff
@@ -0,0 +1,9 @@
+class RemoveCommentsWithoutUser < ActiveRecord::Migration[5.1]
+  def up
+    execute <<~SQL
+      DELETE
+      FROM comments
+      WHERE user_id NOT IN (SELECT id FROM users)
+    SQL
+  end
```
carpodaster: Can't check it our locally myself right now, but I'm curious: what happens to migrations that do not have a `down` method?

klappradla: You can still roll them back - which changes the schema version as normal - but they just don't do anything 🤷‍♂️

klappradla:
Good point @carpodaster 👍 
I will have a look into how much additional complexity this may add to the overall system and then decide if it's worth implementing it ✌️

klappradla:
Setting this to WIP until I have time to come back to it.

klappradla:
Ok, gave this another 🤔 and decided to just go with this 🐼 

My reasons:
- it looks like the use case of looking up old conversations is rather rare
- I want to reduce the complexity of the app to make it easier to operate and maintain
- we can add the "feature" of blanking comments once we realize it would make sense

So TL;DR: I just want the easiest solution possible first 😉

carpodaster:
Sounds good, thanks Max! :100:

Issue #1012, opened 2018-06-01
Update supervisor ability


F3PiX:
Related to PR #997 

In the abilities we still use the old method for granting supervisors access to their team members email addresses, even when hidden. 
With the recent changes, we don't need the old method anymore.

- [ ] Remove the old method (marked FIXME), add a new rule for the supervisor 
   `can :read_email` of users they supervise, even when those users want their email hidden

Related things, not conclusive:
- [ ] Update the hint on the Profile form, that currently doesn't tell that their supervisor will be able to see the email address.    
- [ ] Remove the 'signed_in?' conditional in the `supervisors/base_controller`, and replace with either `confirmed?`, or better: use Devise's authenticate_user!  if possible and relevant. 
- [ ] Doublecheck if it works for students who where supervised  in previous years (or should it?)
- [ ] Add and/or update feature spec. 

I am a supervisor of user1 and not user2
Both students have `hide_email =  true`

Current behaviour:  
When logged in as student1's supervisor, I can view user1's email only  👍 
However, this is done via view helpers, not with proper abilities. 
We shouldn't trust on view helpers to guard privacy sensitive data.

anubiskhan:
How's it going? This is my first time contributing to an open source project and I was hoping to help out with this issue. If there is someone else working on it already just let me know and I'll see if I can't help out somewhere else. Thanks!

anubiskhan:
Looks like it was already [fixed](https://github.com/rails-girls-summer-of-code/rgsoc-teams/commit/975877fd178a0ba54a133fd04844dba61d3f380e#diff-e9c04d21b8c4d777e4231c4c6ce5f8cb)

I'll see if there is another issue I can help out with

PR #1029, opened 2018-07-19
Issue 1012 - Supervisor unable to see team member email if hide_email: true

anubiskhan:
Related issue #1012

While this does result in a passing test, giving the supervisor the ability to see a team member's email regardless of hide_email flag being turned on, it also allows the supervisor to read all attributes for that team member. This is likely a result of me not fully understanding the codebase yet. I tried to emulate some of the other methodology/syntax present in the models/ability.rb file but was unsuccessful. The strangest part for me is that the 'other_user' (models/ability.rb:51-52) is only an email and not actually a User, hence the find_by_email. If this solution is unacceptable I would be happy to continue working on the issue until I have found the best option. Thanks!

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -46,6 +46,11 @@ class Ability
     end
 
     # supervisor
+    if user.supervisor?
+      can :read, :all, User do |other_user|
+        supervises?(User.find_by_email(other_user), user)
+      end
+    end
     # Use old code, see below
 
     # project submitter
@@ -67,9 +72,6 @@ class Ability
     # visibility of email address in user profile
     can :read_email, User, id: user.id if !user.hide_email?
     can :read_email, User if user.admin?
-    can :read_email, User do |other_user|
-      user.confirmed? && (supervises?(other_user, user) || !other_user.hide_email?)
-    end
     can :read, :users_info if user.admin? || user.supervisor?
 
 
--- a/spec/models/ability/team_member_spec.rb
+++ b/spec/models/ability/team_member_spec.rb
@@ -61,16 +61,12 @@ RSpec.describe Ability, type: :model do
       end
 
       context "when viewing user information" do
-
-        # FIXME see issue #1001
-        # The following specs should pass after fixing
         before do
           team_member.update!(hide_email: true)
           user_in_other_team.update!(hide_email: true)
         end
 
         it "can read email address of team members in current season, even when hidden" do
-          pending 'fails; To be solved with issue #1001'
           allow(subject).to receive(:supervises?).with(team_member, user).and_return true
           expect(subject).to be_able_to(:read, team_member.email)
         end


anubiskhan:
I ran the full test suite before and after making this change. All tests pass with the exception of one pending, and one failing test(spec/models/season_spec.rb:208) which was failing when I first cloned the repo.

spec/models/ability/team_member_spec.rb
```diff
@@ -61,16 +61,12 @@
...
-        # The following specs should pass after fixing
         before do
           team_member.update!(hide_email: true)
           user_in_other_team.update!(hide_email: true)
         end
 
         it "can read email address of team members in current season, even when hidden" do
-          pending 'fails; To be solved with issue #1001'
           allow(subject).to receive(:supervises?).with(team_member, user).and_return true
           expect(subject).to be_able_to(:read, team_member.email)
```
klappradla: I'm not entirely sure how this test was originally intended... But from what I remember, `CanCan` works on objects only, it's not fine-grained enough for attributes 😉 

So this should rather be:
```ruby
expect(subject).to be_able_to(:read_email, team_member)
```

And while we're at it, I think the spec would be easier to read and maintain if we'd use the actual subjet's name, like this:

```ruby
expect(ability).to be_able_to(:read_email, team_member)
```

Or, use the shorthand syntax:

```ruby
before { allow(ability).to receive(:supervises?).with(team_member, user).and_return(true) }

it { is_expected.to be_able_to(:read_email, team_member) }
```

app/models/ability.rb
```diff
@@ -46,6 +46,11 @@ def initialize(user)
     end
 
     # supervisor
+    if user.supervisor?
+      can :read, :all, User do |other_user|
```
klappradla: It looks like this is mixing up things a bit 🤔  From how I read the issue, we only need a new ability `read_email` here, to allow the supervisor to access a user's email address. The object passed it will be a user as well, not just an email address.

klappradla (changes requested):
Thanx for jumping on this @anubiskhan 💯 

I think the pending test was a bit misleading in telling what to implement 🙈 I left some comments, hope that makes things clearer for you.

If that does not help you, feel free to always ask!

F3PiX:
@klappradla @anubiskhan Great that you have this thing going! I think I can shed a light on some of the mysteries, but I won't be able to have a good look until somewhere this weekend.

F3PiX:
@anubiskhan I can see how this must be confusing... The issue was not written with a new contributor in mind. My bad, sorry! Let me give you a bit more context.

In previous PR's, we worked on rearranging the Abilities, and this supervisor's ability has not been updated yet. This is the behaviour we want to have:

1) For users who did not set their email address hidden: 
 - email address is available (in views) to all _confirmed_ other users (not for guest users).
2) For users who did set their email address hidden: 
- email addres is not available to any user, except:
  - admin users
  - the user herself (for now: you can make it only when that user herself is _confirmed_)
  - when user is a student in a current team: the team's supervisor

That is currently managed with the old code (marked FIXME):
https://github.com/rails-girls-summer-of-code/rgsoc-teams/blob/6bfd1b83c2f7f760ab07b33426423e3c968089e1/app/models/ability.rb#L66-L74
And also present in the feature tests for guest and confirmed users. 

The first part of this issue (= the first checkbox in the description of the issue) is about cleaning up this old code.

- [ ] As you probably saw, we have abilities for confirmed users, and the `can :read_mail` is already implemented there for other users.
- [ ] The rule to be able to see your own email should be added.  
- [ ] Check if the admin rules cover this `can :read_email` as well
- [ ] Add abilities for supervisor's to the `#team members` part (where you already added your code 👍 ) with the `can: read_email` rule. And the `can :read, :users_info`
You can use the existing method in  https://github.com/rails-girls-summer-of-code/rgsoc-teams/blob/6bfd1b83c2f7f760ab07b33426423e3c968089e1/app/models/ability.rb#L123 

With that in place, all the lines 66-74 (the code snippet above) can be deleted - not only the rule in line 70 - sorry, I can see how that was not clear in the description.    
  
With the test [update proposed by @klapraddla](https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/1029/files/5c530c6af4d83a300e927cbb46b67bbd2fa21f20#diff-e9c04d21b8c4d777e4231c4c6ce5f8cb) those test should reflect this behaviour. 

The `can :read_email` is a bit different than 'normal' abilities, as @klappradla said, because it is kind of a view helper. 
The view helper is used in the User#show view (the user's profile page) and the Community page: 
https://github.com/rails-girls-summer-of-code/rgsoc-teams/search?l=Slim&q=can%3F+%3Aread_email

There are some more steps in the issue 1012 description , but let's see if the tests pass with this first part first. ;-)
Unless you are totally discouraged, and want to take on a different issue. Either way, we are happy to help you out, and please let us know your thoughts!! 

Happy coding!

anubiskhan:
This looks great, I will take another swing at it. Thanks for all of the feedback!

PR #1030, opened 2018-07-29
Remove daily log reminder task

ramonh:
As discussed with the organization team, we've disabled the task to send out daily log reminders.

Since they're disabled, we can go ahead and remove them.

The reason for this is they're no longer required for part time students.


ramonh:
Whoa, good call, thank you, @pgaspar!

ramonh:
Should be done now!

klappradla:
If the decision was to get rid of the feature, ➕ on @F3PiX 's point `2`: please also remove everything related to the feature - we don't want to maintain dead code 😉 

If the idea was to just temporarily disable the feature for a few weeks or so - then I'd say there's no reason to even touch the rake task...

ramonh:
Hey all! Thanks for your feedback. 

So the plan is to remove the feature outright, so I'll remove the related code.

F3PiX:
@ramonh There are a few more things to take care of, for instance:
https://github.com/rails-girls-summer-of-code/rgsoc-teams/blob/master/app/models/team.rb#L43
https://github.com/rails-girls-summer-of-code/rgsocteams/blob/master/app/models/team_performance.rb#L5
Maybe more?

ramonh:
Thanks @F3PiX! I think I got all of them. What do you think?

spec/models/team_performance_spec.rb
```diff
@@ -66,14 +59,6 @@ def create_all_teams
...
       end
 
-      context 'for a team with older activities and older feedback' do
-        subject { TeamPerformance.new(team_both_outdated)}
-
-        it "signals green" do
-          expect(subject.evaluation).to eq(:green)
-        end
-      end
-
```
F3PiX: I am not sure if all this specs should be removed. I didn't dive deep, but I'd expect that we only want to remove the specs that were added  here: https://github.com/rails-girls-summer-of-code/rgsoc-teams/commit/28b932e4a4b7ad3cd8d7ca54a124dcb96dc4cb67#diff-1144034ef8126927cabe1a2a9195b521
Could you double check if we aren't losing usefull specs here?

spec/models/team_performance_spec.rb
```diff
@@ -186,14 +141,6 @@ def create_all_teams
...
```
F3PiX: This one, I am pretty sure it _should_ stay, because it is to ensure that outside of the season, the teams' performance is frozen to green.
I also begin to suspect that `team_both_outdated` is the same as either `team_nothing` or `team_both`. Which is a nice illustration that naming is hard :-) But hey, it was my first year in Ruby ;-)

spec/models/team_spec.rb
```diff
@@ -370,45 +370,6 @@
...
-      end
-
-      it 'does not include the teams with recent updates' do
-        team_with_recent_status_update
-        team_with_recent_feed_entry
-        expect(described_class.without_recent_log_update).not_to include(team_with_recent_status_update)
-        expect(described_class.without_recent_log_update).not_to include(team_with_recent_feed_entry)
-      end
-    end
   end
```
F3PiX: I'd suspect all the specs in this file are not about the Reminder Mail. Or is there a different reason to remove them?

ramonh:
Oof, good call, @F3PiX . I'll clean it up

ramonh:
Hey @F3PiX I restored them, thanks for catching this! Could you take another look please?

app/models/team.rb
```diff
@@ -40,6 +40,7 @@ class Team < ApplicationRecord
   before_create :set_number
   before_save :set_last_checked, if: :checked
 
+
```
F3PiX: This whiteline can go?

spec/models/team_spec.rb
```diff
@@ -370,47 +370,40 @@
...
+    end
+    it 'includes the team with only old updates' do
+      team_with_old
+      expect(described_class.without_recent_log_update).to include(team_with_old)
+    end
+    it 'does not include the teams with recent updates' do
+      team_with_recent_status_update
+      team_with_recent_feed_entry
+      expect(described_class.without_recent_log_update).not_to include(team_with_recent_status_update)
+      expect(described_class.without_recent_log_update).not_to include(team_with_recent_feed_entry)
```
F3PiX: Why did we lose all the whitelines? This makes it harder to read?

spec/models/team_performance_spec.rb
```diff
@@ -74,6 +82,13 @@ def create_all_teams
         end
       end
     end
+
+    describe ".teams_to_remind" do
+      it 'should not remind anyone' do
+        create_all_teams.each { |t| t.update(season: Season.current)}
+        expect(TeamPerformance.teams_to_remind).to match_array([])
+      end
```
F3PiX: This is for the removed behaviour, so I guess it'll pass ;-) but why keep it here?

spec/models/team_performance_spec.rb
```diff
@@ -115,8 +144,17 @@ def create_all_teams
...
     end
+
+    describe ".teams_to_remind" do
+      # create all the teams
+      it 'should remind some' do
+        create_all_teams.each { |t| t.update(season: Season.current)}
+        expect(TeamPerformance.teams_to_remind).to match_array([team_nothing, team_commented, team_both_outdated])
+      end
+    end
   end
```
F3PiX: Same as ^ ?

spec/models/team_performance_spec.rb
```diff
@@ -156,5 +202,12 @@ def create_all_teams
         end
       end
     end
+
+    describe ".teams_to_remind" do
+      it 'should not remind anyone' do
+        create_all_teams.each { |t| t.update(season: Season.current)}
+        expect(TeamPerformance.teams_to_remind).to match_array([])
+      end
+    end
```
F3PiX: Same?

app/models/team_performance.rb
```diff
@@ -2,6 +2,14 @@
...
 # Internal: calculates a Team's performance score for supervisor's dashboard
 
+  def self.teams_to_remind
+    if buffer_days?
+      Team.none
+    else
+      Team.by_season_phase.without_recent_log_update
+    end
+  end
+
```
F3PiX: Now I'm confused. You want to restore _all_ functionality?

ramonh:
Hi @F3PiX, I apologize, I must have really mucked things up.

So to restore the specs you gave me feedback on before, I needed to restore the functionality.

I'll clean this up and give it another go.

spec/models/team_spec.rb
ramonh: Not sure what's going on here, frankly.

ramonh: All I did was re-indent them so they're proper? Maybe I did a bad

ramonh: Arg, nevermind I found it. Thanks Maud!

ramonh:
Okay @F3PiX I'm sorry for confusing you.

I think I've cleaned everything up properly!

F3PiX:
Yep, looks look to me! Thanks!

ramonh:
Yay! Thanks for bearing with me @F3PiX

Issue #970, opened 2018-03-26
show people's names in the application review


metakermit:
Hi 👋 

just going through the applications (a bit late, sorry 😬 ) and thought that it would be handy if names and GitHub usernames of team members were also shown on the [applications page](https://teams.railsgirlssummerofcode.org/mentors/applications) and the application details page. Some of the people have already gotten in touch with us on GitHub and we know their names/usernames, but it's a bit hard to keep track of who's from what team, when all the team names are a bit abstract.

Cheers!

ramonh:
Hey @metakermit Sorry I didn't get around to this sooner.

So the reason we weren't showing this before is because the mentor review stage of application reviews was during the blind phase.

Now that we've done away with that, however, it makes complete sense to show these again, so thank you very much for the feedback!

I think this would also be a great issue for students to use to start familiarising themselves with the teams app, so I'll label it thusly.

metakermit:
Awesome, thanks Ramón and have a great weekend :)

ramonh:
Same to you, thanks Dražen!

klappradla:
Should be via #1031 and ready for you to use next year @metakermit 👋

metakermit:
Nice, thanks a lot 🙌

PR #1031, opened 2018-08-19
Show student names in application review

ramonh:
Related issue #970
This adds the student name as the header instead of saying "First" and "Second" student

### DIFF
--- a/app/views/mentors/applications/show.html.slim
+++ b/app/views/mentors/applications/show.html.slim
@@ -24,9 +24,9 @@ div.panel.panel-default
   div.panel-body
     p = application.project_plan
 
-h3 First Student
+h3= application.student0.name
 = render 'student', student: application.student0
-h3 Second Student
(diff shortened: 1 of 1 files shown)


ramonh:
Oh, woops, sorry for the false alarm. I'll look into why this is failing

ramonh:
Good to go! @F3PiX, @klappradla could you please take a look?

spec/features/mentors/applications/show_spec.rb
```diff
@@ -17,6 +17,8 @@
     visit mentors_application_path(application)
 
     expect(page).to have_content 'Team We could be Heroines'
+    expect(page).to have_content '(First Student)'
```
klappradla: I'd suggest to also expect the actual names here as well.

ramonh:
I did it, I think! Thanks for the feedback. @klappradla could you please take another look?

Issue #868, opened 2017-10-10
User profile: postal addresses


mkalininait:
RGSoC is an international program, and it's sometimes hard to parse postal addresses of people from other countries.

**Suggestion:**

User profile, edit form: instead of one input field "Postal address" implement several (* are required):
 - Recipient's name*
 - Company
 - Address Line 1*
 - Address Line 2
 - City*
 - State/Province/Other 
 - Postal code*
 - Country*
 - Phone number

nerdyvaishali:
I think Phone number should be a mandatory field as the whole point of asking this infromation is to send packages. For the privacy issues, may be this fields should be only visible to organizers [or a person who handles swag packaging].

carpodaster:
> RGSoC is an international program, and it's sometimes hard to parse postal addresses of people from other countries.

That's why I would think that a multi-line textarea is better than having individual fields. Every country formats postal address differently. If a user puts their address exactly the way how it must appear on the envelope, shouldn't we be fine?

alicetragedy:
@carpodaster yeah, I think that was the original thought behind the text area.

The issue with the packages is that we need to fill out custom forms for each of the non-EU packages, and it some cases it was difficult to fill out from a text block in a text area. We were unsure if “house behind the only bus stop in town” is an acceptable entry for “Street” 😂  (I'm obviously exaggerating, but some of the addresses were very cryptic and it was hard to tell if this was a regular local address or not, so we had to look that up manually and/or ask people, which was a bit of a time suck).

Also: I'm not sure if this has anything to do with the actual ticket, but last year for swag we looked into partly automatising the process, i.e. exporting the data, uploading it to the platform of the chosen shipping service (e.g. postal service, dhl, fedex, hermes), and having the packages be picked up rather than doing everything manually. It could be that these services ask for the data to be formatted in a certain way.

shebesabrina:
Hey @mkalininait, I wanted to see if I could help with this issue. It seems you're needing some help with mandatory fields including phone numbers, and addresses. As well as making sure you have everything needed for shipment anywhere in the world. Mind if I try giving this a shot?

anasofiapinho:
[...]

PR #1032, opened 2018-09-05
Postal addresses

shebesabrina:
Related issue #
Adds multiple fields for a shipping address and extracts the postal address as its own model.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -72,5 +72,6 @@ group :test do
   gem 'shoulda-matchers', git: 'https://github.com/thoughtbot/shoulda-matchers.git'
   gem 'simplecov', require: false
   gem 'timecop'
+  gem 'pry-rails'
   gem 'webmock'
 end
--- a/Gemfile
+++ b/Gemfile
@@ -68,6 +68,7 @@ group :test do
   gem 'rspec-html-matchers'
   gem 'rubocop', require: false
   gem 'selenium-webdriver'
+  gem 'chromedriver-helper'
   gem 'shoulda-matchers', git: 'https://github.com/thoughtbot/shoulda-matchers.git'
   gem 'simplecov', require: false
   gem 'timecop'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -48,6 +48,8 @@ GEM
       minitest (~> 5.1)
       tzinfo (~> 1.1)
     addressable (2.4.0)
+    archive-zip (0.11.0)
+      io-like (~> 0.3.0)
     arel (8.0.0)
     ast (2.4.0)
     autoprefixer-rails (8.5.0)
@@ -92,6 +94,9 @@ GEM
       launchy
     childprocess (0.9.0)
       ffi (~> 1.0, >= 1.0.11)
+    chromedriver-helper (1.2.0)
+      archive-zip (~> 0.10)
+      nokogiri (~> 1.8)
     coderay (1.1.2)
     concurrent-ruby (1.0.5)
     countries (2.1.4)
@@ -149,6 +154,7 @@ GEM
     i18n (1.0.1)
       concurrent-ruby (~> 1.0)
     i18n_data (0.8.0)
+    io-like (0.3.0)
     jmespath (1.4.0)
     jquery-rails (4.3.3)
       rails-dom-testing (>= 1, < 3)
@@ -396,6 +402,7 @@ DEPENDENCIES
   cancancan
   capybara (~> 2.18)
   capybara-screenshot
+  chromedriver-helper
   country_select
   database_cleaner
   devise
@@ -444,4 +451,4 @@ RUBY VERSION
    ruby 2.5.1p57
 
 BUNDLED WITH
-   1.16.1
+   1.16.4
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -228,6 +228,8 @@ GEM
     pry (0.11.3)
       coderay (~> 1.1.0)
       method_source (~> 0.9.0)
+    pry-rails (0.3.6)
+      pry (>= 0.10.4)
     puma (3.11.4)
     rack (2.0.5)
     rack-test (1.0.0)
@@ -421,6 +423,7 @@ DEPENDENCIES
   pg
   pretender
   pry
+  pry-rails
   puma
   rails (~> 5.1.6)
   rails-controller-testing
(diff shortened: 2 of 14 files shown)


klappradla:
👋 @shebesabrina thanx a ton for getting involved 💚 

Before I get into an actual review of the code: it would be super neat if you could just also add the screenshots you posted in #868 here to the PR. Just makes it easier for people reviewing it to get an idea about what's happening on the visual side.
Aside that: the tests for this PR are failing on Travis (our CI tool). You can just head over there and have a look at the output to see what's the problem.

Gemfile
```diff
@@ -68,8 +68,10 @@ group :test do
   gem 'rspec-html-matchers'
   gem 'rubocop', require: false
   gem 'selenium-webdriver'
+  gem 'chromedriver-helper'
```
klappradla: Can you please remove this gem again from the Gemfile? It does not match the Selenium / Chrome setup we use for the feature specs on this project and having a quick look at the CI output for this PR, it looks also like this is the cause for it being 🔴 for feature specs.

I know that this gem is an alternative way to install Chromedriver, but I have only had bad experience with it and in general don't like the idea of _"misusing"_ gems to install system dependencies 😉 - it usually turns out to be a shortcut that's not worth it after all.
If you ran into any problems with setting up the dependencies for the app, just check the `README` and especially our [`TROUBLESHOOTING` guide](https://github.com/rails-girls-summer-of-code/rgsoc-teams/blob/master/TROUBLESHOOTING.md#chromedriver), we have a dedicated section for Chromedriver there 📚

Gemfile
```diff
@@ -68,8 +68,10 @@ group :test do
...
   gem 'shoulda-matchers', git: 'https://github.com/thoughtbot/shoulda-matchers.git'
   gem 'simplecov', require: false
   gem 'timecop'
+  gem 'pry-rails'
```
klappradla: Can you also please remove this gem from the Gemfile? I know it's super handy and you're free to use `pry` when working on the app. We try to not have any _"unrelated"_ changes to the actual purpose of a PR (in your case all things addresses).

You can totally update the setup to suit your personal workflow while developing the app - just don't commit these changes when opening a PR.

Gemfile.lock
```diff
@@ -444,4 +454,4 @@ RUBY VERSION
    ruby 2.5.1p57
 
 BUNDLED WITH
-   1.16.1
+   1.16.4
```
klappradla: As explained in my comments above: as it's not really relevant for the actual changes in this PR, it would be best if you could just unstage the changes on the `Gemfile` and `Gemfile.lock`. You can still keep these changes on your local fork if they make working with the app easier for you - just exclude them from this PR ✌️

app/controllers/postal_addresses_controller.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+class PostalAddressesController < ApplicationController
```
klappradla: Are you planning to make use of this controller later? I may overlook things, but I don't see it being used right now 👀 
If you're not using it, best get rid of it. Any unused code is code we have to maintain and results in some work 😉

app/controllers/users_controller.rb
```diff
@@ -29,6 +29,7 @@ def create
   def update
     respond_to do |format|
       if @user.update_attributes(user_params)
+        @user.postal_addresses.create(address_params)
```
klappradla: Hm, doing so means I'd create a new `address` record every time I update my user profile. What's the benefit of this?
I may not understand the details of this, but I fear we'll just pollute our database with loads of `address` records then - even though we'd only need one per student 🙈 

I'd suggest a different approach: you could just allow the address attributes as `nested_attributes` on the `User` model and include them in the already permitted `user_params`. Doing so, we can stay with only one line like
```ruby
@user.update_attributes(user_params)
```
And just update (or create) a single address record per user.

app/controllers/users_controller.rb
```diff
@@ -106,8 +107,13 @@ def user_params
       *EmailPreferences::ATTRIBUTES,
       :application_learning_history, :application_skills, :application_code_samples,
       :application_location, :application_minimum_money, :application_money, :application_goals, :application_code_background,
+      :postal_addresses,
       interested_in: [],
       roles_attributes: [:id, :name, :team_id, :_destroy]
     )
   end
+
+  def address_params
```
klappradla: See my comment above in the `update` method: I think using Rails' `nested_attributes` would be a even better fit here. Just check the Rails guides on this, it's should be pretty well documented there. Ping me if you need any help or guidance 👋

app/models/user.rb
```diff
@@ -65,6 +65,7 @@ def student
   has_many :applications, through: :teams
   has_many :todos, dependent: :destroy
   has_many :comments, dependent: :destroy
+  has_many :postal_addresses
```
klappradla: Why `has_many`? I'd personally have gone for `has_one` 🤔 

I don't think we need more than one `address` per `student`.

app/views/users/edit.html.slim
```diff
@@ -107,7 +107,12 @@
   p.help-block This information will only be visible to yourself and the organizers.
   = f.input :tshirt_size, as: :select, collection: User::TSHIRT_SIZES.map { |k| [k, k] }, label: 'T-Shirt size', blank: false, required: false, hint: 'For sponsor T-Shirts, in case they send some.'
   = f.input :tshirt_cut, as: :select, collection: User::TSHIRT_CUTS.map { |s| [s, s] }, label: 'T-Shirt cut', include_blank: true
-  = f.input :postal_address, hint: "Please give your postal address, including your full name, so we can send things we've received from our sponsors for you :)"
+
+  h4 Shipping Address
+  = f.simple_fields_for :postal_addresses do |pa|
+    = pa.input :street
+    = pa.input :state
+    = pa.input :zip, required: true, hint: "Please give your postal address, so we can send things we've received from our sponsors for you :)"
```
klappradla: See my comments above about making use of `nested_attributes` here.
Aside that: why making only one of the fields required? And what about users who don't want to and don't need to provide an address? We only require the address from students, other users of the app are not required to provide this information. With making the field required like it is now, no user is able to update their profile without adding an address with a zipcode 😉 

Also: the hint does not make too much sense for the `zip` field alone. This looks more like a general hint we'd put above or below the address fields. I'd therefore remove it from the individual field and put it on the address fields in general.

PR #1034, opened 2018-10-17
Update dependencies

ramonh:
We're seeing a [warning](https://github.com/rails-girls-summer-of-code/rgsoc-teams/network/alerts) about security vulnerabilities.

This pull request should remedy that.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -132,7 +132,7 @@ GEM
       loofah (>= 2.0)
       sax-machine (>= 1.0)
     ffaker (2.9.0)
-    ffi (1.9.23)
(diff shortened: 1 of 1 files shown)


ramonh:
Thanks, @klappradla! It made the warning disappear so I think we're in the clear. (Rhyming not intended 😂 )

klappradla:
💚 I got an email the build is 🔴 though...

Could also be a flaky spec, maybe you wanna take a look and create an issue.

ramonh:
I saw that too! I think it is indeed a flaky spec, as rebuilding did the trick,

PR #1035, opened 2018-10-18
Fix minor typo

ramonh:
Unless I'm wrong, [analyzation](https://english.stackexchange.com/questions/314968/use-of-the-word-analyzation-in-formal-writing) isn't a correct use of the word. Should be analysis.

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,7 +192,7 @@ it 'is an interesting example', driver: :selenium_chrome do
 end
 ```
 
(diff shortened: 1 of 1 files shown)


klappradla:
🔴 ok, looks like the flakyness is more serious 😉

ramonh:
Oof, you're right.

Looking at Travis, the test that fails is pretty consistent. I'll create an issue for this

PR #1039, opened 2018-11-09
Fix flaky specs

klappradla:
This is supposed to prevent the flakyness visible in #1037. One can run into the failing specs with this seed:

    bundle exec rspec --seed 6251

The reason behind this: the traits in the `season` and `project` factories may result in colliding names for `seasons` -> which makes the record invalid.

To prevent this, I updated both factories, adapting the traits (removing an unused one).


klappradla:
_--- just rebasing on current `master`---_

PR #1040, opened 2018-11-09
Postal address fields

lilwillifo:
Related issue #868 

@klappradla 👋
This extracts Postal Addresses to their own DB backed model. 

A user can add, edit, or remove a postal address. If left blank, the profile will save, but if the address is partially filled out, it alerts them of any missing required fields there. 

I didn't get into address validation. Not sure if you want to pull in a GoogleMaps type gem to confirm that valid addresses are coming in, or if you're okay with relying on user input here. 

I haven't included phone number here as part of this, but that can be easily added if that's what's required. It felt a little strange to include a phone number in a postal address model for shipping, so I left it off for now. 

One note: there's one failing test. It's failing on master for me, so not sure if this is just a local issue for me or something your team is aware of.
1) Season.transition? on New Year's should not be transition
     Failure/Error: it { is_expected.not_to be_transition }
       expected `Season(id: integer, starts_at: date, ends_at: date, name: string, created_at: datetime, updated_at: d...otification_at: datetime, project_proposals_open_at: datetime, project_proposals_close_at: datetime).transition?` to return false, got true
      ./spec/models/season_spec.rb:208

Profile Page:
screen shot 2018-11-09 at 3 19 10 pm

Edit page: 
screen shot 2018-11-09 at 3 19 44 pm


spec/features/users/postal_address_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+        expect(page).to have_content address.address_line_1
+        expect(page).to have_content address.address_line_2
+        expect(page).to have_content address.city
+        expect(page).to have_content address.state_or_province
+        expect(page).to have_content address.postal_code
+        expect(page).to have_content user.postal_address.country
+      end
+
+      it 'does not allow a postal address to be added if required fields are missing' do
+        expect(user.postal_address).to be_nil
```
klappradla: Personal preference, but I'd say this is not really necessary for this test.

spec/features/users/postal_address_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+        fill_in 'user[postal_address_attributes][address_line_1]',    with: address.address_line_1
```
klappradla: See my comments on the previous test for how I'd suggest to format this.

spec/features/users/postal_address_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+        fill_in 'user[postal_address_attributes][postal_code]',       with: address.postal_code
+        click_on 'Save'
+
+        expect(page).to have_content "Postal address city can't be blank"
```
klappradla: 💚

spec/features/users/postal_address_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+        expect(page).to have_content "Postal address country can't be blank"
+      end
+
+      it 'autofills saved postal address info on edit form if it exists' do
+        user.update(postal_address: create(:postal_address))
```
klappradla: Personal preference, but I'd rather move this test into a `context`, maybe something like 
```ruby
context `when the user already has a postal_address set up` do
  let!(:postal_address) { create(:postal_address, user: user) }

  it ...
```

spec/features/users/postal_address_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+        visit edit_user_path(user)
+
+        expect(page).to have_selector("input[value='#{user.postal_address.address_line_1}']")
+        expect(page).to have_selector("input[value='#{user.postal_address.city}']")
+        expect(page).to have_selector("input[value='#{user.postal_address.state_or_province}']")
+        expect(page).to have_selector("input[value='#{user.postal_address.postal_code}']")
+      end
+
+      it 'lets you delete an existing postal address' do
```
klappradla: Same here, I'd rather use a context than setting up the address in the actual example.

Aside that, the `you` part does not really match the tone of the other tests here. How about something like
```ruby
it 'allows to delete an existing postal address' do
  # etc.
```

app/models/postal_address.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+class PostalAddress < ApplicationRecord
+  belongs_to :user
+  validates_presence_of :address_line_1, :city, :postal_code, :country
+
+  def formatted
+    "#{capitalize(address_line_1)} #{capitalize(address_line_2)} #{capitalize(city)}, #{state_or_province} #{postal_code} #{country}"
```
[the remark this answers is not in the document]
lilwillifo: I wanted to pull this logic out of the view, but I like your suggestion of using a partial instead. I'll make that change. And since I'm doing any data validation around field inputs, we need a formatter somewhere to make sure the capitalization is correct.

app/models/user.rb
```diff
@@ -73,6 +74,11 @@ def student
   validates :name, :email, :country, :location, presence: true, unless: :github_import
 
   accepts_nested_attributes_for :roles, allow_destroy: true
+  accepts_nested_attributes_for :postal_address, allow_destroy: true, reject_if: :all_blank
+
+  def postal_address_rejectable?(attr)
```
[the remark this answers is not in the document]
lilwillifo: good catch, thank you! I'll remove.

F3PiX:
The failing test has been solved by  @klappradla himself :-) in #1039 👏 🙇  
@lilwillifo That means that the tests should pass if you rebase your branch. Are you familiar with rebasing, if not, just ask and we can help you out.

app/views/users/edit.html.slim
```diff
@@ -107,7 +107,17 @@
   p.help-block This information will only be visible to yourself and the organizers.
   = f.input :tshirt_size, as: :select, collection: User::TSHIRT_SIZES.map { |k| [k, k] }, label: 'T-Shirt size', blank: false, required: false, hint: 'For sponsor T-Shirts, in case they send some.'
   = f.input :tshirt_cut, as: :select, collection: User::TSHIRT_CUTS.map { |s| [s, s] }, label: 'T-Shirt cut', include_blank: true
-  = f.input :postal_address, hint: "Please give your postal address, including your full name, so we can send things we've received from our sponsors for you :)"
+  h4 Shipping Address
+  - if @user.postal_address
+     = link_to "Remove Address", user_path(@user, user: { postal_address_attributes: { id: @user.postal_address.id, "_destroy" => true }}), method: :put, data: { confirm: 'Are you sure?' }
```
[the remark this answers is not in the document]
lilwillifo: Good point! I've made an update to make this a checkbox rather than a redirect link. I'll push it up shortly.

lilwillifo:
Hey @klappradla I made the requested changes. I rebased off of master, but am still getting a failing test around seasons on Travis CI. Not sure what's going on there. Thanks for your review on this!

klappradla:
For the failing spec, just in case you didn't see it - the output is rather verbose at the moment until #1048 is merged 😹 - Travis also prints, the seed (`45351` for your last build), so you can run

    bundle exec rspec --seed 45351

To get the same errors locally.

Your particular failure comes down to this idea of `Season.current`, which is shooting us in the foot ever since 😹 

You can fix the particular example by deliberately setting the other season to a time outside of the scope of the sequence in the default season factory:
```ruby
RSpec.describe 'Editing Teams', type: :feature do
  let(:user)            { create(:organizer) }
  let(:season)          { Season.current }
  let(:other_season)    { create(:season, :past) }
  let(:team)            { create(:team, name: 'Team Yay', season: season) }
  let(:selection_phase) { season.applications_close_at + 1.day }
  let!(:exercism)       { create(:project, :accepted, name: 'Exercism', season: season) }

 # etc.
```

app/models/postal_address.rb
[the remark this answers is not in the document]
klappradla: >  I'm doing any data validation around field inputs, we need a formatter somewhere to make sure the capitalization is correct

I'm sorry, but I fear I don't understand a word 🙈 If it's too much hassle to explain again, I can also try to read it again tomorrow when I'm more awake though - no worries.

app/views/users/edit.html.slim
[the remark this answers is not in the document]
klappradla: Ok. Do you like a checkbox better? I cannot really imagine how this looks like 🤔 Can you add a screenshot please?

app/views/postal_addresses/_postal_address.html.slim
```diff
@@ -0,0 +1,3 @@
+= @user.postal_address.line1.titlecase + " " + @user.postal_address.line2.titlecase + " "
```
klappradla: Hm, this being a bit _contra_ Rails' conventions, I'd say this is making the maintainability of the app a bit harder 😉 

The partial being in `postal_address/_postal_address`, you could render it by just calling
```ruby
= render @user.postal_address
```

Internally, this will pass a local assigned variable with the name `postal_address` to the partial - so there's a variable called `postal_address` available in the partial (same name as the model).

To fully make sure everything is according to plan, you could even `fetch` this variable:
```ruby
# I personally prefer to do a ruby: section in the beginning
ruby:
  postal_address = local_assigns.fetch(:postal_address)

= postal_address.line1

# etc.
```

Using `@user` here would be bypassing all this nice assignment magic that Rails does for us. Also - and that's the main point - it makes the partial incompatible for any other use aside from views with an `@user` set 😢

app/views/users/show.html.slim
```diff
@@ -65,9 +65,13 @@ nav.actions
     p = @user.company_info
 
   - if can_see_private_info?
-    - fields = ['tshirt_size', 'tshirt_cut', 'postal_address']
     .well.private-info
       h3 Private info
+      - if @user.postal_address
+        h4 Postal Address
+        p
```
klappradla: Has been some time that I read `slim` template code 🙈 

What's the point of this empty paragraph?  Should the postal-address partial rendered in a paragraph? If yes, then I think™ the way to go would be_
```ruby
p
  = render ...
```

But looking at the partial not using any markup at all, I'd rather suggest to do the markup in there.

db/migrate/20181107171037_create_postal_addresses.rb
```diff
@@ -0,0 +1,15 @@
+class CreatePostalAddresses < ActiveRecord::Migration[5.1]
+  def change
```
klappradla: Not suuuuper necessary, but it would come in quite handy if you could just squash the two migrations.

For instance: roll back the two migrations, edit this one to have the right names (and delete the other migration) and then run it again.

db/migrate/20181107172409_remove_postal_address_from_users.rb
```diff
@@ -0,0 +1,5 @@
+class RemovePostalAddressFromUsers < ActiveRecord::Migration[5.1]
+  def change
+    remove_column :users, :postal_address, :text
```
klappradla: Should we care about existing users having set up a `postal_address` already?  Would there be a way to migrate their data to the new schema?

spec/features/users/postal_address_spec.rb
```diff
@@ -0,0 +1,69 @@
+require 'rails_helper'
+
+RSpec.describe 'Add Postal Address', type: :feature do
+  let(:user) { create(:user) }
+  let(:address) { create(:postal_address) }
```
klappradla: As said in a previous comment, I'd personally prefer to just use actual strings, like "Canada" in the `fill_in` bits, rather then an address build via a factory... Nevertheless, using `create` here is not what we want. We don't want to create an address in the database **BEFORE** the test -> it's more the outcome of the test 😉

PR #1047, opened 2018-11-13
Rails 5.2.1

greysteil:
🎁


app/models/user.rb
```diff
@@ -75,7 +75,7 @@ def student
   accepts_nested_attributes_for :roles, allow_destroy: true
 
   before_save :normalize_location
-  after_create :complete_from_github
+  before_create :complete_from_github
```
greysteil: This is more efficient, as it will only do the save once. (I couldn't think of any reason why you'd want to do it twice.)

app/models/user.rb
```diff
@@ -121,8 +121,7 @@ def with_team_kind(kind)
     end
 
     def with_all_associations_joined
-      includes(:roles).group("roles.id").
-      includes(roles: :team).group("teams.id")
+      includes(roles: :team).group("roles.id").group("teams.id")
```
greysteil: Rails 5.2.1 didn't like that `roles` was being included twice, and there's no need to include it twice.

spec/mailers/application_form_mailer_spec.rb
```diff
@@ -50,15 +50,15 @@
 
     it 'renders N/A for the pair if not present' do
       expect(html_body).to match %r{<h2>About your pair</h2><p>N/A</p>}
-      expect(text_body).to match %r{About your pair\n  N/A}
+      expect(text_body).to match %r{About your pair\r\n  N/A}
```
greysteil: Rails 5.2 uses `\r\n` in place of `\n` when building the text part of emails.

greysteil:
This is refusing to build, I guess because it's from my fork rather than on this repo? Test should all be passing if ported across though.

app/models/user.rb
klappradla: I fear I don't understand the reasoning here 🙈 
Do you mean it's more efficient, since under the hood `complete_from_github` will persist a new version of the record?

app/models/user.rb
```diff
@@ -224,7 +223,7 @@ def complete_from_github
     attrs = Github::User.new(github_handle).attrs rescue {}
     attrs[:name] = github_handle if attrs[:name].blank?
     attrs = attrs.select { |key, value| send(key).blank? && value.present? }
-    update_attributes attrs
+    assign_attributes attrs
```
klappradla: Oh, now I get what you mean with your previous comment 💡

greysteil:
My pleasure! Was really easy to get up-and-running with the repo, so congrats on keeping this codebase so friendly to new contributors 🙂

app/models/user.rb
F3PiX: @klappradla I'm worried about this change. I didn't had the chance to dive into it today. But I wouldn't be surprised if this was intended to be `after_create` because of some other delicacies. And I'm not sure that it's covered by tests. 
I find that more worrying than the ineffeciency.

greysteil: I'm pretty confident on this one, but you folks know this codebase much better than I do.

The logic in the hook was just to add some additional details to the model, and fetched from GitHub. Any GitHub errors are already handled, so the method should never fail. I suppose that validations *could* fail on the updated attributes, and having this as `after_create` rather than `before_create` would yield different behaviour in that case - is that something you might care about here?

What I'm not sure of is why the tests were failing on Rails 5.2.1 without the change. For some reason the `email` wasn't getting set when this was an `after_create` but it is now that it's switched. That was the main motivation for me changing it.

klappradla: I already made a note to myself to use my next holidays to actually make it covered by tests @F3PiX 😉 

I get your concerns and actually had the same. But I could not find an obvious reason why doing it this way should not work. The flow does not include any other `after_commit` or anything hooks, so it should all even work without persisting 🤔 
Looking at it, I would probably even have hooked in before validating - since we're actually changing some attributes.

F3PiX: I'm still worried. It's not the obvious things I am worried about, but hidden side effects.:
> For some reason the email wasn't getting set....  
It's the 'for some reason' that worries me. Why was this affected by the upgrade to 5.2? 
IF this is going to have side effects that we're not aware of, it will not be manifest until next season, when new users are created and it will be a pain to trace it back to this change.

klappradla: Makes sense @F3PiX ! So let's roll back this PR and look into the details _"by hand"_ 👀 

I fear I won't have time for the "looking into" part within the next weeks though. I still roll back this PR. If you're fine I can do it later today - better to not have too many other changes depend on it.

F3PiX: 👍 
I wished I had time to look into it, but same here: I won't have time before the end of the year :-(
Thanks for rolling back, now I stop worrying :-)

PR #1048, opened 2018-11-13
Fix deprecation warnings

klappradla:
Related to the Rails update in #1047, this fixes some deprecation warnings introduced from that - just to have the output of the specs a bit easier on the eyes ✌️


klappradla:
With #1059 , this is no longer necessary 😞

PR #1063, opened 2018-11-22
Support OAuth callback proxy app

carpodaster:
# Why
* Github's OAuth apps (required to enable people to login with their Github account) allow only one callback url. We have different OAuth apps set up for dev, staging, and prod.
* The PR review apps are created with a dynamic hostname that will break the static, preconfigured callback setting.

# How
* We configure apps to redirect to an auxiliary app that does nothing more than doing a basic validity check and forwarding the validated OAuth request back to the staging or PR review app that initiated the OAuth process
* The code for that proxy app is here: https://github.com/rails-girls-summer-of-code/rgsoc-teams-oauth-callback-proxy 🔒 

# Testing / QA
* Configure [our OAuth app for staging](https://github.com/organizations/rails-girls-summer-of-code/settings/applications/48873) 🔒 to use the proxy app as its callback URI.
* Attempt to login with your github account from staging and a PR review app (should redirect to the proxy app when clicking on "login"; check your browser's network tab) as well as from a locally checked out Teams App running on localhost:3000 (should not touch the proxy app)

### DIFF
--- a/README.md
+++ b/README.md
@@ -226,3 +226,8 @@ The Heroku scheduler is set to run these tasks:
 ### Camo
 
 Both apps on Heroku use [camo](https://github.com/atmos/camo) to proxy insecure images in activity logs. For this to work, the `CAMO_HOST` and `CAMO_KEY` environment variables need to be set in Heroku's configuration for the app.
+
+### OAuth Callback for Review Apps
+
+In order to allow Github logins for our dynamically created PR review apps, the OAuth workflow is
+intercepted by [a simple web app that](https://github.com/rails-girls-summer-of-code/rgsoc-teams-oauth-callback-proxy). Please refer to its README for deployment and setup instructions.
--- a/config/initializers/devise.rb
+++ b/config/initializers/devise.rb
@@ -4,10 +4,15 @@
 GITHUB_APP_ID     = Rails.env.production? ? ENV['GITHUB_APP_ID']     : 'cc17063f6602cc4878ff'
 GITHUB_APP_SECRET = Rails.env.production? ? ENV['GITHUB_APP_SECRET'] : 'e93988593b6f7bfa1d0ff2cf10438ee3e547f8a4'
 
+# Set this to "https://rgsoc-teams-pr-oauth-proxy.herokuapp.com/oauth-proxy" for staging and review apps,
+# leave blank to stick with OmniAuth's default of "<scheme>://<host>/users/auth/github/callback":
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -226,3 +226,8 @@ The Heroku scheduler is set to run these tasks:
 ### Camo
 
 Both apps on Heroku use [camo](https://github.com/atmos/camo) to proxy insecure images in activity logs. For this to work, the `CAMO_HOST` and `CAMO_KEY` environment variables need to be set in Heroku's configuration for the app.
+
+### OAuth Callback for Review Apps
+
+In order to allow Github logins for our dynamically created PR review apps, the OAuth workflow is
+intercepted by [a simple web app that](https://github.com/rails-girls-summer-of-code/rgsoc-teams-oauth-callback-proxy). Please refer to its README for deployment and setup instructions.
```
klappradla: Suuuuper minor, but I'd reword "a simple web app that" to simply
```markdown
by a [oauth-callback-proxy](...) app.
```

config/initializers/devise.rb
```diff
@@ -4,10 +4,15 @@
 GITHUB_APP_ID     = Rails.env.production? ? ENV['GITHUB_APP_ID']     : 'cc17063f6602cc4878ff'
 GITHUB_APP_SECRET = Rails.env.production? ? ENV['GITHUB_APP_SECRET'] : 'e93988593b6f7bfa1d0ff2cf10438ee3e547f8a4'
 
+# Set this to "https://rgsoc-teams-pr-oauth-proxy.herokuapp.com/oauth-proxy" for staging and review apps,
+# leave blank to stick with OmniAuth's default of "<scheme>://<host>/users/auth/github/callback":
+GITHUB_OAUTH_CALLBACK_URI = ENV['GITHUB_OAUTH_CALLBACK_URI']
```
klappradla: Not necessarily related to this PR, but do you think it would be worth refactoring this to a classic `dotenv` setup?

We could include an example `.env.example` file and developers could adjust it to their needs?

Doing so, we could remove all the `Rails.env.production` things and simply use:
```ruby
GITHUB_APP_ID = ENV.fetch('GITHUB_APP_ID')

# or if it's really really optional:
GITHUB_APP_ID = ENV.fetch('GITHUB_APP_ID', nil)
```

Using `fetch` would have the benefit that the app would already crash when attempting to start instead of failing at some random point in the future.

For CI, there's also the possibility to encrypt environment variables - works like charm ✅ 

What do you think?

carpodaster:
I'm guessing this is no longer needed when we don't pursue PR apps any longer.

PR #1066, opened 2019-01-16
Update Rubocop strategy

klappradla:
This updates our Rubocop strategy from *"ignore all"* to *"include all and ignore later"* 🚔 

As we already discussed in person, this should help us get the application into an easier to maintain state and spare some efforts on PR reviews and onboarding new people, etc.

The approach for this is as follows: I had Rubocop generate a `TODO` file (see [the docs](https://docs.rubocop.org/en/latest/configuration/#automatically-generated-configuration) for how this works) with all the violations. I'll then go through all the entries in the file step by step and fix (or ignore) them.
I already used this approach on a huge legacy project in my old work and it worked out pretty well 👷 

If you're curious, this is the outcome of generating the todo file:
```❯ be rubocop --auto-gen-config
phase 1 of 2: run metrics/linelength cop
inspecting 365 files
.cc...cc.ccccc.c.ccc.c.c.cc...c.cc.cc..c.cc.cc.c...ccc.cc.c.....ccccc..c.cccc.cccc..cccccccc.cccc.c.cc...c...ccccc.c......c.cccc..cc..c.....c....cc.c..c.cc.cc.ccc.cccccc.ccccccc.ccc.ccccc.ccccc.cccccccccccc.cc.c.....cccccc....c..........c..c.c.cc.cc.ccc.c.c.ccccc.c..cccc..cc.c.c.ccc.ccccc.c.c..c.c..c.c.cc..cc....cc...c.cc.....cccc..c....c..c.c.c................cc

365 files inspected, 888 offenses detected
created .rubocop_todo.yml.
phase 2 of 2: run all cops
inspecting 365 files
cwcccccccccccccccwcccwccccccwcwcccccccwcccccwwcwcwcwwwccccccccccwccccccccwwcccccccccccwccccccccwcccccccwcccccccwcwcccccccwwcccccccccccccccccccccccccccccwcccwccccccccccccccwcwcwwccwccccwccccccccccccwwwcwwcwwccccccccccccccccccccccccccwcccccccccwcccccwcccccccccccccccccccwccwccccwwwccwwcccwcccwcwcccccccccccccccwwccccwccccwcwccccccccccccccccccccccccccccccccccccccccccc

365 files inspected, 4563 offenses detected
Created .rubocop_todo.yml.
```

I also already included the Rails and Rspec cops. Adding them later will just mean we're going through the same work again 😉

PR #1067, opened 2019-01-16
Rubocop Party part 1

klappradla:
This is part 1 of going through the Rubocop Todo file I created in #1066. I'm basically just running cops that can be run with the `autocorrect` flag.

It could be for now, that the results look suuuper off 😉 But don't worry, the too weird things will come up again later in the process again and I'll manually have to fix them. If not earlier then probably when dealing with the max. block and line length.

For now, I just want to have Rubocop do most of the work alone 👮


klappradla:
Ok, in order to make the "review" easier (if one can even do a proper review of things like this at all), I changed the base branch of this. So this is now targeting #1066 👷

klappradla:
Thinking about this again, this is probably way too overwhelming 🙈 

I'll find a better way to introduce the changes. Maybe via better commit messages..

F3PiX:
Thanks Max! The ToDo file is a great way to go through all. The thing is though, I find most cops nitpicking, and I'd prefer not to use the ones that are just 'opinion' without serving a greater goal. To give an idea: cops for the dynamic FactroyBot's , magic comment are things that improve code, more than 'just' style. Those will be a great help and time saver for both contributors and reviewers. 
But I (still) don't see that one space differences in alignment will decrease readability nor maintainability. While I do have strong opinions on other things, maybe some of which you or someone else don't care about. 
When we introduced Rubocop, we talked about implementing a simple version, only the minimal cops. I still like that idea best. Minimal to me is the 'greater goal' things like mentioned above and also the things we all wholeheartedly agree upon, like the Ruby hash style without arrows. 
I _do_, of course have passionate opinions on the things that I claimed are "'just' style". I take it we all have, and I don't think we should force those one way or the other.

I'd vote for a minimal implementation of Rubocop.  With a TODO file :-)

PR #1088, opened 2019-01-18
Rubocop fixes pt. 1 - the 🚀 edition

klappradla:
So, as said in #1067, I'll do this a bit more structured.

So this is round one of fixing a few low hanging fruits from the Rubocop TODO-file I created in #1066. 

This is **the whitespace edition™**, so it's about adding consistent spacing to arrays or hashes, removing extra empty lines, ensuring blank lines between methods, etc.

It probably does not make too much sense to go through the diff (you can still do so if you want to) because it's mostly shoving around blank lines. But you can have a look at the commit messages, they give you an overview of what I did.

**NOTE**: I'm targeting the branch of my previous PR to make the diff smaller.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -169,27 +169,6 @@ Layout/MultilineOperationIndentation:
     - 'app/exporters/applications.rb'
     - 'spec/controllers/supervisors/comments_controller_spec.rb'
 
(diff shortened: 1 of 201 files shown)


klappradla:
_(doing some rebasing from hell... - let's see)_

PR #1089, opened 2019-01-18
Rubocop fixes 2

klappradla:
This is part two of grabbing some low hanging fruits of #1066. This time it's about removing some obvious duplications. E.g. duplicated methods (only the 2nd one is used) or duplicated `private` modifiers (the first one is effective). So mostly things that probably slipped in while copy pasting things...


klappradla:
_(small rebase party...)_

klappradla:
Note: I also fixed another flaky spec (see 7ff9adc), but it does not really change anything logic wise.

So merging nevertheless ⛵️ 💨

PR #1096, opened 2019-01-25
Update directories excluded by rubocop

klappradla:
This updates the list of **excluded** directories for rubocop. For rubocop *"excluded"* can be read as *"ignored"*, so it won't even try to look for ruby files in these directories.

By excluding the `log` and `tmp` directories, we can speed up the local execution of rubocop a bit. Over the time (in Rails apps), especially the `tmp` directory collects loads of files (e.g. from `letter_opener`) and we want to spare rubocop having to dig through all of them.

Background: I recently ran into this at work, where rubocop would take +2 min on my local machine, because it would try to dig trough the bloated `storage` (from `ActiveStoreage`) and `tmp` directories first.

You can have a bit more insights on what things rubocop is actually looking at, if you run it with the *"debug"* flag:

    bundle exec rubocop -d


klappradla:
_(quickly rebasing this on top of current `master`)_

PR #1137, opened 2019-04-05
Update Rails version

klappradla:
I'm back 👋 

This updates Rails to the next minor version - `5.2.3`. There was a security vulnerability in the previous versions. For that, I also had to do some tiny adaptions to some tests here and there to make it work with this version of the framework.

Most notably is is probably moving retrieving additional information from Github from an `after_create` into a `before_create` hook. I'm not 100% sure why the `after_create` hook was no longer doing what it was supposed to do 🤔. Looking at the [order of events and callbacks](https://api.rubyonrails.org/classes/ActiveRecord/Callbacks.html) I think the problem lies in the callback being **after** create / but the record not being reloaded from the DB yet.

Doing it in a `before_create` hook fixes the problem and is overall way more straight forward anyways, as we can just adjust the attributes **before** writing it to the DB instead of writing to the DB, pulling it out and writing it again.


config/initializers/cookies_serializer.rb
```diff
@@ -2,4 +2,5 @@
 
 # Specify a serializer for the signed and encrypted cookie jars.
 # Valid options are :json, :marshal, and :hybrid.
-Rails.application.config.action_dispatch.cookies_serializer = :marshal
+# TODO: change this to :json on the next rails update
+Rails.application.config.action_dispatch.cookies_serializer = :hybrid
```
carpodaster: I tried to change to cookie serializer once and it caused thousands of cookie deserialisation errors. I am pretty sure I also used hybrid which is supposed to be the upgrade path, right? But it's been a while. Maybe thinks have gotten better? ;)

config/puma.rb
```diff
@@ -9,7 +9,7 @@
 
 # Specifies the `port` that Puma will listen on to receive requests; default is 3000.
 #
-port ENV.fetch("PORT") { 3000 }
+port        ENV.fetch("PORT") { 3000 }
```
carpodaster: 🤔

klappradla: 😹 interesting - will get rid of it...

klappradla:
Thanx @carpodaster! Build is still 🔴 though _(probably due to autoloading things 🤔 )_ - will have a look as soon as I get to it. Setting this to WIP until then.

config/application.rb
```diff
@@ -14,17 +14,14 @@ class Application < Rails::Application
     config.autoload_paths += %W(#{config.root}/app)
```
klappradla: Added this back @carpodaster. In Rails 5, the whole `app` directory is already autoloaded 🤷‍♂️. Forgot that this app uses `spring` which afaik somehow interferes with it.

I'll sort this out in a separate PR. With using `bootsnap` (which works like charm) one could probably even ditch `spring` without any performance / speed impact. I'll see.

PR #1141, opened 2019-04-08
Fix autoloading the app directory

klappradla:
This is a follow up to #1137, especially [this particular line](https://github.com/rails-girls-summer-of-code/rgsoc-teams/pull/1137#discussion-diff-272809807L14).

Background: Rails 5 by default autoloads the whole `app` directory _(see [the guides](https://guides.rubyonrails.org/autoloading_and_reloading_constants.html#autoload-paths-and-eager-load-paths) on all the details about autoloading)_ and encourages to put all the business logic the app requires for "running" there instead of `lib`. One is free to add as many custom subdirectories in `app` as one wants (e.g. `services` or `workers`) to structure the business logic - they'll also all be autoloaded.

In #1137 I initially removed the extra line adding `app` to the autoload path which made Travis turn 🔴 as it could not find all the files in the `Exporters` module. This is due to Rails' default autoloading not considering the first level of directories in `app` as part of the name of a file - it just treats it all as one big pool 😉 . See e.g. [this question on Stackoverflow](https://stackoverflow.com/questions/43998837/rails-app-autoloading-classes-defined-in-modules) where someone ran into the same problem when updating to Rails 5.

In order to have the `app/exporters` directory be recognized as an `Exporters` module, the path would need to be `app/exporters/exporter`, which is a bit weird. I therefore put them into `services` where they in my opinion still fit fine.

As a small drive-by fix I also removed `spring` from the app 🚜 . It's know to cause more troubles than pleasure and is not suuuuper necessary which `bootsnap` in place in my opinion.


klappradla:
Thanx for taking a look @carpodaster 🙏 

I just went with `services`, as it was the smaller change for not having to change namespaces _(using `exporters/exporters` just felt a bit weird)_ 🤷‍♂️

PR #1152, opened 2019-12-28
Fix projects#index not working properly during season transition

PandaWhisperer:
- If any past or current season is selected, only show selected projects for that season
- If an upcoming season was selected, show non-rejected projects
- If no season was selected, and the current season is in transition, show non-rejected projects
- If no season was selected, show selected projects for the current season


Montana:
good collab! glad it works :)

PR #1155, opened 2020-01-03
Fix 500 Error in community_controller#index when filtering by role

PandaWhisperer:
This fixes a SQL error on `/community` when filtering by role. 

Spec included.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ before_install:
   - sudo apt-get install chromium-chromedriver
   - gem update --system
   - gem --version
-  - gem install bundler
+  - gem install bundler -v 1.17.3
 before_script:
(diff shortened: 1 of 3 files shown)


Montana:
Great collab again! Going to close & merge.

Montana:
Ran locally, could not repro bug. Used Rollbar for error detection.

Montana:
Ran locally, could not repro bug. Used Rollbar for error detection via:

``` ruby
task :environment do
  Rollbar.configure do |config |
    config.access_token = '...'
  end
end
```

Montana:
Ran locally, could not repro bug. Used Rollbar for error detection via:

``` ruby
task :environment do
  Rollbar.configure do |config |
  end
end
```

PR #1166, opened 2020-02-04
Remove newsletter checkboxes and replace by signup link

PandaWhisperer:
As per To-Do list. Removes the checkboxes in the "Communication" section of the edit user page and replaces them with a link to the signup page (opens in new tab/window). 

Updated specs included.


Montana:
Rechecked locally, used Rollbar for double check via 

```ruby
require File.expand_path('../application', __FILE__)
require File.expand_path('../rollbar', __FILE__)

notify = ->(e) do
  begin
    Rollbar.with_config(use_async: false) do
      Rollbar.error(e)
    end
  rescue
    Rails.logger.error "Synchronous Rollbar notification failed.  Sending async to preserve info"
    Rollbar.error(e)
  end
end
```

Montana:
Remove newsletter checkboxes, and replace with signup link.

PR #1173, opened 2020-03-17
[Bugfix] Joining helpdesk team

PandaWhisperer:
## Fixes

- Allow "Join [Teamname]" button on `/team/:id` page to join the Helpdesk team
- Show error message when trying to join any other team
- Add specs for `JoinController`

### DIFF
--- a/app/controllers/join_controller.rb
+++ b/app/controllers/join_controller.rb
@@ -16,6 +16,8 @@ class JoinController < ApplicationController
   end
 
   def allow_helpdesk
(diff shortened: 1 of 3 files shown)


Montana:
Checked with Rollbar Scope, and approved. Good collab. 

```ruby
scope = {request: request_data,
         person: person_data,
         context: context_data
}
Rollbar.scoped(scope) do
  begin
    @app.call(env)
  rescue Exception => exception
    # ...
  end
end
```

